PR #63, opened 2020-05-05
Local grapl improvements

jgrillo-grapl:
- [x] Reduce log volume in rust services
- [x] Reduce log volume in python services
- [x] Implement logic to make services start up more gracefully
- [x] Reduce buffering period for local deployments (https://github.com/grapl-security/sqs-lambda/pull/2)


generic-subgraph-generator/src/main.rs
```diff
@@ -964,6 +961,7 @@ fn handler(event: SqsEvent, ctx: Context) -> Result<(), HandlerError> {
 
                 sqs_lambda::sqs_service::sqs_service(
                     queue_url,
+                    vec![], // FIXME: what should this be?
```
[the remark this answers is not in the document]
colin-grapl: https://github.com/grapl-security/grapl/blob/17-GraphGeneratorPlugin/graph-generator-lib/src/lib.rs#L289

For context, check this bit of code. Basically we have two types of messages to process:
* Messages that triggered the lambda
* Messages that we pull in during execution

We *only* want to fail the lambda in the event that any of the *Triggered* messages fails. Failing messages elsewhere is not an issue, they just won't get acked. So we have this bit of code that explicitly tracks messages provided exterrnally.

analyzer-dispatcher/Cargo.toml
```diff
@@ -5,7 +5,8 @@ authors = ["Insanitybit <insanitybit@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-graph-descriptions = {package="grapl-graph-descriptions" , version="*"}
+graph-descriptions = {package="grapl-graph-descriptions", version="*"}
+grapl-config = {path="../grapl-config"}
```
colin-grapl, line 9: Maybe we should just put grapl-config on crates.io? Any thoughts there?

sysmon-subgraph-generator/src/main.rs
```diff
@@ -963,7 +963,7 @@ async fn inner_main() -> Result<(), Box<dyn std::error::Error>> {
                 SendMessageRequest {
                     message_body: serde_json::to_string(&output_event)
                         .expect("failed to encode s3 event"),
-                    queue_url: "http://sqs.us-east-1.amazonaws.com:9324/queue/node-identifier-queue".to_string(),
```
jgrillo-grapl, line 966: This seems like a mistake

PR #65, opened 2020-05-13
Internal 9 rustfmt black gh actions

jgrillo-grapl:
- [x] apply black and rustfmt to all the rust and python code
- [x] configure github actions
- [x] make docker builds generate deployment artifacts


.github/workflows/grapl-lint.yml
```diff
@@ -0,0 +1,48 @@
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+
+      - name: Check rust formatting
+        run: |
+          cd src/rust
+          cargo fmt -- --check
+
+  python-lint:
```
colin-grapl, line 23: I think grapl_analyzerlib does this as a precommit hook, we'd probably want tests and linters running for an PR, not just deployment

etc/local_grapl/grapl_provision.py
```diff
@@ -291,6 +291,7 @@ def drop_all(client) -> None:
 
 
 if __name__ == '__main__':
+    time.sleep(10)
```
colin-grapl, line 294: I'd drop this a bit, since provisioning blocks other services from working. Maybe 2 or 3 seconds.

Connections to dgraph won't fail even if the server is in a bad state, and all of the operations should silently retry .

.github/workflows/grapl-lint.yml
jgrillo-grapl: I wrote this to trigger on the `check_run` event, which I *think* will fire every time a github check is executed. So if we configure a check, set branch protection on that check, then every push (e.g. to a PR branch) will trigger a `check_run` and run the lints. At least that's how I was hoping it works.

etc/local_grapl/grapl_provision.py
jgrillo-grapl: I believe I set this because I was seeing errors connecting to either localstack (sqs) or minio. Like one of those two hadn't come up yet, but the provisioner was trying to run against them and failing. I'll see if I can turn this down a bit and have it still work.

.github/workflows/grapl-lint.yml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: [3.6]
```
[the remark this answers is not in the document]
jgrillo-grapl: will fix

.github/workflows/grapl-deploy.yml
```diff
@@ -0,0 +1,107 @@
...
+          [ "$VERSION" == "staging" ] && VERSION=beta
+
+          echo '::set-env name=VERSION::$VERSION'
+
+      - name: Build Grapl
+        run: |
+          TAG=$VERSION docker-compose --build-args release_target=release -f docker-compose.yml -f docker-compose.build.yml build
+          docker system prune -f
+
+      - name: Deploy Grapl to AWS
```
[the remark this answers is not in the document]
jgrillo-grapl: will fix

PR #67, opened 2020-05-16
17 graph generator plugin

colin-grapl:
* Add a graph-model-plugin-deployer to make it easy to deploy Python plugins via Git hook or API
* Add a generator API that we can use to radically simplify building graph generators

You can see the changes to the graph-generator-lib here:
src/rust/graph-generator-lib/src/lib.rs


src/rust/sysmon-subgraph-generator/src/main.rs
```diff
@@ -699,337 +638,27 @@ where
...
-    Ok(())
-}
-
-fn main() -> Result<(), Box<dyn std::error::Error>> {
-    simple_logger::init_with_level(grapl_config::grapl_log_level())
-        .expect("Failed to initialize logger");
-
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    simple_logger::init_with_level(log::Level::Info).unwrap();
```
jgrillo-grapl, line 643: should probably use `grapl_config::grapl_log_level()` here

colin-grapl: I'll patch now.

PR #68, opened 2020-05-16
Use env loglevel

colin-grapl:
Minor change to use environment loglevel


src/rust/sysmon-subgraph-generator/src/main.rs
```diff
@@ -640,7 +640,7 @@ where
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    simple_logger::init_with_level(log::Level::Info).unwrap();
```
jgrillo-grapl, line 643: this should be `simple_logger::init_with_level(grapl_config::grapl_log_level());`

colin-grapl: Done

PR #71, opened 2020-05-19
build release workflow, remove deploy workflow, fix other workflows

jgrillo-grapl:
Hopefully this is all we need to do to populate a Github release with release artifacts.


.github/workflows/grapl-release.yml
```diff
@@ -27,9 +26,17 @@ jobs:
 
           echo '::set-env name=VERSION::$VERSION'
 
+      - name: Get upload URL for release corresponding to this release event
+        run: |
+          URL = "$(curl --request GET \
```
jgrillo-grapl, line 31: This mess seems like it could leak the github token...

colin-grapl: I don't think so. It's an HTTPS request to github. The only thing I could imagine is that we accidentally log something - are github action logs even public? Maybe run curl in quiet mode? '-o /dev/null' maybe

colin-grapl: Ah, I  guess if it logs the exec itself.

jgrillo-grapl:
> This looks good, but I have a few questions:
> 
> * How does versioning of github releases work?
> * Can we differentiate between staging/ master releases?

@colin-grapl versioning is based on tag. That's the `${{ github.ref }}` or `$GITHUB_REF`. I think you can either push a tag or do something in the UI and a tag is created for you, but in any case it's all based on git tags.

For master vs staging releases, you can select a branch to release from when cutting a release in the UI, and you can specify a branch via the `"target_commitish"` value when creating a release via the Releases API. This field is available from the [get release by tag name API](https://developer.github.com/v3/repos/releases/#get-a-release-by-tag-name).

I think we should probably put the branch and tag in the artifact names. What about something like ```sysmon-subgraph-generator-v1.2.3.zip``` for master and ```sysmon-subgraph-generator-v1.2.3-beta.zip``` for staging?

.github/workflows/grapl-release.yml
jgrillo-grapl: Yeah what I'm worried about is something like
1. `curl` errors and leaks the header to stdout
2. `python` errors and leaks the header via a stacktrace
3. `$UPLOAD_URL` now has the `$GITHUB_TOKEN` embedded in it
4. `Upload sysmon-subgraph-generator to Github` fails and leaks the `$GITHUB_TOKEN` back to the UI.. or something..

colin-grapl: You could pipe to an intermediary file, then read that through Python. Hard to imagine that messing up, if Python errors it'll be a key error, or an IO error, it wouldn't print the whole thing.

jgrillo-grapl: Yeah I played with it some, it seems safe.

PR #80, opened 2020-05-21
Update Grapl CDK

inickles-grapl:
This was made from a recent CDK app init template after running into issues with the previous version.

There are a number of differences from code layout, to string ids, to stack structure.

Tested on AWS. 👍


src/js/grapl-cdk/lib/dgraph.ts
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        const logDriver = new ecs.AwsLogDriver({
+            streamPrefix: `ecs${graph_name + id}`,
+        });
+
+        zeroTask.addContainer(id + 'Container', {
+            // --my is our own hostname (graph + id)
+            // --peer is the other dgraph zero hostname
+            image: ecs.ContainerImage.fromRegistry("dgraph/dgraph"),
```
colin-grapl, line 43: Can we pin to a version of dgraph? This is what we use locally:
`dgraph/standalone:v1.2.2`

src/js/grapl-cdk/lib/dgraph.ts
```diff
@@ -0,0 +1,195 @@
...
+
+        alphaTask.addContainer(id + graph_name + 'Container', {
+            image: ecs.ContainerImage.fromRegistry("dgraph/dgraph"),
+            command: [
+                "dgraph", "alpha", `--my=${id}.${graph_name}.grapl:7080`,
+                "--lru_mb=1024", `--zero=${zero}.${graph_name}.grapl:5080`,
+                "--alsologtostderr"
+            ],
+            logging: logDriver,
+            memoryReservationMiB: 2048,
```
colin-grapl, line 99: Can we calculate this value based on the EC2 instance? If we pass in a new EC2 instance with more memory it would suck to forget to change this value. Maybe just reserve 85% of the System? Not sure if that's straightforward to do, if not we can open an issue and punt.

src/js/grapl-cdk/lib/dgraph.ts
```diff
@@ -0,0 +1,195 @@
...
+
+        const logDriver = new ecs.AwsLogDriver({
+            streamPrefix: `ecs${graph_name + id}`,
+        });
+
+        alphaTask.addContainer(id + graph_name + 'Container', {
+            image: ecs.ContainerImage.fromRegistry("dgraph/dgraph"),
+            command: [
+                "dgraph", "alpha", `--my=${id}.${graph_name}.grapl:7080`,
+                "--lru_mb=1024", `--zero=${zero}.${graph_name}.grapl:5080`,
```
colin-grapl, line 95: If possible we should calculate the lru_mb.

DGraph represents 30% of system RAM, so we should do 30% of whatever we reserve for this container.

src/js/grapl-cdk/lib/engagement.ts
```diff
@@ -0,0 +1,253 @@
...
+                    fs.mkdirSync(packageDir);
+                }
+
+                const loginUrl = `https://${loginGatewayId}.execute-api.${aws.config.region}.amazonaws.com/prod/`;
+                const graphQLUrl = `https://${graphQLGatewayId}.execute-api.${aws.config.region}.amazonaws.com/prod/`;
+
+                const replaceMap = new Map();
+                replaceMap.set(`http://"+window.location.hostname+":8900/`, loginUrl);
+                replaceMap.set(`window.location.host+":5000/"`, graphQLUrl);
+
```
colin-grapl, line 221: This doesn't require the http://+? Is that because it's already in the code?

src/js/grapl-cdk/lib/redis.ts
```diff
@@ -0,0 +1,46 @@
...
+    readonly connections: ec2.Connections;
+    readonly cluster: elasticache.CfnCacheCluster;
+
+    constructor(scope: cdk.Construct, id: string, vpc: ec2.IVpc) {
+        super(scope, id);
+
+        // Define a group for telling Elasticache which subnets to put cache nodes in.
+        const subnetGroup = new elasticache.CfnSubnetGroup(this, id + '-subnet-group', {
+            description: `List of subnets used for redis cache ${id}`,
+            subnetIds: vpc.privateSubnets.map(function(subnet) {
```
colin-grapl, line 16: `vpc.privateSubnets.map(subnet => subnet.subnetId),`

I don't really care if we use this, but just fwiw since we're all learning js/ts we can use an arrow function here to cut it down.

src/js/grapl-cdk/lib/service.ts
```diff
@@ -0,0 +1,193 @@
...
+                environment: {
+                    IS_RETRY: "True",
+                    ...environment
+                },
+                timeout: cdk.Duration.seconds(360),
+                memorySize: 512,
+            });
+
+        event_handler.addEventSource(new SqsEventSource(queues.queue, { batchSize: 1 }));
+        event_retry_handler.addEventSource(new SqsEventSource(queues.retry_queue, { batchSize: 1 }));
```
colin-grapl, line 114: We can make this 10 now. This was to work around a bug, and that bug has been fixed for ages.

src/js/grapl-cdk/lib/userauthdb.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    allowRead(service: Service) {
+        this.user_auth_table.grantReadData(service.event_handler);
+        this.user_auth_table.grantReadData(service.event_retry_handler);
+    }
+
+    allowReadWrite(service: Service) {
+        this.user_auth_table.grantReadData(service.event_handler);
+        this.user_auth_table.grantReadData(service.event_retry_handler);
```
colin-grapl, line 39: This seems to only grant Read, the method name implies readWrite.

src/js/grapl-cdk/lib/dgraph.ts
inickles-grapl: I'm not seeing how to calculate based on instance type. Wouldn't we also need to factor the number of tasks (alphaCount and zeroCount) for their respective group capacity?

src/js/grapl-cdk/lib/engagement.ts
inickles-grapl: Hm. Looks like this is correct, as far as I can tell (not well), though not sure as this was merged from staging. I imagine it's intended to replace the following:

`h=function(){return window.location.host+":5000/"},g=window.location.host+":8900/"`

That snippet suggests there should be similar replace map like `replaceMap.set(`window.location.host+":8900/"`, loginUrl);`

I see what looks like there should also be a `replaceMap.set(`http://"+window.location.hostname+":5000/`, graphQLUrl);`

so I'll add those. Maybe @andrea-grapl can verify.

colin-grapl: `"http://"+window.location.hostname+":5000/"`

inickles-grapl: Nevermind, per our off-line discussion and rebuild of the edge_ux, I no longer see that first snippet I included in my last comment.

src/js/grapl-cdk/lib/dgraph.ts
inickles-grapl: I think we should come back to this one. We're launching a t3a.large, which has 32GB RAM, but default to one instance reserving soft limit of 2GB. I'll make a task.

src/js/grapl-cdk/lib/dgraph.ts
colin-grapl: FWIW no we would not, because tasks are 1:1 with instances. But let's punt and add an issue.

PR #89, opened 2020-05-25
Move JWT to SecretsManager

colin-grapl:
This PR moves our JWT out of lambda environment variables and into SecretsManager.

quick notes:
* The aws-cdk constructs now reference an EdgeJwtToken
* The graphql-endpoint, model-plugin-deployer, and engagement-edge now take the SecretId in their environment and fetch it via the SDK.


src/js/grapl-cdk/lib/grapl-cdk-stack.ts
```diff
@@ -371,9 +372,27 @@ class ModelPluginDeployer extends cdk.NestedStack {
     }
 }
 
+class EdgeJwtToken extends cdk.NestedStack {
```
inickles-grapl, line 373: I think we'd want to use a Construct over a NestedStack here, but I'm not sure we even want to break this out into its own class right now, since it does nothing more than the `secretsmanager.Secret` Construct. Just call `new secretsmanager.Secret` instead of this constructor.

I know the current CDK isn't a good example of this, but something we'll want to fix and do in the future is have all Constructs (and derivatives like Stacks, NestedStack, etc.) use a param list similar to `(scope: Construct, id: string, props: EdgeJwtTokenProps = {})`, where we also define an interface for `EdgeJwtTokenProps`. Convention is for the `id` to be supplied when instantiated, not defined, though seems like doing this would be a hack to help enforce it being a singleton it we wanted to.. I wouldn't recommend that..

PR #91, opened 2020-05-26
Publish grapl_graph_descriptions and grapl_analyzerlib to PyPI

jgrillo-grapl:
This PR updates the `grapl-release` workflow to publish python libraries to PyPI. If the release is a `beta` release the packages are published to test PyPI, otherwise they're published to PyPI.


.github/workflows/grapl-lint.yml
```diff
@@ -40,9 +40,16 @@ jobs:
           python3 -mvenv .venv
           source .venv/bin/activate
           python -mpip install --upgrade pip
-          pip install black
+          pip install black mypy
```
wittekm, line 43: Instead of explicitly defining that we install black/mypy here, could we use the mechanism in here:
https://github.com/wittekm/grapl_analyzerlib/commit/0eef2150c0adf80e50217e99c529f889318c7ff0#diff-5ab0aecf70891baa7cf39db5f248f58cR36

What this does is, it looks at the `setup.py`
```
https://github.com/grapl-security/grapl/blob/staging/src/python/grapl_analyzerlib/setup.py#L38
```
and installs:
- `.` (the "install_requires" stuff)
- `extras_require["dev"]` (the black/mypy stuff)

the reason being that this keeps our CI build system/reqs, and the requirements you use when you install it locally for development, exactly the same

PR #101, opened 2020-05-29
Use bump2version for python library releases

jgrillo-grapl:
- Fix dependencies between python libraries and python services s.t. local builds use the locally-built versions
- Remove generated code from grapl_analyzerlib
- Integrate bump2version into the grapl-release workflow to increment patch versions before attempting to push to PyPI


.github/workflows/grapl-release.yml
```diff
@@ -11,6 +11,37 @@ jobs:
...
+        run: |
+          python -mvenv venv && . venv/bin/activate
+          pip install bump2version
+          deactivate
+
+      - name: Bump grapl_graph_descriptions version
+        run: |
+          . venv/bin/activate
+          cd src/rust/graph-descriptions
+          bump2version --no-tag patch
```
jgrillo-grapl, line 29: One weird consequence here is that we'll never publish a `.0` patch version to PyPI. We'll never publish `grapl_graph_descriptions 1.0.0` or `2.0.0`. Instead, the version will be e.g. `2.0.1`. OTOH, we'll never fail to publish due to extant versions, which is the bug this PR is meant to address.

colin-grapl: Weird, but whatever.

PR #108, opened 2020-06-01
build in debian images instead of alpine

jgrillo-grapl:
We can't use Alpine images for the builds due to this issue: https://github.com/rust-lang/cargo/issues/7563#issuecomment-553526254

Our binaries need to be statically linked to work with AWS Lambda.


colin-grapl:
Works for me

PR #111, opened 2020-06-02
Grapl CDK cleanup - mostly naming.

inickles-grapl:
- Adding version info as description to Lambdas.
- Naming cleanup. Lambdas and SQS Queues have much better looking names than those generated from CDK/CloudFormation.
- Pinning dgraph version. This should have been in a previous commit.


src/js/grapl-cdk/lib/dgraph.ts
```diff
@@ -40,13 +40,13 @@ class Zero extends cdk.Construct {
         zeroTask.addContainer(id + 'Container', {
             // --my is our own hostname (graph + id)
             // --peer is the other dgraph zero hostname
-            image: ecs.ContainerImage.fromRegistry("dgraph/dgraph"),
+            image: ecs.ContainerImage.fromRegistry("dgraph/standalone:v1.2.2"),
```
colin-grapl, line 43: I don't think we want to use standalone. Standalone is not clustered or intended for production.

inickles-grapl: changed to `dgraph/dgraph:v1.2.2`

src/js/grapl-cdk/lib/grapl-cdk-stack.ts.orig
```diff
@@ -0,0 +1,539 @@
...
+
+        const master_graph = new DGraphEcs(
+            this,
+            'master-graph',
+            grapl_vpc,
+            mgZeroCount,
+            mgAlphaCount,
+        );
+
+<<<<<<< HEAD
```
jgrillo-grapl, line 440: Looks like a merge conflict wasn't fully resolved in this file

inickles-grapl: gah - this file shouldn't even be there. good catch.

PR #116, opened 2020-06-11
Bump graph generator lib

colin-grapl:
Simpy bumps us to 0.1.8

### DIFF
--- a/src/rust/graph-generator-lib/Cargo.toml
+++ b/src/rust/graph-generator-lib/Cargo.toml
@@ -3,6 +3,8 @@ name = "graph-generator-lib"
 version = "0.1.8"
 authors = ["Insanitybit <insanitybit@gmail.com>"]
 edition = "2018"
+description = "Library for writing Grapl generator services"
+license = "Apache-2.0"
 
 [dependencies]
 graph-descriptions = {package="grapl-graph-descriptions", version="*"}
--- a/src/rust/graph-generator-lib/Cargo.toml
+++ b/src/rust/graph-generator-lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "graph-generator-lib"
-version = "1.0.0"
+version = "1.0.1"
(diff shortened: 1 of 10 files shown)


src/rust/graph-generator-lib/Cargo.toml
```diff
@@ -1,19 +1,21 @@
...
+graph-descriptions = {package="grapl-graph-descriptions", version="0.2.10"}
 
 failure = "0.1.*"
 
 sha2 = "0.7.1"
 rusoto_s3 = { version="0.43.0", default_features = false, features=["rustls"] }
 rusoto_core = { version="0.43.0", default_features = false, features=["rustls"] }
 rusoto_sqs = { version="0.43.0", default_features = false, features=["rustls"] }
-grapl-config = {path="../grapl-config", version="*"}
+grapl-config = { version="0.0.1" }
```
jgrillo-grapl, line 18: We're no longer depending on the locally-built version?

src/rust/graph-generator-lib/Cargo.toml
```diff
@@ -1,19 +1,21 @@
...
-version = "1.0.0"
+version = "0.1.8"
 authors = ["Insanitybit <insanitybit@gmail.com>"]
 edition = "2018"
+description = "Library for writing Grapl generator services"
+license = "Apache-2.0"
 
 [dependencies]
-graph-descriptions = {package="grapl-graph-descriptions", path="../graph-descriptions", version="*"}
+graph-descriptions = {package="grapl-graph-descriptions", version="0.2.10"}
```
jgrillo-grapl, line 10: Why is this version pinned now?

colin-grapl: We have to if we want to publish to crates.io

colin-grapl: I can revert? But we'd have to do this every time we publish

PR #119, opened 2020-06-12
Implement DGraph TTL cleanup job

jgrillo-grapl:
This PR implements a cleanup job which runs as a scheduled AWS Lambda against DGraph, deleting any portion of the graph which has expired with respect to a configured TTL.


src/python/grapl-dgraph-ttl/app.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def delete_batch(client: GraphClient, uids: Iterable[str]) -> None:
+    del_ = {"delete": [{"uid": uid} for uid in uids]}
+
+    txn = client.txn()
+    try:
+        mut = txn.create_mutation(del_obj=del_)
+        txn.mutate(mutation=mut)
+        txn.commit()
```
colin-grapl, line 77: You can just pass `commit_now=True` into the .mutate call

src/python/grapl-dgraph-ttl/app.py
```diff
@@ -0,0 +1,91 @@
...
+    ttl_cutoff_s: int,
+    last_uid: Optional[str] = None,
+) -> Dict[str, Any]:
+    after = "" if last_uid is None else f" after: {last_uid}"
+    paging = f"first: {batch_size}{after}"
+    query = f"""
+    {{
+        q(func: le(last_index_time, {ttl_cutoff_s}) {paging}) {{
+            uid,
+            expand(_all_) {{ expand(_all_) }}
```
colin-grapl, line 34: We can turn this second `expand(_all_)` into just a `uid`, since to delete an edge you only need the src_uid, edge_name, dst_uid.

src/python/grapl-dgraph-ttl/app.py
```diff
@@ -0,0 +1,91 @@
...
+        }}
+    }}
+    """
+
+    txn = client.txn()
+    try:
+        return txn.query(query)
```
colin-grapl, line 41: You probably wanted to do `json.loads(txn.query(query).json)` here, based on the type annotation?

src/python/grapl-dgraph-ttl/app.py
```diff
@@ -0,0 +1,91 @@
...
+
+def calculate_ttl_cutoff_s(now: datetime.datetime, ttl_s: int) -> int:
+    delta = datetime.timedelta(seconds=ttl_s)
+    cutoff = now - delta
+    return int(cutoff.timestamp())
+
+
+def expired_node_uids(
+    client: GraphClient, now: datetime.datetime, ttl_s: int, batch_size: int
+) -> Iterator[Iterable[str]]:
```
colin-grapl, line 54: You're not going to want to return an Iterable[str]. Probably instead something like:

`Iterable[Union[str, Tuple[str, str, str]]]`

Since we're either returning the node uid to delete, or the reverse edges to delete. So far this code looks like it only deletes the uids, but we do need to delete the reverse edges too, or they'll hang.

PR #120, opened 2020-06-13
Adding ability for generator lib s3 client to assume AWS STS role.

inickles-grapl:
In order for Grapl generators to retrieve logs from buckets in other accounts we need them to assume STS role.

NOTE: this code is currently untested - technical difficulties, but it's up now for review. We'll test before merging in to staging.

- https://aws.amazon.com/premiumsupport/knowledge-center/lambda-function-assume-iam-role/
- https://github.com/rusoto/rusoto/blob/master/AWS-CREDENTIALS.md


inickles-grapl:
This has been tested with great success 👌

PR #121, opened 2020-06-18
CDK for grapl-dgraph-ttl

jgrillo-grapl:


src/js/grapl-cdk/lib/grapl-cdk-stack.ts
```diff
@@ -311,6 +311,44 @@ class EngagementCreator extends cdk.NestedStack {
...
+    constructor(
+        scope: cdk.Construct,
+        name: string,
+        prefix: string,
+        vpc: ec2.IVpc,
+        master_graph: DGraphEcs
+    ) {
+        super(scope, name + "-stack");
+
+        this.name = name + prefix;
```
inickles-grapl, line 330: I know this class is modeled after the ModelPluginDeployer, but let's just remove this line and member var def, as it looks backwards and it's not used.

src/js/grapl-cdk/lib/grapl-cdk-stack.ts
```diff
@@ -311,6 +311,44 @@ class EngagementCreator extends cdk.NestedStack {
     }
 }
 
+class DGraphTtl extends cdk.NestedStack {
+    readonly event_handler: lambda.Function;
+    readonly name: string;
```
inickles-grapl, line 318: Let's just remove these.

src/js/grapl-cdk/lib/grapl-cdk-stack.ts
inickles-grapl: I think we can remove `prefix` constructor param at that point.

PR #124, opened 2020-06-23
Integrate cdk-watchful

jgrillo-grapl:
This PR integrates cdk-watchful (https://github.com/eladb/cdk-watchful) into grapl-cdk to instrument our Lambdas and DynamoDB instances.


src/js/grapl-cdk/destroy_all.sh
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
+for BUCKET in $(aws s3 ls s3:// | cut -d' ' -f 3)
```
inickles-grapl, line 2: This will delete all buckets in the account, right? We can't do that.

I think if we want to support this script, we should either parse specific buckets created from templates in cdk.out/ directory, or build from new deployName param.

jgrillo-grapl: I figured the use-case for this was to completely blow away the entire account, e.g. to clean up our sandbox accounts. I can make it a little less heavy handed by doing a bit of filtering though.

jgrillo-grapl: Punting on this for now, it's going to be more involved than is warranted right now.

PR #126, opened 2020-06-24
Normalize dgraph dns name handling

jgrillo-grapl:
This PR changes how we handle DGraph DNS names and ports such that it's the same everywhere.


src/python/analyzer_executor/src/analyzer-executor.py
```diff
@@ -233,14 +233,21 @@ def chunker(seq, size):
     return [seq[pos : pos + size] for pos in range(0, len(seq), size)]
 
 
-def execute_file(name: str, file: str, graph: SubgraphView, sender, msg_id):
-    alpha_names = os.environ["MG_ALPHAS"].split(",")
+def mg_alphas() -> Iterator[Tuple[str, int]]:
```
jgrillo-grapl, line 236: This function is repeated in a lot of places, once we have a common library https://github.com/grapl-security/issue-tracker/issues/57 it can be factored out into that.

PR #127, opened 2020-06-24
CDK - Add support for multiple Grapl deployments to single AWS account

inickles-grapl:
Support deploying multiple version of Grapl to a single AWS account. This introduces a deploy name from which all resource and stack names are derived from.

Environment variables are no longer used inputs (bucket prefix or version). These are now defined in bin/grapl-cdk.ts, making it easier to track deployment state.

Additionally, all Constructs now use an interface for constructor params beyond scope and id.


src/js/grapl-cdk/lib/dgraph.ts
```diff
@@ -115,38 +115,43 @@ class Alpha extends cdk.Construct {
     }
 }
 
+export interface DGraphEcsProps {
```
jgrillo-grapl, line 118: This will need to be updated to contain `alphaPort` after merging staging into this branch.

src/js/grapl-cdk/.gitignore
```diff
@@ -2,7 +2,6 @@
 !jest.config.js
 *.d.ts
 node_modules
-.env
```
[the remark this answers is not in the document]
inickles-grapl: No more environment variables as inputs, so we don't need that file anymore. The bucket prefix (and grapl version string) will now be derived from deployName in bin/grapl-cdk.ts.

Better enables all inputs to a deployment to be saved and checked in. I imagine we'll want a branch for each deployment, where all inputs for it are saved. That way we can manage/upgrade deployments within an AWS account independently and safely.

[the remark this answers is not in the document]
jgrillo-grapl: OK got it, thanks.

src/js/grapl-cdk/lib/dgraph.ts
inickles-grapl: 👍 merge conflict forced my hand

src/js/grapl-cdk/lib/historydb.ts
```diff
@@ -36,19 +37,23 @@ export class HistoryDb extends cdk.Construct {
...
-        this.network_connection_history = create_table(this, 'network_connection_history_table');
-        this.ip_connection_history = create_table(this, 'ip_connection_history_table');
-        this.dynamic_session_table = create_table(this, 'dynamic_session_table');
+        this.proc_history = create_table(this, props.prefix + '-process_history_table');
+        this.file_history = create_table(this, props.prefix + '-file_history_table');
+        this.outbound_connection_history = create_table(this, props.prefix + '-outbound_connection_history_table');
+        this.inbound_connection_history = create_table(this, props.prefix + '-inbound_connection_history_table');
+        this.network_connection_history = create_table(this, props.prefix + '-network_connection_history_table');
+        this.ip_connection_history = create_table(this, props.prefix + '-ip_connection_history_table');
+        this.dynamic_session_table = create_table(this, props.prefix + '-dynamic_session_table');
```
colin-grapl, line 53: Can you make sure these table names are correct in the node-identifier?

We should probably just pass in the table names to that service via ENV, which will involve a bit of code change.

inickles-grapl: Oof, yeah looks like that'll break.

Agree about passing the names via env, in the name of the crusade against hard-coded vals. I'll start that.

src/rust/node-identifier/src/dynamic_sessiondb.rs
```diff
@@ -79,7 +79,7 @@ where
 
         let put_req = PutItemInput {
             item: serde_dynamodb::to_hashmap(&mapping).unwrap(),
-            table_name: "static_mapping_table".to_owned(),
+            table_name: std::env::var("STATIC_MAPPING_TABLE").expect("STATIC_MAPPING_TABLE"),
```
jgrillo-grapl, line 82: this should probably go in grapl-config, also we need to make sure this is being passed into the service as an env var in `docker-compose.yml`

src/rust/node-identifier/src/lib.rs
```diff
@@ -112,7 +112,10 @@ where
                     Some(unid) => unid,
                     None => bail!("Could not identify ProcessNode"),
                 };
-                let session_db = SessionDb::new(self.node_id_db.clone(), "process_history_table");
+                let session_db = SessionDb::new(
+                    self.node_id_db.clone(),
+                    std::env::var("PROCESS_HISTORY_TABLE").expect("PROCESS_HISTORY_TABLE"),
```
jgrillo-grapl, line 117: All these should also go in grapl-config, same comment about `docker-compose.yml`

src/rust/node-identifier/src/dynamic_sessiondb.rs
inickles-grapl: Yeah, good catch, have a docker-compose.yml locally.

I'll look into grapl-config.

jgrillo-grapl: @inickles-grapl yes that sounds right

src/rust/node-identifier/src/lib.rs
inickles-grapl: Resolved in newer commits

src/js/grapl-cdk/lib/historydb.ts
inickles-grapl: Resolved in newer commits.

PR #128, opened 2020-06-24
Default warn for dependencies

colin-grapl:
This is a very minor change that just flips logging level to 'warn' for our dependencies.


src/rust/grapl-config/src/lib.rs
```diff
@@ -65,11 +65,13 @@ pub fn grapl_log_level() -> log::Level {
 }
 
 pub fn _init_grapl_log(service_name: &str) {
-    let filter = EnvFilter::from_default_env().add_directive(
-        format!("{}={}", service_name, grapl_log_level())
-            .parse()
-            .expect("Invalid directive"),
-    );
+    let filter = EnvFilter::from_default_env()
+        .add_directive(format!("warn").parse().expect("Invalid directive"))
```
jgrillo-grapl, line 69: Should we instead set the dependency log level to `GRAPL_LOG_LEVEL` or maybe `GRAPL_DEP_LOG_LEVEL` or something configurable?

colin-grapl: RUST_LOG (loaded by from_default_env) should override it already, this is just a default basically.

jgrillo-grapl: Ah got it, thanks

PR #131, opened 2020-06-27
Routing

andrea-grapl:
Use URL fragments for routing using HashRouter from 'react-router-dom' rather than using LocalStorage. Tested successfully locally.


andrea-grapl:
https://github.com/grapl-security/issue-tracker/issues/75

src/js/engagement_view/src/components/SideBarContent.tsx
```diff
@@ -44,77 +48,127 @@ const useStyles = makeStyles({
...
-                    </TableRow>
+            <TableRow key={props.uid}>
+                <TableCell component="th" scope="row">
+                <Button className = {classes.lensName}
+                    onClick={
+                        () => { 
+                            props.setLens(props.lens)    
+                        }
+                }>
+                    {/* #TODO: change color of lens name based on score */}
```
colin-grapl, line 70: Can we get an issue opened to track this?

src/js/engagement_view/src/modules/uploads/plugins/apiRequests.tsx
```diff
@@ -30,7 +30,7 @@ export const deletePlugin = async ( pluginName: string ): Promise <boolean> => {
             body: JSON.stringify({plugins_to_delete: [pluginName]})
         }
     );
-    const body = await res.json();
+    await res.json();
     return true;
```
colin-grapl, line 34: Seems like we should be checking the response for success, not just returning true.

src/js/engagement_view/src/components/SideBarContent.tsx
andrea-grapl: Created issue: https://github.com/grapl-security/grapl/issues/132

PR #138, opened 2020-06-30
Fix cors error in graphql-endpoint

colin-grapl:
This was pretty simple. We didn't pass in the env var for the origin, now we do.


src/js/grapl-cdk/lib/grapl-cdk-stack.ts
```diff
@@ -21,6 +21,7 @@ import { EngagementEdge } from './engagement';
 import { GraphQLEndpoint } from './graphql';
 
 import { Watchful } from "./vendor/cdk-watchful/lib/watchful";
+import {RemovalPolicy} from "@aws-cdk/core";
```
inickles-grapl, line 24: My preference would be to not import like this - we've already imported aws-cdk/core. Instead of using `RemovalPolicy` use `cdk.RemovalPolicy` as we doe elsewhere in the file.

Looks like this re-import exists in engagement.ts as well, I think we should remove from both.

PR #141, opened 2020-07-01
modelPluginDeployer cdk hack

colin-grapl:


src/js/grapl-cdk/lib/grapl-cdk-stack.ts
```diff
@@ -394,6 +395,7 @@ class ModelPluginDeployer extends cdk.NestedStack {
         super(parent, id);
 
         const serviceName = props.prefix + '-ModelPluginDeployer';
+        this.integrationName = id + props.prefix + 'Integration';
```
inickles-grapl, line 398: I propose we do

```
this.integrationName = serviceName + '-Integration';
```

as we do for graphql: https://github.com/grapl-security/grapl/blob/3a5b65f8c7d608189f6e2ba3b40e42b1eed3c1f7/src/js/grapl-cdk/lib/graphql.ts#L20

We should also change Engagement Edge to match: https://github.com/grapl-security/grapl/blob/3a5b65f8c7d608189f6e2ba3b40e42b1eed3c1f7/src/js/grapl-cdk/lib/engagement.ts#L108

PR #146, opened 2020-07-02
Use dobi for builds

jgrillo-grapl:
This PR integrates https://github.com/dnephin/dobi into the build system, replacing `docker-compose build`.


.github/workflows/grapl-build.yml
```diff
@@ -25,7 +25,19 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Install dobi
+        run: |
+          wget https://github.com/dnephin/dobi/releases/download/v0.13.0/dobi-linux
+          chmod +x dobi-linux
+
       - name: Build Grapl
```
jgrillo-grapl, line 33: It just occurred to me that we can gain some parallelism in the CI build if we make 3 separate jobs here e.g. `Build Grapl Rust services`, `Build Grapl Python services`, and `Build Grapl JS services`. This could impact disk usage, we'll probably need to figure out how to make the Rust build complete before running the Python and JS builds in parallel.

jgrillo-grapl: This seems to have reduced PR build times by about 10min

jgrillo-grapl: @colin-grapl regarding release builds, the main reason to not parallelize builds is that we need the `dist` mount to be in the same Github Actions VM (as the release workflow is currently written). We could rewrite the release workflow to work around this by having a separate publish step for each build Job, but it just doesn't seem worth the effort. It's OK if release builds take a while IMO.

PR #150, opened 2020-07-07
CDK - manually create Lambda execution role so we can name it

inickles-grapl:
We refer to Lambda execution roles directly and the auto-generated names are sometimes not legible. This gives a clear name and uses a common role for both handlers.

### DIFF
--- a/src/js/grapl-cdk/lib/service.ts
+++ b/src/js/grapl-cdk/lib/service.ts
@@ -177,17 +177,14 @@ export class Service {
     }
 
     publishesToTopic(publishes_to: sns.ITopic) {
-        const topicPolicy = new iam.PolicyStatement();
-
-        topicPolicy.addActions('sns:CreateTopic');
-        topicPolicy.addResources(publishes_to.topicArn);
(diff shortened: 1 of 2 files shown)


inickles-grapl:
Tested this in sandbox with plugins.

src/js/grapl-cdk/lib/service.ts
```diff
@@ -78,6 +78,20 @@ export class Service {
             environment.RUST_BACKTRACE = '1';
         }
 
+        const role = new iam.Role(scope, 'ExecutionRole', {
```
jgrillo-grapl, line 81: should this role also get used in places where we manually define a non-Service lambda? E.g. grapl-dgraph-ttl and grapl-model-plugin-deployer?

inickles-grapl: Hmm. Good question. I'd say not necessary if we don't expect to ever work with them directly, as we do for generators connecting to log sources in other accounts, but still nice to have and easy enough to fix, so I'll add.

PR #153, opened 2020-07-09
Tests in CI

jgrillo-grapl:
This PR makes the existing tests work, and runs them in CI builds.


src/python/grapl_analyzerlib/tests/test_process_node.py
```diff
@@ -88,15 +92,15 @@ class TestProcessQuery(unittest.TestCase):
...
+    #     local_client = MasterGraphClient()
     #
     #     drop_all(local_client)
     #     provision()
 
     @hypothesis.settings(deadline=None)
     @given(process_props=process_props())
     def test_single_process_contains_key(self, process_props: ProcessProps) -> None:
-        local_client = DgraphClient(DgraphClientStub("localhost:9080"))
+        local_client = MasterGraphClient()
```
colin-grapl, line 103: Should this be LocalMasterGraphClient?

jgrillo-grapl: `LocalMasterGraphClient` and `MasterGraphClient` are identical in their implementations. Their behavior is completely determined by the `MG_ALPHAS` environment variable: https://github.com/grapl-security/grapl/blob/internal-37-tests/src/python/grapl_analyzerlib/grapl_analyzerlib/grapl_client.py#L8-L12

IMO we should probably get rid of one of them.

PR #155, opened 2020-07-13
Numerous bug fixes

colin-grapl:
I just went through and fixed a bunch of errors I had.

* Fixes cors errors in Python services by having them move to a regex
* Adds some env args to get logging in Python services
* Fixes an ACL issue where the model plugin bucket couldn't be read by the notebook
* Fixes a bug in the plugin retriever in grapl_analyzerlib, where it hadn't been updated to handle the new storage format


src/js/grapl-cdk/lib/engagement.ts
```diff
@@ -222,7 +223,12 @@ export class EngagementEdge extends cdk.NestedStack {
 }
 
 export class EngagementNotebook extends cdk.NestedStack {
-    constructor(scope: cdk.Construct, id: string, props: GraplServiceProps) {
+    constructor(
```
inickles-grapl, line 226: Build is currently broken, call to this in lib/grapl-cdk-stack.ts doesn't pass the new arg

inickles-grapl: Instead of adding params to constructor, I'd started extending the props argument, which seems to be std practice for CDK. I think we should do that here.

For example:
```
export interface NodeIdentifierProps extends GraplServiceProps {
    writesTo: s3.IBucket;
}

...

    constructor(parent: cdk.Construct, id: string, props: NodeIdentifierProps) {
```

https://github.com/grapl-security/grapl/blob/staging/src/js/grapl-cdk/lib/grapl-cdk-stack.ts#L71

PR #157, opened 2020-07-13
Don't unwrap a failed DGraph upsert

colin-grapl:
This PR is fairly trivial. Instead of unwrap'ing a DGraph transaction we just return an error, which will give us a 'partial failure' instead of a total failure, and allow the service to retry it later without blocking other transaction progress.


src/rust/graph-merger/src/main.rs
```diff
@@ -564,7 +561,7 @@ where
             }
         };
 
-        //        identities.into_iter().for_each(|identity| completed.add_identity(identity));
+        // identities.into_iter().for_each(|identity| completed.add_identity(identity));
```
inickles-grapl, line 570: should this just be removed?

colin-grapl: I can remove it and open a ticket to track that instead.

PR #165, opened 2020-07-15
Fix analyzer executor

colin-grapl:
This fixes a number of issues with how the AnalyzerExecutor loads plugins. In particular, it tried to write to paths in the lambda that are not writable, and now it writes to /tmp/, which also gets added to the sys.path .


src/js/grapl-cdk/bin/grapl-cdk.ts
```diff
@@ -5,7 +5,7 @@ import * as cdk from '@aws-cdk/core';
 import { GraplCdkStack } from '../lib/grapl-cdk-stack';
 import { EngagementUx } from '../lib/engagement';
 
-const deployName = 'Grapl-MYDEPLOYMENT';
+const deployName = 'Grapl-Meta-us-east-2';
```
colin-grapl, line 8: I need to change this back.

PR #172, opened 2020-07-21
Fix debug builds

colin-grapl:
We were previously skipping debug builds and producing empty binaries.


src/rust/Dockerfile
```diff
@@ -131,6 +131,10 @@ RUN if test "${release_target}" = "release"; then \
       cargo clean; \
       cargo build --target=x86_64-unknown-linux-musl --release; \
     fi
+RUN if test "${release_target}" = "debug"; then \
+      cargo clean; \
+      cargo build --target=x86_64-unknown-linux-musl; \
```
jgrillo-grapl, line 136: Doesn't the `cargo test` call on L128 do this already?

jgrillo-grapl: oh, it only builds node-identifier and node-identifier-retry-handler, because those are the only services with tests

PR #196, opened 2020-08-03
CDK changes for DGraph Docker Swarm cluster

jgrillo-grapl:

https://github.com/grapl-security/issue-tracker/issues/21

These changes introduce CDK functionality to configure an environment where we can use Docker Swarm to manage a DGraph cluster. Steps to provision the DGraph cluster are documented in the CDK README.

Tested manually in AWS sandbox.

image


src/js/grapl-cdk/lib/swarm.ts
```diff
@@ -0,0 +1,175 @@
...
+                "ec2:StartInstances",
+                "ec2:StopInstances",
+                "ec2:RebootInstances",
+                "ec2:TerminateInstances",
+                "ec2:CreateTags",
+                "route53:ListHostedZonesByName",
+                "route53:ChangeResourceRecordSets"
+            ],
+            resources: ["*"]
+            // resources: [
```
jgrillo-grapl, line 159: I couldn't make this work, not sure why. I guess we can remove `accountId` and `region` from `SwarmProps` since they're unused.

src/js/grapl-cdk/lib/grapl-cdk-stack.ts
```diff
@@ -392,6 +399,46 @@ class EngagementCreator extends cdk.NestedStack {
...
+    private readonly dgraphSwarmCluster: Swarm;
+
+    constructor(
+        parent: cdk.Construct, id: string, props: DGraphSwarmClusterProps
+    ) {
+        super(parent, id);
+
+        this.dgraphAlphaZone = new route53.PrivateHostedZone(
+            this,
+            id + "-DGraphSwarmZone",
```
inickles-grapl, line 418: For `id`s like this I suggest we don't concat the parent `id`. Just pass `DGraphSwarmZone` and let CDK do the rest for us. We used to have to do this for classes that didn't derive from cdk.Construct (indirectly here), but when it does we should avoid it.

src/js/grapl-cdk/lib/swarm.ts
inickles-grapl: Was just wondering why they were params and not from `cdk.Aws.ACCOUNT_ID` and `cdk.Aws.REGION`.

src/js/grapl-cdk/README.md
```diff
@@ -67,3 +74,109 @@ To deploy Grapl with the CDK, execute the following
...
+     and click *Start session*. A new window will open in your browser
+     with a terminal prompt on the bastion host.
+  2. Execute the following commands:
+
+``` bash
+# install docker
+sudo yum install -y docker
+
+# install docker-machine
+base=https://github.com/docker/machine/releases/download/v0.16.0 &&
```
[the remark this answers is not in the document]
jgrillo-grapl: Yeah it's important to keep track of these versions somewhere. Ideally when this workflow is scripted a script template can be populated automatically with a configured "blessed" version or something.. I'll make a note in the automation ticket.

src/js/grapl-cdk/lib/grapl-cdk-stack.ts
jgrillo-grapl: For posterity, Ian and I talked in Slack and this will be addressed in a subsequent PR.

PR #222, opened 2020-08-19
Docker Swarm CDK and provision

inickles-grapl:

There were a couple issues this resolves:

1. Grapl Jupyter provision from wasn't working due to DGraph not whitelisting IPs
2. Docker Swarm AWS resources need deployment prefix in their names to support multiple deployments within an AWS environment.

1. The DGraph alphas now whitelist 10.0.0.0/8
2. DGraph AWS resources now have Grapl deployment name prefixed
3. Docker/DGraph provision now enables the docker systemctl service, and we run the provision as `ec2-user` instead of `ssm-user`.

Tested by running a single deployment in AWS sandbox account and running all provision steps, as well as deployed AWS plugin and uploaded model plugin via Grapl UI.

### DIFF
--- a/src/js/grapl-cdk/README.md
+++ b/src/js/grapl-cdk/README.md
@@ -176,9 +176,10 @@ docker stack deploy -c docker-compose-multi.yml dgraph
 # add A records to route53 to make the alpha nodes discoverable
 AWS02_IP=$(docker-machine ip "$AWS02_NAME")
 AWS03_IP=$(docker-machine ip "$AWS03_NAME")
-HOSTED_ZONES_RESPONSE=$(aws route53 list-hosted-zones-by-name --dns-name "alpha.dgraph.graplsecurity.com")
+DNS_NAME=$(echo $GRAPL_DEPLOYMENT | awk '{print tolower($0)}').dgraph.grapl
+HOSTED_ZONES_RESPONSE=$(aws route53 list-hosted-zones-by-name --dns-name "$DNS_NAME")
 HOSTED_ZONE_ID=$(echo $HOSTED_ZONES_RESPONSE | python -c 'import json; import sys; print(json.load(sys.stdin)["HostedZones"][0]["Id"]);')
-echo {\"Changes\": [{\"Action\": \"UPSERT\", \"ResourceRecordSet\": {\"Name\": \"alpha.dgraph.graplsecurity.com\", \"Type\": \"A\", \"TTL\": 300, \"ResourceRecords\": [{\"Value\": \"$AWS01_IP\"}, {\"Value\": \"$AWS02_IP\"}, {\"Value\": \"$AWS03_IP\"}]}}]} > $HOME/batch.json
+echo {\"Changes\": [{\"Action\": \"UPSERT\", \"ResourceRecordSet\": {\"Name\": \"$DNS_NAME\", \"Type\": \"A\", \"TTL\": 300, \"ResourceRecords\": [{\"Value\": \"$AWS01_IP\"}, {\"Value\": \"$AWS02_IP\"}, {\"Value\": \"$AWS03_IP\"}]}}]} > $HOME/batch.json
 aws route53 change-resource-record-sets --hosted-zone-id $HOSTED_ZONE_ID --change-batch file://$HOME/batch.json
 
 # check that all the services are running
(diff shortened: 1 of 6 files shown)


src/js/grapl-cdk/README.md
```diff
@@ -52,9 +52,9 @@ export const graplVersion = 'YOUR_VERSION';
 
 Some tips for choosing a deployment name:
 
-  - Keep "Grapl" as prefix. This isn't necessary, but will help
+-   Keep "Grapl" as prefix. This isn't necessary, but will help
     identify Grapl resources in your AWS account.
-  - Choose a globally unique name, as this will be used to name S3
+-   Choose a globally unique name, as this will be used to name S3
     buckets, which have this requiement. Using a name that includes
```
colin-grapl, line 58: typo - requirement

src/js/grapl-cdk/README.md
```diff
@@ -137,28 +150,34 @@ docker-machine create --driver "amazonec2" --amazonec2-private-address-only --am
...
+eval $(docker-machine env "$AWS03_NAME" --shell sh)
 docker swarm join --token $WORKER_JOIN_TOKEN "$AWS01_IP:2377"
 
-# start DGraph
+# get DGraph compose template
 cd $HOME
-wget https://github.com/dgraph-io/dgraph/raw/master/contrib/config/docker/docker-compose-multi.yml
-eval $(docker-machine env aws01 --shell sh)
+wget https://github.com/grapl-security/grapl/raw/staging/src/js/grapl-cdk/docker-compose-multi.yml
+sed -e 's/aws01/'"${AWS01_NAME}"'/g' -i docker-compose-multi.yml
```
jgrillo-grapl, line 170: IMO rather than a `sed` here we should just parametrize the `docker-compose-multi.yml` with the environment variables. See e.g. what we do with tags here: https://github.com/grapl-security/grapl/blob/staging/docker-compose.yml#L59

src/js/grapl-cdk/lib/grapl-cdk-stack.ts
```diff
@@ -409,22 +409,25 @@ export class DGraphSwarmCluster extends cdk.NestedStack {
...
+                zoneName: 'alpha.dgraph.graplsecurity.com',
             }
         );
 
-        this.dgraphSwarmCluster = new Swarm(this, props.prefix + "-DGraphSwarmCluster", {
+        this.dgraphSwarmCluster = new Swarm(this, 'DGraphSwarmCluster', {
+            prefix: props.prefix,
             vpc: props.vpc,
-            internalServicePorts: [ec2.Port.tcp(5080), ec2.Port.tcp(7080)]
+            internalServicePorts: [ec2.Port.tcp(5080), ec2.Port.tcp(6080), ec2.Port.tcp(7080), ec2.Port.tcp(7081), ec2.Port.tcp(7082), ec2.Port.tcp(9080), ec2.Port.tcp(9081), ec2.Port.tcp(9082)],
```
jgrillo-grapl, line 430: 6080, 9080, 9081, and 9082 shouldn't be necessary to define here as they're external ports. We expose the external ports here: https://github.com/grapl-security/grapl/blob/staging/src/js/grapl-cdk/lib/grapl-cdk-stack.ts#L436

So *that* method should be updated to 

``` typescript
public allowConnectionsFrom(other: ec2.IConnectable): void {
    this.dgraphSwarmCluster.allowConnectionsFrom(other, ec2.Port.tcp(9080));
    this.dgraphSwarmCluster.allowConnectionsFrom(other, ec2.Port.tcp(9081));
    this.dgraphSwarmCluster.allowConnectionsFrom(other, ec2.Port.tcp(9082));
}```

src/js/grapl-cdk/README.md
inickles-grapl: yeah that's way better

src/js/grapl-cdk/lib/grapl-cdk-stack.ts
inickles-grapl: addressed in later commit

src/js/grapl-cdk/README.md
inickles-grapl: addressed in later commit

PR #224, opened 2020-08-20
Add a Python `grapl_common` lib.

wimax-grapl:
# NOTE
Depends on #225 
This PR is currently set up to merge commits into `grapl-security:internal-78-analyzer-deployment-system` - not staging - this is done because it was built on top of that branch.

I'll switch it back to staging later.


dobi.yaml
```diff
@@ -127,12 +127,21 @@ image=graph-descriptions-build:
   tags:
     - latest
 
+image=grapl-common-build:
+  image: grapl/grapl-common-python-build
+  context: src/python/grapl-common/
+  dockerfile: Dockerfile
+  target: grapl-common-python-build
+  tags:
```
jgrillo-grapl, line 135: I believe we need a `depends` clause here e.g. ```
depends:
  - grapl-graph-descriptions-python-build``` because in the dockerfile you have ```COPY --from=grapl/grapl-graph-descriptions-python-build /home/grapl/venv venv```

dobi.yaml
```diff
@@ -449,8 +460,6 @@ job=build-analyzer-deployer:
 job=run-analyzer-deployer-unit-tests:
   use: analyzer-deployer-build
   command: /bin/bash -c "source venv/bin/activate && cd analyzer-deployer && py.test -n auto -m 'not integration_test'"
-  depends:
-    - analyzer-deployer-build
```
wimax-grapl, line 453: fyi, according to dobi docs, `use: $image` auto includes it in the depends: list

jgrillo-grapl: 👍 makes sense

dobi.yaml
```diff
@@ -420,11 +430,13 @@ job=run-node-identifier-integration-tests:
...
 
+job=run-grapl-common-unit-tests:
+  use: grapl-common-build
+  command: /bin/bash -c "source venv/bin/activate && cd grapl_common/tests && ls && py.test -n auto -m 'not integration_test'"
+
 job=run-grapl-analyzerlib-unit-tests:
   use: grapl-analyzerlib-build
   command: /bin/bash -c "source venv/bin/activate && cd grapl_analyzerlib && py.test -n auto -m 'not integration_test'"
-  depends:
-    - grapl-analyzerlib-build
```
jgrillo-grapl, line 427: I think this might be necessary given https://github.com/grapl-security/grapl/pull/226/checks?check_run_id=1023683435

PR #225, opened 2020-08-21
Internal 78 analyzer deployment system

jgrillo-grapl:

https://github.com/grapl-security/issue-tracker/issues/78

This PR adds a new service skeleton for the `analyzer-deployer` which is responsible for managing analyzers. When complete, it will allow users to control analyzers' lifecycles including creation, deployment, updates, and teardowns.

Unit tests


dobi.yaml
```diff
@@ -406,6 +439,31 @@ job=run-grapl-analyzerlib-integration-tests:
     - grapl-analyzerlib-build
     - integration-env
 
+job=build-analyzer-deployer:
```
wimax-grapl, line 442: i _would_ say that i don't love the naming scheme on a lot of these - like, `build-analyzer-deployer` versus `analyzer-deployer-build` - i almost wonder if we should name the images like `analyzer-deployer-build-img` or something

docker-compose.yml
```diff
@@ -192,6 +192,23 @@ services:
   # Python services
   #
 
+  analyzer-deployer:
+    image: grapl/analyzer-deployer:${TAG:-latest}
+    command: /bin/sh -c '. venv/bin/activate && cd /home/grapl/app && chalice local --no-autoreload --host=0.0.0.0 --port=8125'
+    environment:
```
wimax-grapl, line 198: would prefer consistent "s

dobi.yaml
jgrillo-grapl: Yeah adding `-img` and `-job` be better. I was thinking "nouns for images, verbs for jobs" initially but that's not 100% obvious.

src/python/analyzer-deployer/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM grapl/grapl-python-build:latest AS analyzer-deployer-build
+USER grapl
+WORKDIR /home/grapl
+COPY --chown=grapl . ./analyzer-deployer
+COPY --from=grapl/grapl-python-build /home/grapl/venv venv
+RUN /bin/bash -c "source venv/bin/activate && pip install -r analyzer-deployer/requirements.txt && python -m mypy_boto3"
```
wimax-grapl, line 6: Could you add some comments on what all this zipping maintenance is doing? (like, I can roughly figure it out, but doesn't hurt to over-comment)

src/python/analyzer-deployer/analyzer_deployer/app.py
```diff
@@ -0,0 +1,104 @@
...
+    return CreateAnalyzerResponse("id", 0, "key")  # FIXME
+
+
+@app.route("/1/analyzers", methods=["POST"])
+def create_analyzer():
+    dynamodb_client = boto3.resource('dynamodb')
+    return dataclasses.asdict(_create_analyzer(dynamodb_client))
+
+
+# The view function above will return {"hello": "world"}
```
wimax-grapl, line 86: this looks like leftover stuff - there is no view function

src/python/analyzer-deployer/requirements.txt
```diff
@@ -0,0 +1,3 @@
+boto3==1.14.44
+boto3-stubs[essential]==1.14.44
+chalice==1.18.0
```
wimax-grapl, line 3: possibly also py.test?

src/python/analyzer-deployer/tests/test_default.py
```diff
@@ -0,0 +1,6 @@
+import unittest
```
wimax-grapl, line 1: since we have a "real" test, imo we should just delete this

Issue #232, opened 2020-08-26
Service-level metrics


wimax-grapl:
Current state:
- We have metrics on our infra (CDK stuff) through Watchful.
- Includes things like lambda exec times, dynamodb dashbaords, etc.

What we'd like to have:
- Runtime metrics emitted by services
- Example: In particular, we would love to understand the relationship between our data ingest and our data storage. 
    - Unique events ingested
    - MB ingested,
    - MB stored, etc.

wimax-grapl:
MVP order of operations:

Get one service (probably the sysmon generator?) emitting a statsd metric
Set up the subscription to the buffering lambda
Have the buffering lambda output to cloudwatch metrics

PR #236, opened 2020-08-29
MetricClient and statsd_format, but for Rust

wimax-grapl:
heyyyy it's my first real Rust diff

this is a straight up port of the python statsd_format to rust

relates to #232


src/rust/grapl-observability/src/statsd_formatter.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+pub fn statsd_format(
+    metric_name: &str,
+    value: f64, // should i make a union type?
+    typ: MetricType,
+    sample_rate: Option<f64>,
+    tags: &[TagPair],
+) -> Result<String, Error> {
+    // unlikely we'd go over this
```
wimax-grapl, line 59: remove comment, i was trying to buf something

src/rust/grapl-observability/src/statsd_formatter.rs
```diff
@@ -0,0 +1,161 @@
...
+            MetricType::Histogram => "h",
+        }
+        .to_string()
+    }
+}
+
+pub fn statsd_format(
+    metric_name: &str,
+    value: f64, // should i make a union type?
+    typ: MetricType,
```
wimax-grapl, line 55: rename metric_type

src/rust/grapl-observability/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+edition = "2018"
+description = "Metrics and logging tooling for Grapl"
+license = "Apache-2.0"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+log = "0.4.*"
+regex = "1.1.0"
+lazy_static = "1.2.0"
+failure = "0.1.*"
```
colin-grapl, line 14: failure is deprecated. Since this is a library you should just roll your own error type instead.

Check this out.
https://github.com/dtolnay/thiserror

src/rust/grapl-observability/src/metric_error.rs
```diff
@@ -0,0 +1,15 @@
+use core::fmt;
+use std::error::Error;
+
+#[derive(Debug)]
+pub struct MetricError {
+    // TODO: should this own a string or just have a reference to it?
```
colin-grapl, line 6: See my previous comment about error handling.

src/rust/grapl-observability/src/statsd_formatter.rs
```diff
@@ -0,0 +1,161 @@
...
+use lazy_static::lazy_static;
+use regex::Regex;
+
+lazy_static! {
+    static ref INVALID_CHARS: Regex = Regex::new("[|#,=:]").unwrap();
+}
+
+// TODO look at lifetimes, should this own or not, etc
+// seems odd to .to_string() everytihng going in here
+pub struct TagPair {
```
colin-grapl, line 13: You're right, this likely does not require ownership.

You can instead do:

```rust
pub struct TagPair<'a> {
    tag_key: &'a str,
    tag_value: &'a str
}

```

src/rust/grapl-observability/src/statsd_formatter.rs
```diff
@@ -0,0 +1,161 @@
...
+
+pub enum MetricType {
+    Gauge,
+    Counter,
+    Millis,
+    Histogram,
+}
+
+impl MetricType {
+    fn statsd_type(&self) -> String {
```
colin-grapl, line 41: These are all static strings so you can just return a &'static str . That'll avoid the allocation.

src/rust/grapl-observability/src/statsd_formatter.rs
```diff
@@ -0,0 +1,161 @@
...
+            MetricType::Millis => "ms",
+            MetricType::Histogram => "h",
+        }
+        .to_string()
+    }
+}
+
+pub fn statsd_format(
+    metric_name: &str,
+    value: f64, // should i make a union type?
```
colin-grapl, line 54: What are you looking to do?

src/rust/grapl-observability/src/statsd_formatter.rs
```diff
@@ -0,0 +1,161 @@
...
+    typ: MetricType,
+    sample_rate: Option<f64>,
+    tags: &[TagPair],
+) -> Result<String, Error> {
+    // unlikely we'd go over this
+    reject_invalid_chars(metric_name)?;
+
+    let mut sections: Vec<String> = Vec::with_capacity(6);
```
colin-grapl, line 62: If the end goal is to build up a String it's probably simpler to just do String::with_capacity() and use .push_str

It'll allocate less, though it does mean it might be slightly less pretty since you wouldn't do the join("|") but instead just push a "|" where appropriate.

src/rust/grapl-observe/src/statsd_formatter.rs
```diff
@@ -0,0 +1,199 @@
...
+        metric_type = metric_type.statsd_type()
+    )?;
+
+    match (metric_type, sample_rate) {
+        (MetricType::Counter, Some(rate)) => {
+            // a rate of 1.0 we'll just ignore
+            if rate >= 0.0 && rate < 1.0 {
+                write!(buf, "|@{sample_rate}", sample_rate = rate)?;
+            } else {
+                return Err(MetricInvalidSampleRateError());
```
jgrillo-grapl, line 80: I believe the `return` can be implicit here. I.e.
```rust
    } else {
        MetricInvalidSampleRateError()
    }
```

src/rust/grapl-observe/src/statsd_formatter.rs
```diff
@@ -0,0 +1,199 @@
...
+            MetricType::Histogram => h,
+        }
+    }
+}
+
+pub fn statsd_format(
+    metric_name: &str,
+    value: f64,
+    metric_type: MetricType,
+    sample_rate: Option<f64>, // TODO: change to an `impl Into<f64>`
```
jgrillo-grapl, line 58: Is this TODO still a thing?

src/rust/grapl-observe/src/statsd_formatter.rs
```diff
@@ -0,0 +1,199 @@
...
+            // separator
+            if !first_tag {
+                write!(buf, ",")?;
+            } else {
+                first_tag = false;
+            }
+            pair.write_to_buf(&mut buf)?;
+        }
+    }
+    return Ok(buf);
```
jgrillo-grapl, line 100: Do we also need to write a `"\n"` to the end here? The spec says each metric value is terminated by a newline, but I'm not 100% sure how this output is intended to be used. On that note, it would help to have a docstring.

src/rust/grapl-observe/src/statsd_formatter.rs
```diff
@@ -0,0 +1,199 @@
...
+    Millis,
+    Histogram,
+}
+
+impl MetricType {
+    fn statsd_type(&self) -> &'static str {
+        let g: &'static str = "g";
+        let c: &'static str = "c";
+        let ms: &'static str = "ms";
+        let h: &'static str = "h";
```
colin-grapl, line 43: These can actually be const.

src/rust/grapl-observe/src/statsd_formatter.rs
colin-grapl: Also you probably want Into<Option<f64>>

src/rust/grapl-observe/src/statsd_formatter.rs
wimax-grapl: I tried this, and since it's not the last thing in the function, it is not an implicit return.

I wrote and added a new test case here that would catch that.

src/rust/grapl-observe/src/statsd_formatter.rs
wimax-grapl: Totally reasonable, added a docstring saying that nobody should be directly consuming this, instead using the new MetricClient struct I've added.

src/rust/grapl-observe/src/statsd_formatter.rs
jgrillo-grapl: 👍 sounds good

src/rust/grapl-observe/src/metric_client.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+some followup TODOs:
+    - add tags to the public functions (not needed right now)
+    - is there any need for the `ms` metric type?
+*/
+impl MetricClient {
+    pub fn new() -> MetricClient {
+        let mut buf: String = String::with_capacity(256);
```
colin-grapl, line 15: Doesn't have to be mut.

src/rust/grapl-observe/src/metric_client.rs
```diff
@@ -0,0 +1,61 @@
...
+        return MetricClient { buffer: buf };
```
colin-grapl, line 16: No need for the return here

src/rust/grapl-observe/src/metric_client.rs
```diff
@@ -0,0 +1,61 @@
+use crate::metric_error::MetricError;
+use crate::statsd_formatter::{statsd_format, MetricType, TagPair};
+
+pub struct MetricClient {
```
colin-grapl, line 4: Can we add a `#[derive(Debug, Clone)]` here?

PR #237, opened 2020-09-01
Basic metric in Sysmond generator

wimax-grapl:
#232 

Adds 1 metric to Sysmond Subgraph Generator, mostly as an exampel and proof-of-concept.

bruh theres barely any tests in the rust codebase


src/rust/sysmon-subgraph-generator/src/metrics.rs
```diff
@@ -0,0 +1,25 @@
+use failure::Error;
+use grapl_observe::metric_reporter::{MetricReporter, TagPair};
+use log::*;
+
+#[derive(Clone)]
+pub struct SysmonSubgraphGeneratorMetrics {
```
wimax-grapl, line 6: the idea being we just add a bunch of functions here to do the actual logging; and keep passing around the SysmonSubgraphGeneratorMetrics object

src/rust/sysmon-subgraph-generator/src/metrics.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub metric_reporter: MetricReporter,
+}
+
+impl SysmonSubgraphGeneratorMetrics {
+    pub fn report_handle_event_success(&mut self, failed: &Option<Error>) {
+        let reported_status = if let Some(ref e) = failed {
```
wimax-grapl, line 12: would this be better off as a `failed.map_or_else("completed", "failed")`?

colin-grapl: At minimum instead of `ref e` it should just be `_`. map_or_else would be fine too, no real preference from my end, both are fairly clear.

colin-grapl (approved):

PR #238, opened 2020-09-02
Add a half-baked CDK Build to Docker/Dobi

wimax-grapl:

None

This adds a CDK build step to `dobi js`. It does not run tests yet.

`dobi js` + mucking with the CDK code to cause a typecheck error

### DIFF
--- a/src/js/grapl-cdk/Dockerfile
+++ b/src/js/grapl-cdk/Dockerfile
@@ -1,0 +1,16 @@
+# NOTE: this build is half-baked, and primarily supplies us with typechecks at CI time.
+# I haven't yet figured out a way to run the tests, primarily owing to the fact we need the ZIP'd artifacts
+# of other services/builds.
+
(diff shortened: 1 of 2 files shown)


src/js/grapl-cdk/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# NOTE: this build is half-baked, and primarily supplies us with typechecks at CI time.
+# I haven't yet figured out a way to run the tests, primarily owing to the fact we need the ZIP'd artifacts
```
jgrillo-grapl, line 2: I outlined two different potential approaches to this in the #infra channel. Otherwise, this is already better than what we've currently got so 💯

wimax-grapl: ^ for future reference:
https://grapl-internal.slack.com/archives/C017DKYF55H/p1599011803166100

PR #240, opened 2020-09-03
CDK framework for services to forward their MONITORING| logs to another lambda

wimax-grapl:
#232

Right now, nothing. it adds a new prop to each Service that lets us specify that that service's logs should be forwarded to another lambda.

did the cdk bootstrap which worked
npm run test, however, does not, it blows stack???


src/js/grapl-cdk/bin/grapl-cdk.ts
```diff
@@ -5,7 +5,7 @@ import * as cdk from '@aws-cdk/core';
 import { GraplCdkStack } from '../lib/grapl-cdk-stack';
 import { EngagementUx } from '../lib/engagement';
 
-const deployName = 'Grapl-MYDEPLOYMENT';
+const deployName = 'Grapl-wimax-test-09-02-2020';
```
wimax-grapl, line 8: oh, fuck

jgrillo-grapl: lol I do that almost every time too

jgrillo-grapl (approved):
It's unfortunate the CDK unit test doesn't work.. I don't really know why that is. The important thing is that the desired results are observed in sandbox, which sounds like the case here.

Issue #243, opened 2020-09-03
Ability to replay data back to DGraph


colin-grapl:
### What did you want to do?

In the event of data loss in DGraph we should be able to replay data from S3 back into the graph.

### What did you actually end up doing instead?

Currently we just lose data.

### What can Grapl do differently?

Data can be replayed into the graph-merger service, which will handle adding it back.

PR #244, opened 2020-09-03
Add a replay script

colin-grapl:
https://github.com/grapl-security/grapl/issues/243

This is a simple script that replays all data from the subgraph-generated-bucket into the graph-merger. This is for cases were data has been entirely lost from DGraph.

I've tested these changes locally.


etc/replay_data.py
```diff
@@ -0,0 +1,131 @@
...
+        QueueUrl=queue_url,
+        MessageBody=into_sqs_message(
+            bucket=output_bucket,
+            key=output_path,
+        ),
+    )
+
+
+def list_objects(client: S3Client, bucket: str) -> Iterator[str]:
+    return (
```
jgrillo-grapl, line 70: this could be more idiomatically (and clearly) written like this:
```python
def list_objects(client: S3Client, bucket: str) -> Iterator[str]:
    for p in client.get_paginator("list_objects_v2").paginate(Bucket=bucket):
        for e in p["Contents"]:
            yield p["Key"]
```
Also, do we need key checks? Those `thing[".."]` will throw `KeyError` if the key is absent.

jgrillo-grapl (approved):
1 nit, otherwise LGTM

PR #249, opened 2020-09-09
Metric Forwarder: parsing input, converting to Cloudwatch, sneding to Cloudwatch

wimax-grapl:
#232

`deser_logs_data`: deserialize the crazy gzip input from Lambda into pure strings
`cloudwatch_logs_parse`: parse those pure strings into a platform-agnostic `Stat` type
`cloudwatch_send`: turn Stat into Cloudwatch types and send it to Cloudwatch
`main.rs` please triple check my usage of Arc<Mutex

i have some very small unit tests, but unfortunately this is not super testable until it's complete.
I haven't added it as a service to our CDK yet, but that's up next, at which point I'll be able to test on my personal instance.

### DIFF
--- a/src/rust/metric-forwarder/src/cloudwatch_logs_parse.rs
+++ b/src/rust/metric-forwarder/src/cloudwatch_logs_parse.rs
@@ -32,17 +32,18 @@ fn parse_log(log_str: &str) -> Result<Stat, MetricForwarderError> {
            <ts>                  \t         <dont care>                \t <log contents>\n
     */
     let split: Vec<&str> = log_str.splitn(3, CLOUDWATCH_LOGS_DELIM).collect();
-    if split.len() != 3 {
-        return Err(MetricForwarderError::PoorlyFormattedLogLine(
+    match &split[..] {
+        [timestamp, _, statsd_component] => {
+            let stat = parse_statsd_component(statsd_component).map(|statsd_message| Stat {
+                timestamp: timestamp.to_string(),
+                msg: statsd_message,
+            });
+            stat
+        }
+        _ => Err(MetricForwarderError::PoorlyFormattedLogLine(
             log_str.to_string(),
-        ));
+        ))
     }
-    let timestamp = split[0];
-    let statsd_component = parse_statsd_component(split[2]);
-    statsd_component.map(|statsd_message| Stat {
-        timestamp: timestamp.to_string(),
-        msg: statsd_message,
-    })
 }
 
 fn parse_statsd_component(log_str: &str) -> Result<statsd_parser::Message, MetricForwarderError> {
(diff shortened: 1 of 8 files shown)


src/rust/metric-forwarder/src/cloudwatch_send.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    let request_futures = put_requests.map(|input| client.put_metric_data(input));
+    let responses: Vec<PutResult> = future::join_all(request_futures).await;
+
+    // TODO: retries
+    let failures: usize = responses.iter().filter(|resp| resp.is_err()).count();
+    match failures {
+        0 => Ok(()),
+        _ => Err(MetricForwarderError::PutMetricDataError(failures)),
+    }
```
wimax-grapl, line 65: unsure if this should be 'Err if any failure' or just a warn

colin-grapl: I think it might make more sense to collect the errors into the Err itself, or at least one of them. Seems fine to return an Err here and then let the caller decide how to log/ handle it, or retry it, etc.

src/rust/metric-forwarder/src/cloudwatch_logs_parse.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+fn parse_log(log_str: &str) -> Result<Stat, MetricForwarderError> {
+    /*
+    a typical CloudWatch Logs log message looks like this:
+        "2017-04-26T10:41:09.023Z\tdb95c6da-2a6c-11e7-9550-c91b65931beb\tmy log message\n"
+           <ts>                  \t         <dont care>                \t <log contents>\n
+    */
+    let split: Vec<&str> = log_str.splitn(3, CLOUDWATCH_LOGS_DELIM).collect();
+    if split.len() != 3 {
```
colin-grapl, line 35: You could destructure the slice like this:

```rust
match &split[..] {
    [timestamp, _, statsd_component] => {do your thing},
    incorrect => {} // any other length goes here
}
```

nbd, just a note

src/rust/metric-forwarder/src/cloudwatch_send.rs
```diff
@@ -0,0 +1,177 @@
...
+    async fn put_metric_data(&self, input: PutMetricDataInput) -> PutResult;
+}
+
+#[async_trait]
+impl<T> CloudWatchPutter for T
+where
+    T: CloudWatch + Sync + Send,
+{
+    async fn put_metric_data(&self, input: PutMetricDataInput) -> PutResult {
+        self.put_metric_data(input).await
```
colin-grapl, line 33: I'm a little worried about infinite recursion here. It might be better to do this via explicit universal call syntax

```rust
CloudWatch::put_metric_data(&self, input).await
```

colin-grapl (approved):

PR #253, opened 2020-09-11
Add steps to release workflow for grapl-common

wimax-grapl:
I realized I missed this in the original diff. Basically copy-paste of the analyzerlib one


jgrillo-grapl (approved):
LGTM

PR #262, opened 2020-09-16
Fix issue with grapl-model-plugin-deployer

colin-grapl:
This is a hotfix. I'll create an issue for this retroactively, but I want to start the PR now.

When deploying new nodes, we need to fetch the existing schemas of nodes from dgraph and reconstruct them as python
The problem is that we don't know what the reverse edges are, ie: We can know that it's a ManyToX relationship, but not if it's a ToOne or a ToMany
We basically just set the reverse edge to null and say "don't worry about it, just worry about what we know"
There was an inversion in some of the code that led to the edge being OneToMany, not ManyToOne, and so that broke some of our model plugin deployer code.

Tested via localgrapl.

### DIFF
--- a/etc/local_grapl/grapl_provision.py
+++ b/etc/local_grapl/grapl_provision.py
@@ -408,8 +408,6 @@ if __name__ == "__main__":
             if i > 10:
                 LOGGER.error("mg provision failed with: ", e)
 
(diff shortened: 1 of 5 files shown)


jgrillo-grapl (approved):
👍

PR #267, opened 2020-09-17
Promote Metric Forwarder to a true CDK service (and a bunch of random fixes I found along the way)

wimax-grapl:
#232

- Add all the dobi/docker/cdk scaffolding for Metric Forwarder as a lambda service; also some bugfixes in the docker files
- quicker parallelized `extract-grapl-deployment-artifacts`
- Added Sysmon as a 'client' of the metric forwarder (its logs are sent to the metric forwarder)
- Completely revamped the printlns for monitoring
(Before: Assumed the Lambda runtime spit out timestamp info; it does not, so I now encode it, like `MONITORING|timestamp_here|statsd_stuff_here`)
- Add some debuggability to the metric forwarder (info!s, better errors)

Deployed to CDK, and it works! I see the sysmon metric in Cloudwatch.

### DIFF
--- a/dobi.yaml
+++ b/dobi.yaml
@@ -386,7 +386,7 @@ compose=integration-env:
     # rust images
     - analyzer-dispatcher
     - generic-subgraph-generator
-    - metric-forwarder # TODO should I???
+    - metric-forwarder  # though, not currently used in integration
     - graph-merger
     - node-identifier
     - node-identifier-retry-handler
--- a/dobi.yaml
+++ b/dobi.yaml
@@ -39,6 +39,18 @@ image=analyzer-dispatcher:
   depends:
     - rust-build
 
+image=metric-forwarder:
+  image: grapl/metric-forwarder
+  context: src/rust/metric-forwarder
+  dockerfile: Dockerfile
+  args:
+    release_target: "{env.GRAPL_RELEASE_TARGET:debug}"
+  target: grapl-metric-forwarder
+  tags:
+    - "{env.TAG}"
+  depends:
+    - rust-build
+
 image=generic-subgraph-generator:
   image: grapl/grapl-generic-subgraph-generator
   context: src/rust/generic-subgraph-generator
@@ -374,6 +386,7 @@ compose=integration-env:
     # rust images
     - analyzer-dispatcher
     - generic-subgraph-generator
+    - metric-forwarder # TODO should I???
     - graph-merger
     - node-identifier
     - node-identifier-retry-handler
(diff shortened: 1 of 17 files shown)


dobi.yaml
```diff
@@ -374,6 +386,7 @@ compose=integration-env:
     # rust images
     - analyzer-dispatcher
     - generic-subgraph-generator
+    - metric-forwarder # TODO should I???
```
wimax-grapl, line 389: since we haven't figured out a way to loop this into integration yet, unsure if it's worth including here

src/js/grapl-cdk/extract-grapl-deployment-artifacts.sh
```diff
@@ -2,18 +2,29 @@
...
 if [ -z "$VERSION" ]
 then
-  echo "Please set VERSION= (it's same as TAG= for dobi)"
+  echo "Please set VERSION="
   exit 1
 fi
 
+export GRAPLROOT=$(realpath ../../..)
+export ZIPS_DIR=$(realpath zips)
+
```
wimax-grapl, line 11: my bash-fu is not phenomenal, and I overuse export VAR=thing when I could drop the export

src/rust/grapl-observe/src/metric_reporter.rs
```diff
@@ -17,6 +18,7 @@ pub struct MetricReporter {
     to stdout; then, later, a lambda reads in these messages and writes them to Cloudwatch.
     (originally recommended in an article by Yan Cui)
     */
+    // TODO: I think I gotta mutex or arc this; two threads grab it at once and sometimes print the wrong thing!
     buffer: String,
```
wimax-grapl, line 22: turns out my understanding of rust borrowing is not good; i thought the borrow checker might, somehow, insert locks to prevent two threads from mutating this buffer at once. I was very wrong! Turns out I was seeing some unexpected re-use of the buffer within sysmon-generator during some runs.



This will be my next task.

src/rust/sysmon-subgraph-generator/src/main.rs
```diff
@@ -668,7 +669,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     grapl_config::init_grapl_log!();
     info!("Starting sysmon-subgraph-generator");
 
-    let metrics = SysmonSubgraphGeneratorMetrics {
+    let mut metrics = SysmonSubgraphGeneratorMetrics {
```
wimax-grapl, line 672: woops this is not necessary

dobi.yaml
jgrillo-grapl: IMO definitely worth including it here, if only as a reminder that it exists. This list should stay in sync with the docker-compose.yml--e.g. every service in that file should be represented here.

colin-grapl (approved):

PR #269, opened 2020-09-17
Dependency inject a std::io::Write into MetricReporter for better testing

wimax-grapl:
#232 i guess

better testing bruh

cargo test


wimax-grapl:
getting a Writer callback is far too frustrating, whatever

PR #270, opened 2020-09-18
better testing around MetricReporter - dependency inject utcnow, Write

wimax-grapl:
#232 metrics

modify the MetricReporter so we can inject stuff and test it

unit tests baybeeee


jgrillo-grapl (approved):
LGTM

PR #287, opened 2020-09-22
added EngagementCreatorMetrics as a sample of metrics in a Python service

wimax-grapl:
#232 

Adds rudimentary metrics to one of our service, engagement creator.
...also:
- fixed up grapl_common a bit (the setup.py was slightly wrong)
- more strict mypy in grapl_common and engagement-creator
- introduced `histogram_ctx()`
 

added some tests

### DIFF
--- a/src/python/engagement-creator/src/engagement-creator.py
+++ b/src/python/engagement-creator/src/engagement-creator.py
@@ -5,7 +5,7 @@ import sys
 import time
 import traceback
 from collections import defaultdict
-from typing import Any, Dict, Iterator, Tuple, Sequence, Optional, cast, TypeVar
+from typing import Any, Dict, Iterator, Tuple, Sequence, Optional, cast, TypeVar, Type
 
 import boto3
 import botocore.exceptions  # type: ignore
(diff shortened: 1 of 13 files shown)


src/python/engagement-creator/src/engagement-creator.py
```diff
@@ -206,92 +211,109 @@ def nodes_to_attach_risk_to(
...
     mg_client = MasterGraphClient()
-
     s3 = get_s3_client()
+    metrics = get_metrics()
+
     for event in s3_event["Records"]:
-        if not IS_LOCAL:
-            event = json.loads(event["body"])["Records"][0]
+        with metrics.time_to_process_event():
+            _process_one_event(event, s3, mg_client, metrics)
```
jgrillo-grapl, line 225: If an exception is thrown in `_process_one_event(..)` we'll record a duration but not an "events_processed" count. This seems like the right thing to me as "events_processed" are only the successfully processed events, but should we also count failures? The use case I'm thinking of is that if error handling is very slow, so a block of errors bogs down processing, it would be nice to be able to correlate the latency increase with a failure spike.

jgrillo-grapl (approved):
A couple minor questions/comments, nothing blocking.

src/python/engagement-creator/src/engagement-creator.py
wimax-grapl: Sure, makes sense. I'll do

```
            try:
                _process_one_event(event, s3, mg_client)
            except:
                metrics.event_processed(status="failure")
                raise
            else:
                metrics.event_processed(status="success")
```

src/python/engagement-creator/src/engagement-creator.py
```diff
@@ -206,92 +211,109 @@ def nodes_to_attach_risk_to(
     return [node for node in nodes if node.node_key in risky_node_keys_set]
 
 
+def get_metrics() -> EngagementCreatorMetrics:
```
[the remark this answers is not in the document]
wimax-grapl: i'll go with overly verbose `create_metrics_client()`

Issue #231, opened 2020-08-26
E2E integration tests 


wimax-grapl:
Get end to end integration tests running in CI/CD with actual integration tests. 

Things like 
- Can I deploy a plugin? 
- Do analyzers that use plugins work? 
- Does the system actually start and reach a steady state?

wimax-grapl:
https://grapl-internal.slack.com/archives/C017DKYF55H/p1600876549003500
> Except these tests are different. Rather than e.g. testing how service code runs against DGraph or DynamoDB, they test how grapl services interact with eachother.
> 
> I.e.
> (1) Deploy an analyzer
> (2) Check that the analyzer is actually deployed properly
> (3) Drop data in S3
> (4) Check that the analyzer fires
> (5) Check that the expected data are in DGraph

other notes:
- use `depends` but it doesnt guarantee health checks - our 'service' has to do that
- sounds like the service will basically have an entry point, "wait on" all resources, and then call into a standard testing harness
- would be really nice if services exposed healthcheck endpoints

wimax-grapl:
Okay - definitely needs more tests, but _one_ is done ("check that the expected data are in dgraph")

PR #298, opened 2020-09-25
Add one more sleep to e2e test

wimax-grapl:
#231 

The e2e test failed once yesterday because the lens only had 1 node in scope instead of 3. turns out adding those edges is not atomic. 
Upon re-run it worked fine.

by submitting this diff, yolo


colin-grapl (approved):

jgrillo-grapl (approved):
👍

PR #305, opened 2020-09-28
Metric Forwarder: aggregate similar metrics to lower the total number of calls to CloudWatch (and save money)

wimax-grapl:
#232 

```
     * This section of the code will convert input of (boiled down for simplicity)
     *   [MetricDatum('metric', 1.0), MetricDatum('metric', 5.0), MetricDatum('metric', 1.0)]
     * into
     * {
     *   MetricDatum('metric', _): {
     *     // value -> num_occurences
     *     1.0: 2,
     *     5.0: 1
     *   }
     * }

and then

    /*
     * This section will take the above `datum_to_count_map` and convert it into a flat vec<MetricDatum>, where each
     * MetricDatum internally contains the {1.0: 2, 5.0: 1} in its .counts and .values
     */
```

i wrote unit tests


src/rust/metric-forwarder/src/accumulate_metrics.rs
```diff
@@ -0,0 +1,250 @@
...
+     * This section will take the above `datum_to_count_map` and convert it into a flat vec<MetricDatum>, where each
+     * MetricDatum internally contains the {1.0: 2, 5.0: 1} in its .counts and .values
+     */
+    datum_to_count
+        .into_iter()
+        .map(|(WrappedMetricDatum(mut datum), count_map)| {
+            // Erase the scalar "value", then fill "counts" and "values" vecs
+            datum.value = None;
+            let mut counts = vec![];
+            let mut values: Vec<f64> = vec![];
```
colin-grapl, line 46: I think you could do *something* like
```rust
let (counts, values) = count_map.drain().partition();
```
Or at the least it seems like a fine place to do a `Vec::with_capacity(count_map.len())`?

src/rust/metric-forwarder/src/accumulate_metrics.rs
```diff
@@ -0,0 +1,250 @@
...
+            && self.0.dimensions == other.0.dimensions
+            && self.trunc_timestamp() == other.trunc_timestamp();
+    }
+}
+impl Eq for WrappedMetricDatum {}
+
+struct WrappedDimensions<'a>(&'a Vec<Dimension>);
+impl<'a> Hash for WrappedDimensions<'a> {
+    fn hash<H: Hasher>(&self, _state: &mut H) {
+        let _dims = self.0;
```
colin-grapl, line 94: Seems like this is missing something?

src/rust/metric-forwarder/src/accumulate_metrics.rs
wimax-grapl: partition() doesn't work for this use case afaict, since it would be deciding "okay, given a specific entry, like (k,v), does it go into FirstCollection or SecondCollection?" 

basically, what I want is to extract all the keys and all the values, in the same order.

src/rust/metric-forwarder/src/accumulate_metrics.rs
wimax-grapl: oh shit

colin-grapl (approved):

PR #312, opened 2020-09-30
Push 'staging' to docker every time it gets a new push

wimax-grapl:
#231 - this is something I want to get done on the way to "E2E testing for the AWS Plugins repo"

- Adds a new workflow that spits out Docker images to Dockerhub (whenever pushed to staging) with a stable tag: "-staging". 
- Pulls lots of copied code out into etc/build_scripts.

crossing fingers

### DIFF
--- a/.github/workflows/grapl-staging-release.yml
+++ b/.github/workflows/grapl-staging-release.yml
@@ -1,0 +1,172 @@
+name: Grapl Staging Release
+
+on:
+  # Every time we push to staging,
+  # release that to Dockerhub with the tag 'staging'.
+  push:
+    branches:
+    - staging
+
+jobs:
+  release-rust-services:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install dobi
+        run: |
+          wget https://github.com/dnephin/dobi/releases/download/v0.13.0/dobi-linux
+          chmod +x dobi-linux
+
+      - name: Export tag
+        run: |
+          echo "::set-env name=TAG::staging"
+
+      - name: Build Rust services
+        run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/grapl-staging-release.yml
```diff
@@ -0,0 +1,108 @@
...
+          wget https://github.com/dnephin/dobi/releases/download/v0.13.0/dobi-linux
+          chmod +x dobi-linux
+
+      - name: Export tag
+        run: |
+          echo "::set-env name=TAG::staging"
+
+      - name: Build Python services
+        run: |
+          GRAPL_RELEASE_TARGET=release TAG="$TAG" ./dobi-linux --no-bind-mount python
```
wimax-grapl, line 64: wait this seems dumb

jgrillo-grapl: The `TAG="$TAG"` part? I mean.. I don't really have an opinion..

One question I have here is should this workflow also run the tests? I think there's a fair argument either way. On the one hand, in the majority case this workflow will run immediately after a PR merge, and the PR just ran the tests. However, the PR tests were run on the PR branch, not staging post-merge. So there's room for breakage there, potentially. Also, it's possible for an admin to force push to staging, in which case no tests would be run and we'd just blindly build & release to dockerhub.

So I'm not 100% committed to the idea that we need to run tests here but I think there are some reasons it could make sense. If we do run tests we should make sure *none* of the images are published if *any* tests fail.

jgrillo-grapl (approved):
LGTM, one question but it's not a blocker.

wimax-grapl:
I think that's a fair point. I'm going to get this in now (so I can test the behavior) + file a diff after it to do the "test on push" + any breakages I find after I commit it.

PR #318, opened 2020-09-30
Speed up the sqs-lambda buffering when running locally (incl e2e tests)

wimax-grapl:
#231 

Change the SQS Completion Policy from "wait for 10 messages / 3 seconds" to "wait for 1 message / 1 second" for every service that consumes `sqs-lambda`

submitting them to github and looking at how E2E test performance goes 😃


jgrillo-grapl (approved):
👍

PR #329, opened 2020-10-01
Fix release workflow

wimax-grapl:
I needed the 'export tag' step that exists in rust


jgrillo-grapl (changes requested):
this should go to staging

jgrillo-grapl (approved):
👍

PR #342, opened 2020-10-08
Shared library for tests across `grapl` and `grapl-aws-plugins` repos.

wimax-grapl:
#231 

- Add a new library + image, `grapl_tests_common`, for shared test logic across `grapl` and `grapl-aws-plugins` repo (and future ones). 
- Port the existing e2e tests to use this.
- Obviously, this needs to be generalized a lot more (i.e. it only supports sysmon logs right now) _but_ we'd eventually need a lib like this anyway
- Needs a docker image release step

it's all tests 😎


inickles-grapl (approved):
lgtm

Issue #353, opened 2020-10-14
Analyzer executor unable to find downloaded model plugins in local Grapl.


inickles-grapl:
# What version of Grapl are you using?

This affects current staging.

### Is this behavior happening in AWS, locally, or both?

Local only.

### Steps to reproduce the issue

1. Run Grapl local
2. Deploy plugin
3. Deploy analyzer
4. Upload test data

### Expected behavior

I expected my analyzer to run.

### Observed behavior

```
grapl-analyzer-executor_1              | Traceback (most recent call last):
grapl-analyzer-executor_1              |   File "analyzer_executor/src/analyzer-executor.py", line 194, in execute_file
grapl-analyzer-executor_1              |     exec(file, globals())
grapl-analyzer-executor_1              |   File "<string>", line 7, in <module>
grapl-analyzer-executor_1              | ModuleNotFoundError: No module named 'model_plugins.my_model_plugin'
```

### Analysis

The analyzer executor downloads model plugins to `/tmp`, where it expects to find them, as this runs out of `/tmp` in AWS. However, in local Grapl this runs out of `/home/grapl`.

PR #356, opened 2020-10-15
In the analyzer executor, prepend "/tmp" to PATH instead of append

inickles-grapl:

https://github.com/grapl-security/grapl/issues/353

The analyzer executor was appending "/tmp" to its path - this prepends instead and removes the empty model_plugins dir from src.

I tested this locally by deploying the aws plugin, its analyzer, uploaded test events and saw a hit in the engagements. After the commit to parameterize, I rant this test without setting MODEL_PLUGINS_DIR, which defaults to "/tmp", and then again with "/home/grapl" locally.


jgrillo-grapl (approved):
👍

colin-grapl (approved):

wimax-grapl (approved):

PR #357, opened 2020-10-15
Grapl model plugin deployer: return a 401 Unauthorized if must log in

wimax-grapl:
none in particular

instead of a 400
return a 401 for "must log in"

good one, good joke guys


inickles-grapl (approved):
lgtm 👍

PR #373, opened 2020-10-22
Proof-of-concept: split up analyzer_executor into multiple files

wimax-grapl:
none

- fun fact: NONE of our python lambdas are >1 file. while that's _okay_ for now, I want to rip the bandaid off before we end up with 5000 line files
- makes the `analyzer_executor` lib something that can actually be tested now (previously, a filename with a `-` is not a valid Python module and cannot be imported)
- enables us to split files up further in the future
- also, splits apart the entrypoint for the 'run local' stuff versus the lambda handler stuff

- put it on AWS and sent a test event (poorly formatted one). it failed, but not due to import error - instead, due to json stuff
Screenshot 2020-10-22 at 4 22 45 PM
- locally: just gonna ran e2e tests

### DIFF
--- a/dobi.yaml
+++ b/dobi.yaml
@@ -630,6 +630,15 @@ job=run-analyzer-executor-unit-tests:
   use: analyzer-executor-build
   command: /bin/bash -c "source venv/bin/activate && cd analyzer_executor && py.test -n auto -m 'not integration_test'"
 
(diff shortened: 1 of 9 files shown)


jgrillo-grapl (approved):
👍

PR #377, opened 2020-10-24
Osquery support

### LINKED ISSUE #376: OSQuery Support
### What did you want to do?

As a grapl user, I would like to be able to ingest my OSQuery log sources.

### What did you actually end up doing instead?

Having to convert osquery logs into the generic format first / not use osquery log sources.

### What can Grapl do differently?

Natively support OSQuery by building a generator to ingest logs.

d0nut-grapl:

Resolves #376 

Adds support for OSQuery logs from a specially built OSQuery pack for Grapl. Additionally, some bug fixes to make generated subgraphs more accurate.

Integration tests + manual testing with data in `etc/sample_data/osquery_data.log`


src/rust/generators/osquery-subgraph-generator/src/generator.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    async fn handle_event(&mut self, input: Self::InputEvent) -> OutputEvent<Self::OutputEvent, Self::Error> {
+        info!("Processing {} incoming OSQuery log events.", input.len());
+
+        let (subgraphs, errors): (Vec<Option<Graph>>, Vec<Option<failure::Error>>) = input.into_iter()
+            .map(|log| {
+                match Graph::try_from(log) {
+                    Ok(graph) => (Some(graph), None),
+                    Err(e) => {
+                        warn!("Unable to convert partial OSQuery log into subgraph. {}", e);
```
inickles-grapl, line 48: imo this should be `error!`

src/rust/graph-descriptions/build.rs
```diff
@@ -66,7 +66,7 @@ fn main() {
         "#[builder(default)]",
     );
     config.field_attribute(".graph_description.File.file_name", "#[builder(default)]");
-    config.field_attribute(".graph_description.File.file_path", "#[builder(default)]");
+    // config.field_attribute(".graph_description.File.file_path", "#[builder(default)]");
```
inickles-grapl, line 69: should all these comments still be here, or should they be removed?

inickles-grapl (changes requested):
stoked 👍 . had some questions, but there's one spot you def wanna make a change: in cdk, your deployment vals made it in.

src/js/grapl-cdk/bin/grapl-cdk.ts
```diff
@@ -5,7 +5,7 @@ import * as cdk from '@aws-cdk/core';
 import { GraplCdkStack } from '../lib/grapl-cdk-stack';
 import { EngagementUx } from '../lib/engagement';
 
-const deployName = 'Grapl-MYDEPLOYMENT';
+const deployName = 'Grapl-nlattimer';
```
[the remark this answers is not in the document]
d0nut-grapl: lol good catch

src/js/grapl-cdk/lib/grapl-cdk-stack.ts
```diff
@@ -878,6 +925,12 @@ export class GraplCdkStack extends cdk.Stack {
             ...enableMetricsProps,
         });
 
+        new OSQueryGraphGenerator(this, 'osquery-subgraph-generator', {
```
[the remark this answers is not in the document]
d0nut-grapl: Yea i'd prefer to scrub everything together, personally

src/rust/generators/osquery-subgraph-generator/src/generator.rs
d0nut-grapl: Hmm the problem is that we actually expect a ton of these to fire. What happens is that we're parsing every possible osquery log into a generic `PartiallyDeserializedOSQueryLog` (the purpose is to read the log `name` field to determine what kind of log this is). 

We try to convert each of these into a `Graph`. Any log-types that aren't supported (e.g. any of the built-in OSQuery packs) will result in (an expected) failure. After that, some errors could occur if deserialization of supported logs fail or something but this will be less frequent.

src/rust/graph-descriptions/build.rs
d0nut-grapl: Ahh, yea. I'll probably restore these lines because this didn't work the way I wanted it to but I was hoping to get a compile-time error with derive-builder if certain fields weren't set but I forgot this was a runtime check.

src/rust/generators/osquery-subgraph-generator/src/generator.rs
```diff
@@ -0,0 +1,75 @@
...
+                        (None, Some(e))
+                    }
+                }
+            }).unzip();
```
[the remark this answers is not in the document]
d0nut-grapl: Just updated this to use partition and it was much better. Thanks for the suggestion!

src/rust/generators/osquery-subgraph-generator/src/serialization.rs
```diff
@@ -8,22 +8,23 @@ pub struct OSQueryLogDecoder;
...
-            .map(|chunk| {
-                match serde_json::from_slice(chunk) {
-                    Ok(result) => (Some(result), None),
-                    Err(e) => (None, Some(e))
-                }
+            .map(|chunk| match serde_json::from_slice(chunk) {
+                Ok(result) => (Some(result), None),
+                Err(e) => (None, Some(e)),
             })
             .unzip();
```
inickles-grapl: Do wanna do the partition() here as well?

inickles-grapl (approved):

src/rust/generators/osquery-subgraph-generator/src/generator.rs
inickles-grapl: I'm not sure that finding a known and unsupported log type should result in a WARN, maybe not even Result in Error in code.

PR #379, opened 2020-10-26
Add some additional logging to Analyzer Executor. And remove copied `mg_alphas()` code.

wimax-grapl:
none

I don't actually know if analyzer-executor ever goes into an analyzer's `on_response()`. It makes sense to LOGGER.debug() when that happens. 

Aaaaand I also cleaned up the disparate GraphClient stuff because it annoyed me :) 

our beautiful and wonderful test infrastructure

### DIFF
--- a/src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
+++ b/src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
@@ -6,25 +6,23 @@ import logging
 import os
 import sys
 import traceback
-
-
 from collections import defaultdict
 from datetime import datetime
-from multiprocessing import Process, Pipe
+from multiprocessing import Pipe, Process
 from multiprocessing.connection import Connection
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
-from typing import Any, Optional, Tuple, List, Dict, Iterator, Union
+from typing import Any, Dict, List, Optional, Union
 
 import boto3  # type: ignore
 import redis
 from grapl_analyzerlib.analyzer import Analyzer
-from grapl_analyzerlib.execution import ExecutionHit, ExecutionComplete, ExecutionFailed
+from grapl_analyzerlib.grapl_client import GraphClient
+from grapl_analyzerlib.execution import ExecutionComplete, ExecutionFailed, ExecutionHit
 from grapl_analyzerlib.nodes.base import BaseView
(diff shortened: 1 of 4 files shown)


src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -122,24 +111,6 @@ def check_caches(
     return False
 
 
-def handle_result_graphs(analyzer, result_graphs, sender):
```
wimax-grapl, line 125: so yeah fun fact
a) `vulture` is a great lib to detect unused code
b) these are unused

src/python/engagement-creator/src/engagement-creator.py
```diff
@@ -131,6 +130,7 @@ def recalculate_score(lens: LensView) -> int:
 
 
 def set_score(client: GraphClient, uid: str, new_score: int, txn: Any = None) -> None:
```
wimax-grapl, line 132: why the comment instead of just deleting?
cuz like
its really odd
that these fns are just sitting there

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
colin-grapl: sweet

colin-grapl: Ahhh, oh yeah, these used to be helpers for an optimization that I have sadly disabled.

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -171,32 +142,25 @@ def exec_analyzers(
         for an_name, queries in querymap.items():
             analyzer = analyzers[an_name]
 
-            for i, query in enumerate(queries):
+            for _, query in enumerate(queries):
```
colin-grapl, line 145: IF we aren't using 'i' we should just remove the enumerate entirely. This is another case of an optimization that was removed.

src/python/engagement-creator/src/engagement-creator.py
colin-grapl: They could totally be dead tbh

colin-grapl (approved):

Issue #393, opened 2020-11-03
Engagement View broken on local grapl


inickles-grapl:
### What version of Grapl are you using?

Current staging, cae3beb6bf3eeb3218c41b833895ccd863dc0822.

### Is this behavior happening in AWS, locally, or both?

Locally

### Steps to reproduce the issue

`TAG=staging docker-compose up`

### Expected behavior

I expect to find the engagement view at http://localhost:1234/

### Observed behavior

Engagement View container exited.

### Analysis

The docker-compose file launches the engagement-view container with the image default command, as it does not specify one in the service definition: https://github.com/grapl-security/grapl/blob/staging/docker-compose.yml#L340.

The default command for the grapl/grapl-engagement-view image used to be that of the image it was built from (`syntaqx/serve`):
```
CMD ["serve", "-dir", "/var/www"]
```

But the default command grapl/grapl-engagement-view was changed in https://github.com/grapl-security/grapl/commit/d17786ac6d3c9ce9d18c3fee3ebf2a92163cf608#diff-0cc7eece045875bc3efe79f50d49c6dc0e71ec824ff0d2c00674b1abf2f29dfe.

We could either:

1. Revert the default command
2. Override the command to use in the docker-compose file. Something like:

```
command: "serve -dir /var/www"
```

https://github.com/grapl-security/grapl/blob/staging/src/js/engagement_view/Dockerfile#L8
https://hub.docker.com/r/syntaqx/serve/dockerfile

PR #396, opened 2020-11-05
restore cmd to run for engagement-view in local grapl.

inickles-grapl:

https://github.com/grapl-security/grapl/issues/393

This restores the command to run in the engagement-view container for local grapl. We were using the default from the base image, which was recently overwritten in a build stage. This tells docker-compose to use the original command, the original default from base image.

Ran `docker-compose up` and saw the engagement view restored.


wimax-grapl (approved):
we could technically add a `wait-for-it` somewhere that health checks the http server on :1234

but whatever

PR #412, opened 2020-11-10
Dependabot fixes

inickles-grapl:

This fixes a security alert from dependabot: https://github.com/advisories/GHSA-wvhm-4hhf-97x9.

Also, this removes the package update config for staging, we don't care about those, only security fixes.


jgrillo-grapl (approved):
LGTM

PR #415, opened 2020-11-10
Restore master - let's try #401 again

inickles-grapl:

This is an attempt to re-do https://github.com/grapl-security/grapl/pull/401, which was merged via rebasing. Instead, we'd like to merge with commit.


wimax-grapl:
(I believe this is replaced by #419 )

PR #421, opened 2020-11-12
Use Redis locally

alexjmacdonald:

(I don't see one in the public issues list, but we talked about it)

1. Provisions a Redis instance in `docker_compose.yml` for local grapl.  It's currently called "big-dumb-redis" because it's intended to be transitory.  Final goal is to align with production, which is presently architected as 1:1 redis service : grapl component.
2. Rewrites AnalyzerExecutor as a class, with a singleton.  The singleton model minimizes change to calling code, and having it as a class allows some dependency injection, which simplified integration tests.

Once this PR merges, I can move on converting the other components over to using Redis, and writing corresponding integration tests -- patterned after this work.

After that's all done, we can move the one 1:N local Redis instance into N 1:1 Redis instances, so dev resembles prod.  (Or we could move to 1 Redis instance in prod, I don't know the rationale behind each service having its own Redis instance.)

**Hot tip:** Along the way, @wimax-grapl found that Redis instances don't clear their store between integration test runs!!!  Fascinating, and probably a source of future pain.  `docker-compose` kills the Redis backing store via shell, but maybe there is a more elegant and polite way to ask Dockerized Redis to not preserve state across integration tests.

- `docker-compose up` LGTM
- `run-e2e-tests` passes (some flakes, but at least one success)
- unit tests pass
- added `run-analyzer-executor-integration-tests` dobi step, which pass


wimax-grapl:
for the isort, `cd src/python && isort .`

docker-compose.yml
```diff
@@ -30,6 +30,20 @@ services:
         # Hitting :8080/admin/export will force an export to be written to this directory.
         target: /dgraph/export
 
+  # dev uses 1 big redis instance, prod has 1:1 redis per grapl service
+  big-dumb-redis:  # silly name bc temporary
```
wimax-grapl, line 34: temporary eh

alexjmacdonald: why not keep `big-dumb-redis`?  no objections here

(it's temporary bc prod has one redis instance per service, IIRC, and having just one locally is an intermediate step to that goal)

src/python/analyzer_executor/tests/test_cache_integration.py
```diff
@@ -0,0 +1,112 @@
+import unittest
+import pytest
+from faker import Faker
```
wimax-grapl, line 3: it looks like we pulled in `faker` just for random words; you may want to, instead, explore Hypothesis, which we already use in grapl. 

It integrates neatly with pytest fixtures; look for `st.text()` (which is an alias for hypothesis.strategies.text)

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -25,25 +25,24 @@
...
-
-IS_LOCAL = bool(os.environ.get("IS_LOCAL", False))
-IS_RETRY = os.environ["IS_RETRY"]
-
-GRAPL_LOG_LEVEL = os.getenv("GRAPL_LOG_LEVEL")
-LEVEL = "ERROR" if GRAPL_LOG_LEVEL is None else GRAPL_LOG_LEVEL
+# Set up logger (this is for the whole file, including static methods)
 LOGGER = logging.getLogger(__name__)
-LOGGER.setLevel(LEVEL)
+LOGGER.setLevel(os.getenv("GRAPL_LOG_LEVEL", "ERROR"))
```
wimax-grapl, line 30: nice simplification

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -54,157 +53,331 @@ def get(self, key):
...
+class AnalyzerExecutor:
 
-    message_cache = redis.Redis(host=MESSAGECACHE_ADDR, port=MESSAGECACHE_PORT, db=0)
-    hit_cache = redis.Redis(host=HITCACHE_ADDR, port=int(HITCACHE_PORT), db=0)
+    # constants
+    CHUNK_SIZE_RETRY: int = 10
+    CHUNK_SIZE_DEFAULT: int = 100
 
+    # singleton
+    _singleton = None
```
wimax-grapl, line 64: Could we rip out the singleton-ness of this? No real need for it.

- rename `def singleton` to `def create`
- remove the `_singleton` stuff everywhere
- each test case creates its own instance

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -54,157 +53,331 @@ def get(self, key):
...
+            messagecache_addr = os.getenv("MESSAGECACHE_ADDR")
+            messagecache_port = os.getenv("MESSAGECACHE_PORT")
+            LOGGER.debug(f"{messagecache_addr}:{messagecache_port}")
+            try:
+                messagecache_port = int(messagecache_port)
+            except (TypeError, ValueError):
+                LOGGER.error(
+                    f"can't connect to redis, MESSAGECACHE_PORT couldn't cast to int"
+                )
+                messagecache_port = None
```
wimax-grapl, line 98: these sorts of things shouldn't be a log-and-silently-fail, this is a straight up exception. imo, remove the try-except completely

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -54,157 +53,331 @@ def get(self, key):
...
+            else:
+                if bool(messagecache_addr) ^ bool(messagecache_port):
+                    LOGGER.error(
+                        f"message cache falling back to no-op cache (incomplete connection details)"
+                    )
+                else:
+                    LOGGER.debug(
+                        f"message cache falling back to no-op cache (missing connection details)"
+                    )
+                message_cache: EitherCache = NopCache()
```
wimax-grapl, line 116: we should just fail if we can't connect to redis

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -54,157 +53,331 @@ def get(self, key):
...
+    def execute_file(
+        self, name: str, file: str, graph: SubgraphView, sender, msg_id, chunk_size
+    ):
+        try:
+            pool = ThreadPool(processes=4)
 
-def chunker(seq, size):
-    return [seq[pos : pos + size] for pos in range(0, len(seq), size)]
+            exec(file, globals())
+            client = GraphClient()
```
wimax-grapl, line 304: seems like we could just make a `self.client = GraphClient()` in the constructor`

wimax-grapl: since theres at least one other place that constructs a graphclient

src/python/analyzer_executor/tests/test_cache_integration.py
```diff
@@ -0,0 +1,112 @@
...
+@pytest.fixture
+def random_word():
+    fake = Faker()
+
+    def _random_word():
+        return "-".join(fake.words(nb=4))
+
+    return _random_word
+
+
```
wimax-grapl, line 20: Personally:
i really, really hate monkeypatches

i would argue that i'd be happier if `AnalyzerExecutor.get_instance()` didn't inherently interact with `os.`, but instead was like

```
Environment = Mapping[str, Union[str, int]]

def get_instance(environment: Optional[Environment] = None) -> AnalyzerExecutor:
  environment = environment or os.environ
  # now, environment.get("key", default) to your heart's content
```
even if, technically, it exercises less of the true, production codepath

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
alexjmacdonald: The singleton() class method is handy because it condenses all the weird statefulness (pulling from environment, etc.) into one method.  In that sense, it's supposed to be objectionable 😛

This strongly resembles the existing behavior of the flat file.  It moves eager init (everything upon file load) to lazy init (everything upon first singleton() call) while offering dependency injection for testability.  I'm completely on board with improving the design here, though I don't want to go down a garden path of design changes when I'm just trying to get Redis in local.

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
alexjmacdonald: is there _any_ reason someone would want a null cache in prod?  i.e. should the connection info be completely mandatory, or should we allow a fallback NopCache?

src/python/analyzer_executor/tests/test_cache_integration.py
alexjmacdonald: For integration tests, I personally like to black box as much as possible.  To that end, I think my options are either monkeypatching, or rigging the literal OS env itself... and monkeypatching is simpler rn.

I agree with you about the long run though, I think we could stand to have env management in `grapl-common`.  This test straddles unit/integration boundaries and I'd like to have unit-level tests that say "given this config object, do we happy-path to Redis" and integration-level tests that say "give AE does its thing, expect keys XYZ to be inserted into Redis"

How about a ticket to centralize env management as a grapl-common lib, and then refactor the python services to use that?

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
alexjmacdonald: so I tried this and started getting environment errors from GraphClient, I think the environment changes after we start executing within the lambda, so initializing before that is a bad time.  I'll leave this alone for now

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
wimax-grapl: I think the connection should be mandatory - if we can't connect to Redis we have a pretty serious problem with out deployment. 

NopCache should strictly be a unit testing thing.

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
wimax-grapl: I completely agree with the core argument that "this one method is handy because it condenses all the weird statefullness".

I'm saying that, with the singleton, we are indeed introducing a _new_ state - **has this been initialized already?** - that we can completely avoid by, instead, creating a new instance of AnalyzerExecutor for every test.

docker-compose.yml
wimax-grapl: would prefer to rename it something more descriptive like, "shared-redis"

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
alexjmacdonald: I agree with you but I think that's scope creeping past this ticket.  The flat file approach was effectively a singleton; I'm not of the opinion that a singleton is preferable, but _for the core purpose of increasing test coverage and making dev resemble prod_ I want to keep collateral changes scoped down to the bare minimum for implementing tests.

wimax-grapl:
Sure. If you can commit to fixing it afterwards then :+1:

src/python/analyzer_executor/tests/test_cache_integration.py
alexjmacdonald: wrote self a reminder to ticket this (& other follow-on work like de-singleton) next week

wimax-grapl:
yay, typecheck actually found an error! 
```
analyzer_executor/src/lambda_function.py:1: error: Module 'analyzer_executor_lib.analyzer_executor' has no attribute 'lambda_handler_fn'
```
update lambda_function.py to import the singleton and then call that function on it

dobi.yaml
```diff
@@ -615,6 +615,27 @@ job=run-analyzer-deployer-integration-tests:
...
     - integration-env
 
+job=run-analyzer-executor-integration-tests:
+  use: analyzer-executor-build
+  net-mode: grapl-network
+  command: |
+    /bin/bash -c "
+      source venv/bin/activate &&
+      cd analyzer_executor && 
+      export PYTHONPATH="${PYTHONPATH}:$(pwd)/src" &&
```
inickles-grapl, line 625: Looks like the double-quote here escapes and breaks the `command`? Prob want single quotes.

alexjmacdonald: switching the inner quotes to `"`s actually broke this, switching outer quotes to single and inner quotes to double works

wimax-grapl: i am very confused as to how/why this works

wimax-grapl: oh nvm you fixed it

wimax-grapl (approved):

dobi.yaml
alexjmacdonald: I was also very confused why it worked before!  YAML weirdness, or PYTHONPATH is quoted?  I didn't dive into it.

PR #429, opened 2020-11-18
more-robust consumption of secrets manager

wimax-grapl:
This solves Andrea's issue. 
- engagement edge assumes a secretmanager available on 4566 (but that port was never opened)
- engagement edge silently fails if that secretmanager is never made available (it's just in a while-loop of failure, forever)

I've:
- exposed 4566
- made it not silently fail (if it can't retrieve JWT within N seconds, raise an error)
- made it explicitly depend on the user creation in grapl-provision, instead of just wait-forever

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,12 +49,12 @@ services:
     image: localstack/localstack:0.11.5
     environment:
       - PORT_WEB_UI=8099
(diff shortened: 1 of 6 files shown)


jgrillo-grapl (approved):
🔥

PR #431, opened 2020-11-22
Easy py linter command 

wimax-grapl:
none in particular

make isort/black bone simple and not a thin you need to devote any neurons to

locally


jgrillo-grapl (approved):
LGTM

PR #432, opened 2020-11-25
Encrypt EC2 external drives

inickles-grapl:

This sets up disk encryption for external EC2 disks.

Initially tested by running the new commands manually in an EC2 instance. Then I tested these CDK changes with a fresh deployment and SSHing into one of the instances to verify the setup.


jgrillo-grapl (approved):
LGTM

PR #435, opened 2020-11-25
Engagement Edge getNotebook actually returns something when IS_LOCAL; added integration test

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/117
make andrea's life easier when testing locally

previously this would just fail if you were local
now we just send you to :8888

i should write an integration-env test with jwt stuff; in the mean time i wrote a unit test

### DIFF
--- a/src/python/engagement_edge/tests.Dockerfile
+++ b/src/python/engagement_edge/tests.Dockerfile
@@ -1,0 +1,16 @@
+# Motivation for a different `-test` image:
+# I'd like to install stuff into the virtualenv that is only used for tests
+# and quality checks, but not deployed to AWS.
+
+# And you can't define this image in the same Dockerfile, because:
+# dobi evaluates the full Dockerfile when building the first stage of the image; and if
+# grapl-tests-common isn't built yet (as it need not be for engagement-edge-build), it'll fail.
+
+FROM grapl/engagement-edge-build AS engagement-edge-test
+USER grapl
+WORKDIR /home/grapl
+# Steal and install grapl-tests-common
+COPY --from=grapl/grapl-tests-common-python-build /home/grapl/grapl-tests-common grapl-tests-common
+RUN /bin/bash -c "source venv/bin/activate && cd grapl-tests-common && pip install ."
+# Install typechecking stuff
+RUN /bin/bash -c "source venv/bin/activate && cd engagement_edge && pip install .[typecheck]"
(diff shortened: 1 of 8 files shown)


src/python/engagement_edge/tests/test_engagement_edge.py
```diff
@@ -1,15 +1,30 @@
...
+class TestEngagementEdgeClient(unittest.TestCase):
+    def test_get_jwt(self) -> None:
+        client = EngagementEdgeClient(use_docker_links=True)
+        client.get_jwt()
+
+    def test_get_notebook_link(self) -> None:
+        client = EngagementEdgeClient(use_docker_links=True)
+        jwt = client.get_jwt()
+        notebook_url = client.get_notebook(jwt=jwt)
+        assert "localhost:8888" in notebook_url
```
wimax-grapl, line 30: this is that good shit

src/python/grapl_e2e_tests/tests.py
```diff
@@ -26,9 +22,3 @@ def condition() -> bool:
             return length == 4
 
         wait_for_one(WaitForCondition(condition), timeout_secs=240)
-
-
-class TestEngagementEdgeClient(TestCase):
-    def test_engagement_edge_client(self) -> None:
-        client = EngagementEdgeClient(use_docker_links=True)
-        client.get_jwt()
```
wimax-grapl, line 34: some of this was previously in e2e tests, which frankly, only made sense because I expected to have some e2e tests that require the jwt cookie in the future

src/python/engagement_edge/tests.Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Motivation for a different `-test` image:
```
jgrillo-grapl, line 1: We currently install `hypothesis`, `pytest`, `pytest-xdist`, and some other things in the `grapl-python-build` image (and therefore in the virtualenv, and consequently `/site-packages` in the Lambda) of every production service. Would it make sense to make the split at that level too s.t. these deps don't end up in the prod artifacts?

If you unzip one of our deployment artifacts, for example, there are all these test libraries in it. It would be better for each deployment image (and therefore each Lambda artifact) to have only the libraries needed to run the service.

jgrillo-grapl (changes requested):
I think we should push the dependency split up into the base image layer.

src/python/engagement_edge/tests.Dockerfile
wimax-grapl: This is done universally in #444

colin-grapl (approved):

PR #442, opened 2020-11-30
Fix MG_ALPHA for graph merger in cdk.

inickles-grapl:

The value for MG_ALPHAS in AWS has an extra scheme spec. Looks like this is due to swarm/clusterHostPort() also returning scheme.


jgrillo-grapl (approved):
LGTM

PR #446, opened 2020-11-30
Fix github workflows that depend on set-env

wimax-grapl:
https://grapl-internal.slack.com/archives/G017L03AE0L/p1606775942132700
```

Run echo "::set-env name=TAG::staging"
Error: Unable to process command '::set-env name=TAG::staging' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

This is the new way to do that.

### DIFF
--- a/.github/workflows/grapl-release.yml
+++ b/.github/workflows/grapl-release.yml
@@ -124,13 +124,13 @@ jobs:
           else
               CHANNEL="beta"
           fi
-          echo "::set-env name=CHANNEL::$CHANNEL"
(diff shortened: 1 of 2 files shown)


jgrillo-grapl (approved):
:fire:

PR #448, opened 2020-12-01
Fixes for CORS vulnerabilities

d0nut-grapl:

https://github.com/grapl-security/issue-tracker/issues/121

Patches issues in our origin validation logic and failure cases which could allow reading from other origins.

Local validation

### DIFF
--- a/src/python/engagement_edge/tests/test_engagement_edge.py
+++ b/src/python/engagement_edge/tests/test_engagement_edge.py
@@ -10,6 +10,8 @@ JWT_SECRET.secret = "hey im a fake secret"
 class TestEngagementEdgeChalice(unittest.TestCase):
     def test__a_requires_auth_path_fails_without_cookie_headers(self):
         with Client(app) as client:
-            result = client.http.post("/getNotebook")
+            result = client.http.post("/getNotebook", headers={
+                "Origin": "https://local-grapl-engagement-ux-bucket.s3.amazonaws.com"
+            })
             assert result.status_code == 400
             assert result.json_body == {"error": "Must log in"}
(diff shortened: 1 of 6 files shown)


inickles-grapl (approved):

jgrillo-grapl (approved):

src/python/grapl-model-plugin-deployer/src/grapl_model_plugin_deployer.py
```diff
@@ -364,7 +364,7 @@ def upload_plugin(s3_client: BaseClient, key: str, contents: str) -> Optional[Re
 
 
 origin_re = re.compile(
```
wimax-grapl, line 366: typecheck seems to not like this

src/grapl_model_plugin_deployer.py:367: error: invalid syntax

wimax-grapl: suspicion: the {2} here is the culprit

d0nut-grapl: Makes sense! There appeared to be a couple of issues with the regex string (sadly) due to linter and type checker. I'll probably push another commit up if need be.

src/python/engagement_edge/tests/test_engagement_edge.py
```diff
@@ -10,6 +10,11 @@
 class TestEngagementEdgeChalice(unittest.TestCase):
     def test__a_requires_auth_path_fails_without_cookie_headers(self):
         with Client(app) as client:
```
wimax-grapl, line 12: i'm delighted my test caught something

PR #453, opened 2020-12-01
Improve ipynb a bit. Improve documentation about uploading ipynb during provisioning stage.

wimax-grapl:
this is a documentation miss Jesse found during deployment


jgrillo-grapl (approved):
💯

PR #454, opened 2020-12-01
Fix incorrect env name

d0nut-grapl:

https://github.com/grapl-security/issue-tracker/issues/126

Corrects a fatal bug causing the graphql_endpoint service to crash due to an improperly named environment variable. This caused the `prefix` variable to be undefined resulting in an exception when the CORS origin check regular expression was crafted.

Local validation


jgrillo-grapl (approved):
LGTM

PR #458, opened 2020-12-02
Internal 124 upgrade dgraph grpc

andrea-grapl:

https://github.com/grapl-security/issue-tracker/issues/124

- upgraded `grpc`, `dgraph-js` and changed `graphql` endpoint to `graphQlEndpoint/graphql`

We were unable to retrieve a graph because we were calling the wrong graphql endpoint. we recently added a new endpoint to be called for graphql instead

Manually, logged in to local grapl, made sure we could call the new graphql endpoint and retrieve the graph.

### DIFF
--- a/dobi.yaml
+++ b/dobi.yaml
@@ -481,7 +481,7 @@ job=run-node-identifier-integration-tests:
   net-mode: grapl-network
   command: |
     /bin/bash -c "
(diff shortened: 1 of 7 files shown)


wimax-grapl (approved):
would be nice to see some testing, but i know that's still in-flight. did you test this both locally and aws? (don't *need* aws, but would be nice :) )

PR #461, opened 2020-12-03
`grapl::` prelude crate

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/161

Make a one-library-to-rule-them-all to simplify pulling in new versions of stuff


src/rust/grapl-prelude/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+pub use derive_dynamic_node;
+pub use graph_generator_lib;
+pub use grapl_config;
+pub use grapl_graph_descriptions;
```
christophermaier, line 4: This strikes me not so much as a "prelude" (which is a module that publicly exports a bunch of types needed to get meaningful work done in some domain without having to enumerate them all (and basically the only time you use a `*` import). 

Instead, this is more of a meta-package to ensure that you have compatible versions of a bunch of related libraries, correct?

If so,  pulling in all the crates under the `grapl_prelude` namespace helps with keeping your dependencies in line, but doesn't really "save" you anything on the when you end up `use`ing them... it seems to actually make the `use` statements _longer_ by adding `grapl_prelude` to the front.

Perhaps it would be more useful to call this crate `grapl` and then rename the imports so it's more convenient... then you could do things like `use grapl::config::blah;` instead of `use grapl_prelude::grapl_config::blah;`

wimax-grapl: Makes sense. @colin-grapl thoughts?

ajm-grapl (approved):

PR #477, opened 2020-12-08
Internal 126 downgrade grpc

andrea-grapl:

Downgrade grpc to see if updated version introduced bug in retrieving graph

Locally + Manually.


package.json
```diff
@@ -1 +1,6 @@
-{}
+{
+  "dependencies": {
+    "dgraph-js": "^20.3.1",
```
jgrillo-grapl, line 3: We run DGraph version 20.07.1. It's not clear whether the JS client supports this version from the table here: https://github.com/dgraph-io/dgraph-js#supported-versions

Looking at the releases page, the most recent release is 20.03.0: https://github.com/dgraph-io/dgraph-js/releases

Did you find version 20.3.1 published somewhere? It doesn't seem like it should exist...

jgrillo-grapl (changes requested):
Seems like something wrong with the dgraph-js version

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "dgraph-js": "^20.3.1",
+    "dgraph-js": "^20.3.0",
```
jgrillo-grapl, line 3: This should be `20.3.0` not `^20.3.0`

jgrillo-grapl (changes requested):

andrea-grapl:
This was not causing the issue with DGraph. Closing issue.

PR #478, opened 2020-12-09
Subscribe disk use alarms to Operational Alarms

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/126

- makes security / operational alarms mandatory
- gives AlarmSink topics a stable identifier
- subscribes disk_use_percent alarms to that same stable identifier for operational alarm

I tested the `find` function just locally, and it passes my typecheck.
I'm going to kick off a cdk deploy - to my sandbox, not jgrillo's - momentarily...

### DIFF
--- a/src/js/grapl-cdk/swarm/swarm_setup.py
+++ b/src/js/grapl-cdk/swarm/swarm_setup.py
@@ -10,9 +10,9 @@ from typing import TYPE_CHECKING, Any, Iterator, List, NamedTuple, Optional
 import boto3
 
 if TYPE_CHECKING:
-    from mypy_boto3_sns.client import SNSClient
     from mypy_boto3_cloudwatch.client import CloudWatchClient
     from mypy_boto3_cloudwatch.type_defs import MetricTypeDef
+    from mypy_boto3_sns.client import SNSClient
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


src/js/grapl-cdk/swarm/swarm_setup.py
```diff
@@ -80,6 +81,21 @@ def _get_command_result(ssm: Any, command_id: str, instance_id: str) -> str:
 CW_DISK_USAGE_METRIC_NAME = "disk_used_percent"
 
 
+def _find_operational_alarms_arn(
+    prefix: str,
+    sns: Optional[SNSClient] = None,
+) -> str:
```
wimax-grapl, line 87: this is all done because, well, constructing my own ARN is a pain in the ass - how the hell do i find account id and region with boto3, etc

inickles-grapl (approved):

src/js/grapl-cdk/swarm/swarm_setup.py
jgrillo-grapl: I think hitting the instance metadata is the way to go. This is how we do it when we initialize the docker swarm manager: https://github.com/grapl-security/grapl/blob/staging/src/js/grapl-cdk/swarm/swarm_init.py#L7-L12

PR #479, opened 2020-12-10
Fix EngagementView.get_or_create; enable analyzerlib python typechecking in CI

wimax-grapl:
none in particular

Before this we were getting
```
~/venv/lib/python3.7/site-packages/grapl_analyzerlib/nodes/base.py in into_view(self, v)
    133                 graph_client=self.graph_client,
    134                 node_types=self.node_types,
--> 135                 **self.predicates,
    136             )
    137         return None
TypeError: ABCMeta object got multiple values for keyword argument 'node_types'
```
I pop out `node_types` from self.predicates and merge it into self.node_types.

Oh, and I enabled typechecking for analyzerlib in CI :+1:

wrote an integration smoke test for EngagementView.get_or_create

### DIFF
--- a/src/python/grapl_analyzerlib/test2.py
+++ b/src/python/grapl_analyzerlib/test2.py
@@ -1,13 +1,9 @@
 from typing import *
 
-from grapl_analyzerlib.nodes.types import PropertyT
-from grapl_analyzerlib.nodes.viewable import EdgeViewT, ForwardEdgeView
+from grapl_analyzerlib.nodes.viewable import ForwardEdgeView
 from grapl_analyzerlib.nodes.comparators import (
     Cmp,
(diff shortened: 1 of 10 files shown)


src/python/grapl_analyzerlib/test2.py
```diff
@@ -1,13 +1,9 @@
 from typing import *
```
wimax-grapl, line 1: @colin-grapl do we even care about this poorly named test class? wtf is aiud

colin-grapl: We do not. auid refers to `audit user id` though, something that does not exist today.

colin-grapl (approved):

PR #485, opened 2020-12-14
Sqs lambda update

colin-grapl:
None, this is an ad-hoc project to improve some of the costs regarding our lambda executions.

* Updates to the latest sqs-lambda, which allows us to control aspects of lambda execution more easily.
* Latest sqs-lambda fixes a bug where messages may be processed repeatedly
* Reduces the number of retries on each SQS lambda

I'm in the process of testing these. This is a draft PR.


src/rust/Cargo.lock
```diff
@@ -63,9 +63,9 @@ dependencies = [
  "serde_json",
  "sha2 0.8.2",
  "simple_logger",
- "sqs-lambda",
+ "sqs-lambda 0.20.23",
```
inickles-grapl, line 66: I see a number of these for 0.20.23, but they shouldn't they all be 0.21.0?

inickles-grapl (changes requested):

inickles-grapl (approved):

wimax-grapl:
---- test_service stdout ----
thread 'test_service' panicked at 'DYNAMIC_SESSION_TABLE: NotPresent', grapl-config/src/lib.rs:185:51
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

src/rust/grapl-config/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license = "Apache-2.0"
 
 [dependencies]
 log = "0.4.*"
-sqs-lambda = "0.20.*"
+sqs-lambda = "0.21.0"
```
inickles-grapl, line 12: Any particular reason we set patch number for this but not others?

Same for graph-generator-lib

PR #487, opened 2020-12-15
Fix some small bugs with grapl provision and swarm

wimax-grapl:
Grapl Provision ipynb:
- Fix the static typing on it (it skipped functions that had no type annotations 
- Default to py37 kernel
- A biggie - add the Dynamo schema provisioning

Swarm:
- Do the CWAgent daemon start last - my hope is that this fixes the "missing dgraph metric"
- Add a little shortcut for the AWS SSM CLI


etc/Grapl Provision.ipynb
```diff
@@ -149,11 +158,23 @@
     "            schema.add_edge(predicate_meta[\"predicate\"], predicate, \"\")\n",
     "\n",
     "\n",
+    "def store_schema(table, schema: \"BaseSchema\") -> None:\n",
+    "    for f_edge, (_, r_edge) in schema.get_edges().items():\n",
+    "        if not (f_edge and r_edge):\n",
+    "            continue\n",
```
d0nut-grapl, line 164: I wonder if we should add some logging here (just a print statement, perhaps) that says "hey we found an edge that is missing a reverse edge"

I guess we can just add that as-needed if that ever came up again, but it could be a subtle bug in the future that we don't notice until really deep in the debugging process.

d0nut-grapl (approved):
Approved with a potential suggestion to help us avoid head/heartache in the future.

PR #490, opened 2020-12-16
Swarm Setup: do disk alarms first + store export commands on-disk

wimax-grapl:
none

Move the disk alarms up before we ever send off an SSM command.
Because sometimes we'd have a situation where we set up the first node successfully, but the worker nodes' disk alarms couldn't be set up; and running swarm_setup again would fail.

Store export commands on-disk so that future docker stack deploy operations are easier. 

on my aws sandbox

### DIFF
--- a/src/js/grapl-cdk/swarm/swarm_setup.py
+++ b/src/js/grapl-cdk/swarm/swarm_setup.py
@@ -5,7 +5,7 @@ import json
 import logging
 import sys
 import time
-from typing import TYPE_CHECKING, Any, Iterator, List, NamedTuple, Optional
+from typing import TYPE_CHECKING, Any, Iterator, List, NamedTuple, Optional, Sequence
 
 import boto3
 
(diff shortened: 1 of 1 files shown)


src/js/grapl-cdk/swarm/swarm_setup.py
```diff
@@ -239,15 +238,14 @@ def _join_worker_nodes(
     ssm: Any,
     cloudwatch: CloudWatchClient,
     prefix: str,
-    instances: Iterator[InstanceTuple],
+    instances: Sequence[InstanceTuple],
```
ajm-grapl, line 241: is this Sequence bc it's convenient, or is ordering important?  (contrasting to Collection)

jgrillo-grapl (approved):
👍

src/js/grapl-cdk/swarm/swarm_setup.py
wimax-grapl: Sequence is the closest parent of both List and Tuple. Collection also incldues like sets and dict and stuff iirc

PR #493, opened 2020-12-17
Add a DgraphMetricReporter to grapl-observe

colin-grapl:
Adds a new trait for the MetricReporter

This code is pretty trivial, and basically just calls into the MetricReporter with some hardcoded strings, I don't think testing is warranted here.


wimax-grapl (approved):
Approved, assuming you fix the unit tests (they haven't failed yet but I expect them to) and rust lint

PR #496, opened 2020-12-18
Sqs lambda monorepo and metrics

colin-grapl:
https://github.com/grapl-security/grapl/issues/232

* Moves sqs-lambda into the repo.
* Adds metrics to sqs-lambda

I ran localgrapl and validated the metric outputs.


src/rust/sqs-lambda/LICENSE-APACHE
```diff
@@ -0,0 +1,202 @@
+                                 Apache License
```
inickles-grapl, line 1: This file prob not needed with this PR, but it's w/e for now.

src/rust/sqs-lambda/LICENSE-MIT
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2015 The gsblookup-rs Developers
```
inickles-grapl, line 1: Same re prob not needed.

inickles-grapl (approved):
I notice some changes to package locks in src/js, but they seem fine.

PR #499, opened 2020-12-19
Fix issues around consumer policy

colin-grapl:
None, this is a hotfix related to the sqs metrics issue.

Some sanity checks and default-sane behavior improvements in sqs-lambda around how it processes data.

Locally, since this is fixing a bug only present in localgrapl.

### DIFF
--- a/src/rust/generators/graph-generator-lib/src/local.rs
+++ b/src/rust/generators/graph-generator-lib/src/local.rs
@@ -42,11 +42,11 @@ pub(crate) async fn run_graph_generator_local<
 ) {
     let source_queue_url = std::env::var("SOURCE_QUEUE_URL").expect("SOURCE_QUEUE_URL");
 
-    loop {
+    for i in 0u64.. {
         let generator = generator.clone();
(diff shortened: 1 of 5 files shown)


src/rust/generators/graph-generator-lib/src/local.rs
```diff
@@ -42,11 +42,11 @@ pub(crate) async fn run_graph_generator_local<
 ) {
     let source_queue_url = std::env::var("SOURCE_QUEUE_URL").expect("SOURCE_QUEUE_URL");
 
-    loop {
+    for i in 0u64.. {
```
jgrillo-grapl, line 45: is `i` used anywhere?

jgrillo-grapl (approved):
one thing, otherwise lgtm

src/rust/generators/graph-generator-lib/src/local.rs
colin-grapl: Yeah, we Debug log the iteration.

PR #507, opened 2020-12-22
Histogram now supports seconds and microseconds. 

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/132

Now we shoehorn unit data into <tags / dimensions> on the `metric_reporter.py|rs` side
and we optionally pop that data off and consume it on the `metric_forwarder` side, to send the requisite units to Cloudwatch.

I didn't test the reporter side, since it's pretty straightforward piggybacking off of existing, well-tested tag infrastructure.
I did unit test the `metric_forwarder` side.

### DIFF
--- a/src/rust/grapl-observe/src/metric_reporter.rs
+++ b/src/rust/grapl-observe/src/metric_reporter.rs
@@ -19,6 +19,7 @@ pub enum HistogramUnit {
     Millis,
     Micros,
 }
+const RESERVED_UNIT_TAG: &'static str = "_unit";
 
 type NowGetter = fn() -> DateTime<Utc>;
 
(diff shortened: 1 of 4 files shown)


src/rust/grapl-observe/src/metric_reporter.rs
```diff
@@ -110,11 +119,35 @@ where
...
+     * In order to shoehorn units into the statsd protocol, we specify a
+     * special "_unit" tag that will
+     * be popped off in the metric forwarder.
+     */
+    pub fn histogram_with_units(
         &mut self,
         metric_name: &str,
         value: f64,
+        unit: HistogramUnit,
         tags: &[TagPair],
```
colin-grapl, line 140: If we're going to be converting to a Vec anyways we should just take a Vec, or something like:
```
impl Into<Vec<TagPair>>
```
And then instead of `.to_vec` we'd call `.into()`

colin-grapl (approved):

inickles-grapl (approved):

PR #512, opened 2020-12-31
Proof of concept: one-stop-shop for grabbing AWS clients

wimax-grapl:
TODO


wimax-grapl:
deprecated by new sqs executor PR

PR #524, opened 2021-01-07
New build and test system that brings concurrency

inickles-grapl:

This replaces the existing `dobi` build and test system with a new system capable of concurrent builds with `docker buildx bake` and concurrent tests with `docker-compose`.

This reduces the time required to build and test for local development, and drops the time for CI checks to finish from ~50min to ~18min. The time in CI will improve even further when https://github.com/mozilla/sccache/pull/869 lands in sccache, which is slated for the 0.3.0 release, at which point we can use sccache in CI with AWS S3.

There are many notable changes in this PR:

- Fully leverage Dockerfile multi-stage builds. All stages to produce a Docker image exist in a single Dockerfile, allowing us to create any service with a single Docker build command. The number of Dockerfile has been greatly reduced.
- `docker buildx bake` is used to build multiple Docker images in a single BuildKit command, allowing us to leverage BuildKit concurrency. Requires Docker Engine 20.10+.
- Tests are ran with `docker-compose up` to run concurrently. We wrap these executions with a wrapper to check exit codes for each container.
- Introduce a Makefile that defines targets for common uses.
- sccache ready for Rust builds.
- `make zip` target will create zips for AWS deployment, extract script no longer needed.
- pushing to Dockerhub is now done via `docker-compose push`

More information can be found in BUILDING.md

- I've tested each of the Make targets locally, all tests passing, and tests in CI are passing.
- Testing `docker-compose push` was done by running a local registry per https://docs.docker.com/registry/deploying/.


jgrillo-grapl (changes requested):
I like it! Two requests:

1. Let's rip out dobi altogether and replace it with this, no need to run both things together
2. Would be helpful to have how-to docs (maybe in the `CONTRIBUTING.md`?) for some common developer tasks e.g. "add a service", "add unit tests", "add integration tests", "make sure one image builds before another image", etc.

inickles-grapl:
> I like it! Two requests:
> 
> 1. Let's rip out dobi altogether and replace it with this, no need to run both things together
> 2. Would be helpful to have how-to docs (maybe in the `CONTRIBUTING.md`?) for some common developer tasks e.g. "add a service", "add unit tests", "add integration tests", "make sure one image builds before another image", etc.

Cool, yeah that all sounds good. 👍

src/rust/Dockerfile.Makefile
```diff
@@ -0,0 +1,214 @@
...
+ARG SCCACHE_BUCKET
+
+ENV SCCACHE_BUCKET=${SCCACHE_BUCKET}
+ENV SCCACHE_S3_KEY_PREFIX=sccache
+
+SHELL ["/bin/bash", "-c"]
+
+RUN --mount=type=cache,uid=19999,gid=19999,target=/home/grapl/.cargo/registry \
+    --mount=type=cache,uid=19999,gid=19999,target=/home/grapl/sccache \
+    --mount=type=secret,uid=19999,gid=19999,id=aws,dst=/home/grapl/awscreds \
```
wimax-grapl, line 108: (what's up with these hardcoded ids?)

src/rust/Dockerfile.Makefile
```diff
@@ -0,0 +1,214 @@
...
+
+
+# create stage alias for easy reference
+FROM build-sccache-${SCCACHE_LOCATION} AS build
+
+
+# zips
+FROM build AS zips
+
+ENV TAG=latest
```
wimax-grapl, line 127: no longer an argument?

wimax-grapl (commented):
In all honesty I don't have the full background to give this an approve or not, but it seems reasonably thought out. 

Notably, I don't see any e2e tests #justsayin 

if you're like "this PR is big enough already" then i'm down with that, strip out dobi etc in a followup is ok

src/rust/Dockerfile.Makefile
inickles-grapl: > In all honesty I don't have the full background to give this an approve or not, but it seems reasonably thought out.
> 
> Notably, I don't see any e2e tests #justsayin
> 
> if you're like "this PR is big enough already" then i'm down with that, strip out dobi etc in a followup is ok

Originally I was just thinking this would sit side-by-side to dobi while we test it out and get a feel for it. I didn't want to port everything  over yet in case we had other or new ideas for direction or something, but I think it looks promising so I'm open to doing it in just one PR.

inickles-grapl: > (what's up with these hardcoded ids?)

Looked like I couldn't reference a user by name here*. I didn't want to run commands as root and thought running as a UID that probably doesn't map to host would be a good idea, so .. that's what I came up with.

* https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypecache

src/rust/Dockerfile.Makefile
inickles-grapl: It's used in container runs - line 152.

src/rust/Dockerfile
```diff
@@ -1,174 +1,121 @@
...
+
+# copy sources
+COPY . .
 
 #
-# Base image
+# build
 #
+# the `base` target isn't used by the build system, but let's 
+# make `base` and `build` in case it becomes convenient.
```
jgrillo-grapl, line 30: This comment is confusing. I think mostly I'm confused by the word "target". Does that terminology come from referencing an image in the build section of a compose file? In the context of this dockerfile it doesn't really have a clear meaning. Here the `base` image *is* used in L32, so I'm not sure what "used by the build system" means here.

.github/workflows/grapl-build.yml
```diff
@@ -118,24 +119,17 @@ jobs:
...
-          ./dobi-linux run-e2e-tests
-       
+          make test-e2e
+
       # NOTE: This requires >= py37
       - name: 'Collect e2e test artifacts'
         if: ${{ always() }}
         run: |
-          python3 ./etc/ci_scripts/dump_artifacts.py --compose-project "grapl-integration-tests"
+          python3 ./etc/ci_scripts/dump_artifacts.py --compose-project "grapl-integration_tests"
```
wimax-grapl, line 132: the partial-hyphen-partial-underscore seems odd but I'll take your word for it. 
Ideally we'd change it to something less odd if possible.

.github/workflows/grapl-staging-release.yml
```diff
@@ -23,77 +23,14 @@ jobs:
...
-
-      - name: Log in to Docker registry
-        run: |
-          echo "${{ secrets.DOCKERHUB_TOKEN }}" | docker login --username grapl --password-stdin
-
-      - name: Publish JS images to DockerHub
-        run: |
-          docker push grapl/grapl-engagement-view:$TAG
-          docker push grapl/grapl-graphql-endpoint:$TAG
+          docker-compose push
```
wimax-grapl, line 36: man, this actually replaces all that complexity? that rules.

jgrillo-grapl (approved):
These changes look good to me, I left one comment but it's not a blocker.

docs/development/vscode_debugger.rst
```diff
@@ -50,7 +50,7 @@ Run the tests with the following. (Again, this example is strictly about the E2E
 
 .. code-block:: bash
 
-    TAG=latest DEBUG_SERVICES=grapl_e2e_tests dobi run-e2e-tests
+    DEBUG_SERVICES=grapl_e2e_tests make test-e2e
```
wimax-grapl, line 53: BTW if you'd like me to test this still works lmk

etc/ci_scripts/dump_artifacts.py
```diff
@@ -69,7 +68,7 @@ def dump_all_logs(compose_project: str) -> None:
 
 def dump_volume(compose_project: Optional[str], volume_name: str) -> None:
     # Make a temporary container with the volume mounted
-    # dobi adds a prefix on everything in the docker-compose.
+    # docker-compose prefixes volume names with the compose project name.
```
wimax-grapl, line 71: I know we'd discussed just not using volume at all, and `docker cp`'ing from the image or something; where'd we land on that?

test/docker-compose.typecheck-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+
+# environment variable PWD is assumed to be grapl root directory
+
+services:
+
+  #
+  # Python services
+  #
+
+  typecheck-analyzer-executor:
```
wimax-grapl, line 11: We can table this for another later diff, but a thought I had is that

it would be nice if the typecheck were part of a dockerfile RUN command - that way, the result would be cached  and doesn't have to run again - but hey, not a big deal

(and if every python image gets cache-obliterated anyway if a single service in python/ changes, then, hey whatever)

wimax-grapl (approved):

src/rust/Dockerfile.Makefile
```diff
@@ -0,0 +1,214 @@
...
+			       sysmon-subgraph-generator \
+			       osquery-subgraph-generator; \
+    do \
+        grapl-zip "$f" & \
+    done; \
+    wait
+
+CMD for f in $(ls /home/grapl/zips | sed -e 's/.zip$//g'); do \
+      cp -r /home/grapl/zips/${f}.zip ./${f}-${TAG}.zip; \
+    done
```
christophermaier: Might be nice to extract this to a script that can be run (and tested) independently. I worry about burying significant functionality among one of many build stages in a large Dockerfile. `RUN` and `CMD` directives aren't really suited for inlining complex scripts like this.

test/docker-compose-with-error.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+set -e
+
+usage() { echo "Usage: $0 -f docker-compose.yml" 1>&2; exit 1; }
```
christophermaier: `SERVICES` isn't documented here.

Also, it looks like this only accepts a single `-f` argument, correct? If so, that might be good to call out.

test/docker-compose.integration-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+      py.test -n auto -m "integration_test"'
+    environment:
+      - IS_LOCAL=True
+      - BUCKET_PREFIX=local-grapl
+      - UX_BUCKET_URL="ux_bucket_url"
+
+  engagement-view-integration-tests:
+    image: cypress/included:5.6.0
+    working_dir: /test
+    command: --browser chrome --headless # enctrypoint is: cypress run
```
christophermaier, line 94: "enctrypoint" -> "entrypoint"

wimax-grapl:
Oh, e2e tests failed - let's try bumping up the wait-for-it to like 120 or something. In the mean time I'll try to figure out why grapl-provision is so broken

docker-compose.Makefile.yml
```diff
@@ -0,0 +1,562 @@
...
+  grapl-provision:
+    image: grapl/grapl-provision:${TAG:-latest}
+    build: 
+      context: src
+      dockerfile: ./python/Dockerfile.Makefile
+      target: grapl-provision
+    command: |
+      /bin/bash -c "
+        # it doesn't seem to like waiting for sqs
+        # wait-for-it sqs:9324 &&
```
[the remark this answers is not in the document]
jgrillo-grapl: That's a good question. I don't think we ever got to the bottom of that, but it fails reliably if we try to wait for Localstack's SQS.

Some day we will actually build a real healthcheck system, and use that instead. I.e. services themselves should all expose a `/health` endpoint and verify internally that their dependencies are up before reporting healthy.

src/rust/Dockerfile
inickles-grapl: "target" should probably be "stage" here, "target" is terminology slop I brought over from Make.

What I was trying to say is we don't need to make a new stage here. We could remove L31 and rename the first stage to "build".

inickles-grapl: I agree the comment I have there isn't very clear. I'll try to reword.

.github/workflows/grapl-build.yml
inickles-grapl: Agreed, we can change that.

docs/development/vscode_debugger.rst
inickles-grapl: Yes, please. Thanks!

docker-compose.Makefile.yml
```diff
@@ -0,0 +1,562 @@
...
+      - dynamodb:dynamodb
+    extra_hosts:
+      - sqs:127.0.0.1
+      - amazonaws:127.0.0.1
+      - minio:127.0.0.1
+    depends_on:
+      - grapl-master-graph-db
+      - s3
+      - sqs
+      - dynamodb
```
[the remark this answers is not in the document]
jgrillo-grapl: As of compose v3 `depends_on` no longer has any way to tie in a health check. See: https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on
and https://docs.docker.com/compose/startup-order/

etc/ci_scripts/dump_artifacts.py
inickles-grapl: I decided not to touch this. Thought it might have been necessary, but that's not the case and I didn't want unnecessary scope creep.

test/docker-compose.typecheck-tests.yml
inickles-grapl: Yeah, that sounds good to me. For another diff preferably.

src/python/Dockerfile
```diff
@@ -0,0 +1,487 @@
...
+
+COPY --chown=grapl python/swarm-lifecycle-event-handler swarm-lifecycle-event-handler
+
+RUN source venv/bin/activate && \
+    pip install -r swarm-lifecycle-event-handler/requirements.txt
+
+# zip
+FROM swarm-lifecycle-event-handler-build AS swarm-lifecycle-event-handler-zip
+
+RUN ORIG_DIR=$(pwd); \
```
christophermaier, line 391: It'd be more clear to call this something like PACKAGE_NAME rather than ORIG_DIR; the latter primes you to think that we're capturing a directory to ultimately change back into once we're done doing whatever we need to do.

src/rust/Dockerfile
wimax-grapl: Fixed this in next commits

.github/workflows/grapl-build.yml
wimax-grapl: ended up sticking with it - my preferred delimiters `.` and `::` are just ignored by docker

docs/development/vscode_debugger.rst
wimax-grapl: Glad I did, it was slightly broken :)

Makefile
```diff
@@ -0,0 +1,149 @@
...
+	cd src/rust; cargo fmt -- --check
+	./etc/ci_scripts/py_lint.sh --check-only
+
+#
+# else
+#
+
+.PHONY: clean
+clean: ## Prune all docker build cache
+	docker builder prune -a -f
```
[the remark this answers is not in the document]
wimax-grapl: totally agree.

Makefile
```diff
@@ -0,0 +1,149 @@
...
+	docker-compose -f docker-compose.Makefile.yml up --force-recreate -d
+	# save exit code to allow for `make down` in event of test failure
+	test/docker-compose-with-error.sh -f ./test/docker-compose.integration-tests.yml; \
+	EXIT_CODE=$$?; \
+	$(MAKE) down; \
+	exit $$EXIT_CODE
+
+.PHONY: lint
+lint: ## Run lint checks
+	cd src/rust; cargo fmt -- --check
```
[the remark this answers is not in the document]
wimax-grapl: done, thanks

docker-compose.Makefile.yml
```diff
@@ -0,0 +1,562 @@
...
+
+  s3:
+    image: minio/minio
+    command: server /data
+    ports:
+      - "9000:9000"
+    networks:
+      default:
+        aliases:
+          - minio
```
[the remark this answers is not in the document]
wimax-grapl: A great question. This is pulled over from our existing setup, and in the interest of getting this landed, I'm going to file a ticket about it so we can follow up.
https://github.com/grapl-security/issue-tracker/issues/225

docker-compose.Makefile.yml
```diff
@@ -0,0 +1,562 @@
...
+      - "GRAPH_MERGER_QUEUE_URL=http://sqs.us-east-1.amazonaws.com:9324/queue/grapl-graph-merger-queue"
+      - STATIC_MAPPING_TABLE=local-grapl-static_mapping_table
+      - DYNAMIC_SESSION_TABLE=local-grapl-dynamic_session_table
+      - PROCESS_HISTORY_TABLE=local-grapl-process_history_table
+      - FILE_HISTORY_TABLE=local-grapl-file_history_table
+      - INBOUND_CONNECTION_HISTORY_TABLE=local-grapl-inbound_connection_history_table
+      - OUTBOUND_CONNECTION_HISTORY_TABLE=local-grapl-outbound_connection_history_table
+      - NETWORK_CONNECTION_HISTORY_TABLE=local-grapl-network_connection_history_table
+      - IP_CONNECTION_HISTORY_TABLE=local-grapl-ip_connection_history_table
+      - ASSET_ID_MAPPINGS=local-grapl-asset_id_mappings
```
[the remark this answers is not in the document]
wimax-grapl: The `local-grapl` is, in fact, a real value - we put the `BUCKET_PREFIX` (which should have a better name: DEPLOYMENT_NAME) in front of these.

That way, 1 aws account can have, say, multiple `asset_id_mappings` if it hosts multiple deployments.

src/js/graphql_endpoint/Dockerfile.Makefile
```diff
@@ -0,0 +1,54 @@
...
+        grapl
+
+USER grapl
+WORKDIR /home/grapl
+RUN mkdir -p lambda
+
+# Install the dependencies separately to leverage Docker cache
+COPY --chown=grapl package.json lambda/package.json
+COPY --chown=grapl package-lock.json lambda/package-lock.json
+RUN cd lambda && npm i
```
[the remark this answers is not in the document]
wimax-grapl: Done

src/python/Dockerfile.Makefile
```diff
@@ -0,0 +1,496 @@
+#
+# base build
+#
```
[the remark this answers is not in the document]
wimax-grapl: agreed and done

src/python/Dockerfile.Makefile
```diff
@@ -0,0 +1,496 @@
...
+
+# CMD :
+
+#
+# test deps
+#
+
+FROM grapl-python-build AS python-test-deps
+
+COPY --chown=grapl python/python_test_deps python_test_deps
```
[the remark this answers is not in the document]
wimax-grapl: Made an issue: https://github.com/grapl-security/issue-tracker/issues/226

test/docker-compose.integration-tests.yml
wimax-grapl: done

src/python/Dockerfile
wimax-grapl: Sure, yeah, makes sense. I'm going with 'LAMBDA_DIR_NAME' instead of PACKAGE_NAME.

docker-compose.Makefile.yml
```diff
@@ -0,0 +1,562 @@
...
+      - "UX_BUCKET_URL=localhost:3000"
+      - "USER_AUTH_TABLE=local-grapl-user_auth_table"
+      - "BUCKET_PREFIX=local-grapl"
+    depends_on:
+      - grapl-provision
+
+    tty: true
+    links:
+      - grapl-master-graph-db:master_graph
+      - secretsmanager:secretsmanager.us-east-1.amazonaws.com
```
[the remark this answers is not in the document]
jgrillo-grapl: I created an issue for this https://github.com/grapl-security/issue-tracker/issues/228

PR #544, opened 2021-01-16
[Chore] Temporarily allow certain compilation warnings

christophermaier:
To silence our noisy build output for the time being, we will allow
certain classes of compiler lints that are warnings by default.

Eventually, we'll actually _fix_ these violations and allow the lints
to become warnings again. This allows us to have a nicer local build
experience today, while allowing us a structure for fixing the
violations in a principled way, at our own pace.

Signed-off-by: Christopher Maier <chris@graplsecurity.com>

PR #546, opened 2021-01-19
[CI] Ensure Docker builds obey .cargo/config.toml settings

christophermaier:
Previously, `.cargo` was specifically excluded from Docker
builds. This seems to be related to how we use `cargo vendor` builds.

In particular, the output of `cargo vendor` was saved to
`.cargo/config`. This file was the name of the configuration file
prior to Rust 1.39, but the current preferred name is `cargo.toml`. If
both files are present, `config` is used and `config.toml` is ignored.

Because of this, the configuration added in #544 was not being honored
in any of the container-based build systems.

Here, we stop ignoring `.cargo` in builds, allowing us to copy
`.cargo/config.toml` into the container. Then, we append the output of
`cargo vendor` to this file.

For reference, this is what that appended output looks like:

    [source.crates-io]
    replace-with = "vendored-sources"

    [source.vendored-sources]
    directory = "vendor"

The end result is the same vendoring behavior as before, but with
compilation flags honored.

### DIFF
--- a/src/rust/.dockerignore
+++ b/src/rust/.dockerignore
@@ -1,3 +1,2 @@
 target
 */target
-.cargo
--- a/src/rust/Dockerfile
+++ b/src/rust/Dockerfile
@@ -100,7 +100,8 @@ COPY --chown=grapl Cargo.lock .
 COPY --chown=grapl Cargo.toml .
 # vendor all the deps
 RUN mkdir -p /home/grapl/.cargo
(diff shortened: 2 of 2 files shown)


inickles-grapl (approved):
Note: I expect dobi removal is coming soon with https://github.com/grapl-security/grapl/pull/524, but merging is totes fine.

PR #550, opened 2021-01-19
[Chore] Add foundation for Clippy usage on Rust code

christophermaier:

https://github.com/grapl-security/issue-tracker/issues/215

This introduces the use of [Clippy](https://github.com/rust-lang/rust-clippy) for linting Rust code.

Since detailed configuration of the levels of each Clippy lint isn't currently possible, we have to manage it all at the command line. To ensure that we lint consistently everywhere, all that logic is consolidated into a single script (`run-clippy.sh`).

Basic usage is simply calling the script:

    cd src/rust
    bin/run-clippy.sh

However, this script can be integrated into editors to provide linting-on-save functionality. Editor integrations usually operate on
machine-readable output of tools, though, and so this script can be called with a formatting argument for those usecases, e.g.,

   bin/run-clippy.sh json

For example, integrating into Visual Studio Code (assuming usage of the [Rust Analyzer plugin][1]) would look something like this:

    "settings": {
        // ...
        "rust-analyzer.linkedProjects": [
            "src/rust/Cargo.toml"
        ],
        "rust-analyzer.checkOnSave.enable": true,
        "rust-analyzer.checkOnSave.overrideCommand": [
            "/path/to/your/grapl/checkout/src/rust/bin/run-clippy.sh",
            "json"
        ],
        // ...
    }

Additionally, we use this same script in a new CI job.

The currently-allowed lints are the ones that our codebase currently violates. Eventually we would deny most, if not all, of these, along with several others. By allowing them now, however, we start with a "clean slate", providing a consistent foundation for invoking Clippy anywhere, and ensuring that our code always passes Clippy going forward. In other words, it lays the groundwork for grapl-security/issue-tracker#215.

[1]: https://marketplace.visualstudio.com/items?itemName=matklad.rust-analyzer

Run the following, and observe no violations:

    cd src/rust
    bin/run-clippy.sh

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
 src/rust/.cargo @christophermaier
+src/rust/bin @christophermaier
--- a/.github/workflows/grapl-lint.yml
+++ b/.github/workflows/grapl-lint.yml
@@ -8,7 +8,7 @@ on:
       - master
 
 jobs:
-  rust-lint:
+  rust-format:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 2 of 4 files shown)


jgrillo-grapl (approved):
👍

PR #552, opened 2021-01-19
Formalize rustfmt in preparation for more extensive formatting changes

christophermaier:

https://github.com/grapl-security/issue-tracker/issues/216

Similar to the `run-clippy.sh` script added in #550, this commit introduces a `format` script. It is intended to be the source
of truth for how we format our Rust code, regardless of the environment.

In particular, it is designed to work on the terminal and in CI for basic "is the code formatted?" checks, but also in editor
integrations, allowing for format-on-save actions. For example, to configure Visual Studio Code, assuming the use of the [Rust Analyzer Plugin][1], you would do something like the following:

``` json
    "settings": {
        "editor.formatOnSave": true,
        "editor.formatOnPaste": true,
        "[rust]": {
            "editor.defaultFormatter": "matklad.rust-analyzer",
        },
        "rust-analyzer.linkedProjects": [
            "src/rust/Cargo.toml"
        ],
        "rust-analyzer.rustfmt.overrideCommand": [
            "/path/to/your/grapl/checkout/src/rust/bin/format",
            "--editor"
        ],
    }
```
[1]: https://marketplace.visualstudio.com/items?itemName=matklad.rust-analyzer

CI actions were updated to run `rustfmt` based on this script, as well.

Finally, as a prelude to more aggressive formatting changes coming later, we start using nightly Rust only for formatting.

Run the following and observe no failures:
```sh
cd src/rust
bin/format --check
```


jgrillo-grapl (approved):
LGTM

christophermaier:
The `rust-lint` action is showing up as incomplete on this page, but clicking through to look at the details of all actions shows that the job did, in fact, complete successfully. As such, I'm going to go ahead and merge this.

src/rust/bin/run-rustfmt.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
wimax-grapl, line 1: Worth pointing out that `/etc/ci_scripts/` exists, though I'm not opposed to having this in here

src/rust/bin/run-rustfmt.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -euo pipefail
```
wimax-grapl, line 3: the sign of a man who knows his bash

wimax-grapl (approved):

src/rust/bin/run-rustfmt.sh
christophermaier: My thinking is that this is code that is:
a) specifically about Rust code, and
b) not anything that is particularly tied to CI uses.

In the past, I've been on projects where there was a lot of code tied up in the CI tooling that wasn't specifically about CI. We had about 3 different places to execute the same amount of logic, each of which was slightly different from each other. It was more complicated than it needed to be to ensure everything was doing the right thing. Do I have my editor set up to do the same kind of formatting-on-save that our CI system is doing? Am I executing the same lints locally as in CI? And so on.

With this and the clippy work (#550), I'm trying to avoid those pitfalls. In an ideal world, I'd like any CI scripts to be solely concerned with only the particular needs of our CI system, and simply coordinating the execution of any other automation we've set up more generally (like this script).

src/rust/bin/run-rustfmt.sh
christophermaier: Don't leave home without it.

christophermaier:
@wimax-grapl To fix some CI issues, and to take care of some additional automation, I tweaked the script, including renaming it simply `format`, starting in https://github.com/grapl-security/grapl/pull/552/commits/4bb1179c0b2d851dd862ea52456d8069f50cf3e6. Would you mind taking another look? Thanks.

PR #553, opened 2021-01-20
Add more debuggability to accumulate_metrics tests

wimax-grapl:
A transient failure in https://grapl-internal.slack.com/archives/C0176MS6LBY/p1611158138042600 is odd and hard to repro/debug.

These fancier Err statements will help us debug why it fails in the future.

### DIFF
--- a/src/rust/metric-forwarder/src/accumulate_metrics.rs
+++ b/src/rust/metric-forwarder/src/accumulate_metrics.rs
@@ -182,9 +182,9 @@ mod tests {
                 assert_eq!(to_count_map(only_one_datum), expected);
                 Ok(())
             }
-            otherwise => Err(format!(
+            invalid => Err(format!(
                 "Expected exactly 1 metric, but got {:?}",
-                otherwise
+                invalid
             )),
         }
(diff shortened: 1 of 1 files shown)


christophermaier (approved):

colin-grapl (approved):

src/rust/metric-forwarder/src/accumulate_metrics.rs
```diff
@@ -313,7 +317,7 @@ mod tests {
                 );
                 Ok(())
             }
-            _ => Err("must be exactly 2 metrics".to_string()),
+            invalid => Err(format!("Expected exactly 2 metrics, but got {:?}", invalid)),
```
christophermaier, line 320: It occurs to me that it would be a bit more idiomatic to cast these failures in terms of the `panic!` macro:
```rust
invalid => panic!("Expected exactly 2 metrics, but got {:?}", invalid);
```

christophermaier (changes requested):

christophermaier (approved):

PR #554, opened 2021-01-20
Further generalize Rust linting code

christophermaier:

Adds `cargo check` to the existing `run-clippy.sh` script, which has subsequently been renamed to just `lint`. This mainly makes it possible to get even more feedback in your editor to tighten the feedback loop. 

It also fixes an oversight in the clippy invocation where it wasn't checking code for all features, thus letting integration tests escape its keen eye.

Read individual commit messages for further detail.

```sh
cd src/rust
bin/lint
```


wimax-grapl (approved):

wimax-grapl:
Woops, sorry, I know you like to not squash your commits 🤦

PR #556, opened 2021-01-21
Document some code maintenance chores

christophermaier:
Adds a place to begin collecting the rationale for various chores we should be doing on a regular basis to keep our code in shape.

This also serves as a checklist of things to automate in the future.

(Note to reviewers: in part, these instructions describe features that assume that #552 has merged, which is anticipated later today.)

### DIFF
--- a/CHORES.md
+++ b/CHORES.md
@@ -1,0 +1,75 @@
+Chores
+======
+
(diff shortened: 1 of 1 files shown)


inickles-grapl (approved):
Looks good. Thanks for this.

PR #562, opened 2021-01-25
Upload sysmon logs: actually works now, I tested it

wimax-grapl:
Fixes this bug: https://grapl-internal.slack.com/archives/C0176MS6LBY/p1611585771021500

### DIFF
--- a/etc/local_grapl/bin/upload-sysmon-logs.py
+++ b/etc/local_grapl/bin/upload-sysmon-logs.py
@@ -39,6 +39,7 @@ def hack_PATH_to_include_grapl_tests_common() -> Callable:
 def setup_env(bucket_prefix: str):
     if bucket_prefix == "local-grapl":
         kvs = [
+            ("AWS_REGION", "us-east-1"),
             ("S3_ENDPOINT", "http://localhost:9000"),
             ("S3_ACCESS_KEY_ID", "minioadmin"),
             ("S3_ACCESS_KEY_SECRET", "minioadmin"),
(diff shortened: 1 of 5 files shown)


colin-grapl (approved):

etc/local_grapl/bin/upload-sysmon-logs.py
```diff
@@ -38,13 +38,17 @@ def hack_PATH_to_include_grapl_tests_common() -> Callable:
...
+        kvs = [
+            ("S3_ENDPOINT", "http://localhost:9000"),
+            ("S3_ACCESS_KEY_ID", "minioadmin"),
+            ("S3_ACCESS_KEY_SECRET", "minioadmin"),
+            ("SQS_ENDPOINT", "http://localhost:9324"),
+            ("SQS_ACCESS_KEY_ID", "dummy_cred_aws_access_key_id"),
+            ("SQS_ACCESS_KEY_SECRET", "dummy_cred_aws_secret_access_key"),
+        ]
+        for (k, v) in kvs:
+            # fun fact: os.putenv is bad and this is preferred
```
jgrillo-grapl, line 50: why?

wimax-grapl: https://stackoverflow.com/questions/17705419/python-os-environ-os-putenv-usr-bin-env

wimax-grapl: basically 
> Note: Calling putenv() directly does not change os.environ, so it’s better to modify os.environ.
https://docs.python.org/2/library/os.html#os.environ

PR #563, opened 2021-01-25
Remove unnecessary grapl_common dependence in analyzerlib

christophermaier:
None of the code in `grapl_analyzerlib` actually calls anything in
`grapl_common`, so we can remove the bespoke build bits that claim
otherwise.

### DIFF
--- a/src/python/Dockerfile
+++ b/src/python/Dockerfile
@@ -149,6 +149,11 @@ FROM grapl-python-build AS analyzer-executor-build
 COPY --chown=grapl python/analyzer_executor analyzer_executor
 COPY --chown=grapl --from=grapl-analyzerlib-build /home/grapl/venv venv
 
+# HACK: We actually depend on grapl-common, but code structure makes
+# it difficult to express this.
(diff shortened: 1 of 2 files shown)


colin-grapl (approved):

dobi.yaml
```diff
@@ -287,8 +286,8 @@ image=analyzer-executor-test:
 
 image=analyzer-executor:
   image: grapl/grapl-analyzer-executor
-  context: src/python/analyzer_executor/
-  dockerfile: Dockerfile
+  context: src/python/
```
jgrillo-grapl, line 289: I think ideally the context should be scoped to just the files necessary for the service to run (or build, as the case may be). This context includes the entire python source tree which seems excessive. Might be worth adding a `# hack` comment here too.

jgrillo-grapl (approved):
LGTM, one small nit

christophermaier:
This will be taken care of naturally as work progresses on https://github.com/grapl-security/issue-tracker/issues/220.

PR #566, opened 2021-01-28
Notebook fix

inickles-grapl:

This restores the Jupyter notebook config for local Grapl, which broke when switching to the new Dockerfile from https://github.com/grapl-security/grapl/pull/524.

I ran local Grapl and successfully logged in to the UI.


jgrillo-grapl (approved):
LGTM

PR #568, opened 2021-01-30
Sqs fargate final touches -  cypress broken but whyyy

wimax-grapl:
For some reason the cypress container cannot resolve `nginx:1234` and i cannto figure out why
easy way to debug: change that container to
entrypoint: /bin/bash
command: ""

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -440,7 +440,7 @@ services:
       target: grapl-ux-router-deploy
     command: |
       /bin/sh -c '
-        wait-for-it grapl-provision:8126 --timeout=10;
+        wait-for-it grapl-engagement-view-uploader:8127 --timeout=180 &&
         . venv/bin/activate &&
         cd app &&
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -579,7 +579,26 @@ services:
       - s3
       - sqs
       - grapl-provision
-
+      - grapl-graphql-endpoint
+      - grapl-ux-router
+  
+  grapl-engagement-view-uploader:
```
wimax-grapl, line 585: a new service that waits for s3 bucket to exist (courtesy of grapl-provision) and then uploads the engagement view to s3

test/cypress.json
```diff
@@ -1 +1,4 @@
-{}
+{
+    "baseUrl": "http://nginx:1234",
+    "pageLoadTimeout": 180000
+}
```
wimax-grapl, line 4: an attempt at debugging why cypress couldn't connect to nginx:1234

colin-grapl (approved):

PR #579, opened 2021-02-08
Run `cargo audit` job on a schedule

christophermaier:
We don't want failing `cargo audit` jobs to block our regular builds,
and we also don't need them running all the time.

Here, we schedule it to run every week.


jgrillo-grapl (approved):
LGTM

PR #580, opened 2021-02-09
Get higher signal out of the flaky e2e tests - retries, and allow for a semi-failure

wimax-grapl:
None in particular

- Retry the e2e tests 3x.
- Allow for the "lens with scope 3, not 4" case, which means the pipeline is moooostly working.

Basically, the hope is that we'll get fewer false negatives ("IT'S BROKEN BECAUSE OF SOMETHING YOU DID AHHHH!!!!") from these tests, which are relatively useless right now from the standpoint of providing confidence in your PR

creating this PR and looking at its CI


jgrillo-grapl (approved):
Seems legit

PR #581, opened 2021-02-09
Fix cargo audit Github Actions workflow

christophermaier:
TIL strings must be single-quoted in Github Actions


jgrillo-grapl (approved):
😂

PR #582, opened 2021-02-09
Found a CDK Fargate bug: a lot of things mean for ".service, .retryService" were only applied to the ".service"

wimax-grapl:
none

The primary change here is introducing FargateService `taskRoles()` and `connections()`. 
Basically, a lot of consumers accidentally only granted policies/connections to the main service, not the retry service.
These helpers help coerce programmers into granting it to both.

Also, introduced a `dockerfile_paths` because repeated "../../../../" really annoys me lol

I'm deploying as we speak

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@ venv
 .vscode
 .github
 .git
+src/js/grapl-cdk/cdk.out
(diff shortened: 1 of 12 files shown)


wimax-grapl:
trying to confirm in CDK.... taking foreverrrr

PR #585, opened 2021-02-10
Treat DGraph UID values as integers

colin-grapl:
Ultimately this is related to simplifying of our protobuf values, but this is standalone.

We convert the hex encoded uid values to integers, and then, when performing queries involving `uid` values, we hex encode them back to strings.

`make test`


wimax-grapl:
while e2e tests are not required, i'd love if you could get a successful run of it (rerun it once if it fails). it's just a lot of changes

colin-grapl:
> while e2e tests are not required, i'd love if you could get a successful run of it (rerun it once if it fails). it's just a lot of changes

100%

jgrillo-grapl (approved):
LGTM

PR #586, opened 2021-02-10
Initial usage of Pants

christophermaier:
Begins adoption of [Pants](https://pantsbuild.org) for managing Python monorepo code.

This first stage uses Pants just for linting, but testing and building will be added presently.

The overall idea will be to incrementally migrate to Pants, running it and existing release, etc. code side-by-side to facilitate comparison and eliminate large and risky wholesale changeover events. Eventually, old code will be deleted.

For this particular change, there are some small noticeable differences between how Pants is linting and formatting our code and how our existing script is linting, despite the fact that both are using the same `black` and `isort` binaries, as well as the same configuration. I'm not sure why, but it appears that the old linting is an incorrect ordering for `isort` imports. If we're in agreement on that point, I can go ahead and remove the old linting code now (see https://github.com/grapl-security/grapl/pull/586/commits/157d025b8110a8a7cf6fd30edec5c69bb341b561)

Please read commit messages for further details.


jgrillo-grapl (approved):
LGTM

I'm not sure what's going on with the difference in `isort` behavior. @wimax-grapl any ideas?

wimax-grapl:
Will look into isort stuff, but I'm not super attached to any specific settings as long as we have both black and isort :)

wimax-grapl:
once i finish fixing make deploy

3rdparty/python/BUILD
```diff
@@ -0,0 +1,5 @@
+python_requirements(
```
wimax-grapl, line 1: i'm a little grossed out by all this new stuff at the root directory that's more specifically about src/python
but I'm sure we can tweak it in future versions, absolutely not a blocker
🤷

wimax-grapl (approved):
Looks good to me. The isort stuff is fine *as long as* it is stable between `black` and `isort`, which it appears to be.

PR #587, opened 2021-02-10
Fix Rust context for zip creation, which changed recently.

inickles-grapl:

Make target `zip` broke recently with docker context changes. This fixes that.

I can `make deploy`, which first calls the `zip` target, which now succeeds.

### DIFF
--- a/docker-compose.zips.yml
+++ b/docker-compose.zips.yml
@@ -9,7 +9,8 @@ services:
   grapl-rust-zips:
     image: grapl/rust-zip:${TAG:-latest}
     build:
-      context: src/rust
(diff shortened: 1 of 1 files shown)


docker-compose.zips.yml
```diff
@@ -9,7 +9,8 @@ services:
   grapl-rust-zips:
     image: grapl/rust-zip:${TAG:-latest}
     build:
-      context: src/rust
+      context: src
```
jgrillo-grapl, line 12: Is the dockerignore file in `src/` up to date?

christophermaier (approved):

PR #588, opened 2021-02-10
Ensure that Lambda ZIPs build in CI

christophermaier:
Just to ensure we don't inadvertently break the build logic.


jgrillo-grapl (approved):
LGTM

wimax-grapl (approved):

PR #589, opened 2021-02-10
Consolidate the .dockerignore files

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/236

This unifies our .dockerignore files.

I ran a clean build and verified the context size transferred for the Rust build didn't include my `target` directory. Verified by compared sizes.

### DIFF
--- a/src/.dockerignore
+++ b/src/.dockerignore
@@ -1,0 +1,9 @@
+# node modules found in any project
+**/node_modules
+
+rust/target
+
+js/grapl-cdk/zips
+js/grapl-cdk/cdk.out
+
+js/engagement_view/build
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,0 @@
-target
-*/target
-venv
-.idea
-.vscode
-.github
-.git
--- a/src/js/engagement_view/.dockerignore
+++ b/src/js/engagement_view/.dockerignore
@@ -1,3 +1,0 @@
-node_modules
-build
-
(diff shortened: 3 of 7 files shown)


inickles-grapl:
I think I got all the things needed from the deleted files, excluding those that don't seem necessary.

I asked @christophermaier about `src/python/graplinc-grapl-api/.dockerignore`, he said should be fine to just remove.

colin-grapl (approved):

src/.dockerignore
```diff
@@ -0,0 +1,9 @@
+# node modules found in any project
+**/node_modules
+
+rust/target
+
+js/grapl-cdk/zips
+js/grapl-cdk/cdk.out
+
+js/engagement_view/build
```
christophermaier, line 9: It'd probably be good to add some more things from `.gitignore`, like `__pycache__`, `dist`, egg stuff, etc.

inickles-grapl: It looks to me like a lot of the items in `.gitignore` are no longer relevant. Like everything under `# Distribution / packaging`, as all of that is done in Docker. `dist` was used by the dobi system, so that could probably be removed from `.gitignore`, but even still that's a level up from the dir we're using for Docker context.

inickles-grapl: Nothing in `.gitignore` stands out to me as relevant for `src/.dockerignore`, but maybe I'm missing something.

@christophermaier How about I make a ticket to revisit `.gitignore` `src/.dockerignore` and in a separate PR?

christophermaier: @inickles-grapl Yeah, that's fine 👍

PR #590, opened 2021-02-10
Fix CDK re: Rust dockerfiles. Ensure all taskRole/connection stuff applied to both `service` and `retryService`

wimax-grapl:
None. `make deploy` currently broken.

- Fix the build context for rust services
- Ensure all `service.x.y.z.taskRole` and `service.x.y.z.connections` is also applied to retryService

make deploy -> run test data


inickles-grapl (approved):

wimax-grapl:
Okay, to be honest, with a CDK deploy the test data didn't seem to make it past `graph-merger` and I don't know why.

However, I think this is a better situation than the source tree is currently in, so... I'll keep chipping away at it (and deploy a new version asap)

PR #592, opened 2021-02-10
Cleanup .gitignore

inickles-grapl:

In looking at .gitignore for what I should include in the src/.dockerignore, I noticed what looks to me like a lot of items that are no longer necessary. This not only removes such items, it removes duplicates and adds some comments.

I might have removed too much, but I split this PR into many diffs, with comments explaining each change.

I ran `make zip`, which seems to be a fair built for this test, and checked the output of `git status` for new files that shouldn't be checked in. None were found.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,15 +44,6 @@ wheels/
 *.egg
 .mypy_cache
 
-# PyInstaller
-#  Usually these files are written by a python script from a template
-#  before PyInstaller builds the exe, so as to inject date/other infos into it.
-*.manifest
-*.spec
-
-# Installer logs
-pip-log.txt
-
 *.code-workspace
 *.vscode
 *.zip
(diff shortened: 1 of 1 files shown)


ajm-grapl (approved):

christophermaier (approved):

.gitignore
```diff
@@ -1,74 +1,33 @@
...
-# Installer logs
-pip-log.txt
+*.zip
 
+# IDE files
 *.code-workspace
 *.vscode
-*.zip
-stale_outputs_checked
-.log/
```
jgrillo-grapl, line 73: this was important

.gitignore
```diff
@@ -1,74 +1,33 @@
...
-.hypothesis/
```
jgrillo-grapl, line 74: so was this

PR #593, opened 2021-02-11
Fargate cluster names don't have autogenerated ids

wimax-grapl:
None

We were passing an autogenerated id to `serviceName` for FargateService when, in fact, we should just be passing the service name.
Don't worry, it adds the prefix for us :) 

I'm `make deploy`ing as we speak


colin-grapl (approved):

wimax-grapl:
Update: This didn't work?! lol why

PR #595, opened 2021-02-11
Use Pants for Python unit testing

christophermaier:
Here, we tune Pants to be able to run all our Python unit tests. There
are a few points to note:

First, we are still filtering by pytest markers: things that are not
marked as "integration_test" are unit tests. This was done to make as
minimal change in existing code layout and organization as
possible. It does, however, introduce a few wrinkles. Pants will look
across all the Python test files it knows about and then parse them
during the Pytest collection phase. If Pytest finds no tests that
match the markers, that is a kind of test failure. Thus, when running
tests across the entire project, we can get failures simply because a
particular project only has integration tests, for instance. This has
been addressed with the use of a Pytest
plugin (`pytest-custom_exit_code`) that allows us to treat this
situation as a passing test.

Secondly, because we are filtering on pytest markers, and because this
involves loading up Python code in order to read those markers, any
failure at this stage will also be a test failure. As it turns out,
the `test_engagement_edge.py` integration test is written in such a
way that if the `BUCKET_PREFIX` environment variable is not present,
an exception will be raised... at code loading time. Rather than
address this here, I have opted to use a Pants tag on that test (I've
called it "needs_work" for now). Because Pants tags exist at a higher
level than the source code, they can be used to restrict the inputs to
be fed to Pytest in the first place. In the new
`test-unit-python-pants` Makefile target, we actually use both this
Pants tag and Pytest marker arguments for filtering.

Thanks to this work, it was revealed that one unit test we had only worked
if the `IS_LOCAL` environment variable was set. That test was removed. 
Apart from that, however, the only changes I've made to actual test code is
adding `integration_test` markers where they were missing.

Longer term, I believe it would be more flexible to slightly
reorganize the tests to make it easier to target specific classes of
test (i.e. unit vs integration). This could be with specific naming
conventions on files, or separating integration tests into a separate
directory, etc. None of that is done here, though, in the interest of
keeping a light touch.

Aside from those organizational concerns, you can now run all the
existing Python unit tests _without_ Docker, and much faster by
running `make test-unit-python-pants`

Note: this does not remove or change anything in the Docker-based
system for running Python tests (yet). However, it does make it
unnecessary, since Pants runs tests in isolated environments anyway.


src/python/engagement_edge/tests/test_engagement_edge.py
```diff
@@ -8,19 +8,9 @@
 # gross hack because engagement edge is pseudo singleton
 JWT_SECRET.secret = "hey im a fake secret"
 
-
-class TestEngagementEdgeChalice(unittest.TestCase):
-    def test_requires_auth_fails_without_cookie_headers(self):
```
jgrillo-grapl, line 13: This seems like an important test, maybe we should create a bugfix ticket to refactor this code s.t. we can inject the `BUCKET_PREFIX` data dependency at test time? Regressing on this seems like it could be a security issue, but I'm not terribly familiar with this service...

jgrillo-grapl (approved):
LGTM, one comment but not a blocker

src/python/engagement_edge/tests/test_engagement_edge.py
```diff
@@ -8,19 +8,9 @@
 # gross hack because engagement edge is pseudo singleton
 JWT_SECRET.secret = "hey im a fake secret"
 
-
-class TestEngagementEdgeChalice(unittest.TestCase):
```
wimax-grapl, line 12: imo it's a worthy test to have. As long as it's on our plate to resurrect I'm okay with removing it.

(And I actually think that early bailing out of the program if all env vars aren't specified might actually be desirable)

wimax-grapl (approved):

src/python/engagement_edge/tests/test_engagement_edge.py
christophermaier: I agree that bailing out of the _program_ if an environment variable isn't present is a great idea. The problem here is that when this test file is read in during pytest's collection phase, the fact that the environment variable isn't present causes _that_ to throw an exception. It doesn't even get to the point where it can _run_ the tests.

The problem with this particular test is that passing actually required `IS_LOCAL` to be set, so it was really testing _that_ logic, rather than anything real.

That being said, I'll file a ticket to re-create a test that actually exercises the true logic.

christophermaier: Issue added: https://github.com/grapl-security/issue-tracker/issues/244

PR #597, opened 2021-02-11
All docker-compose images are done with build context src/

wimax-grapl:
#589 failed to account for some of our build images which still used different build contexts than `src/`
By removing those `.dockerignore`s, stuff is now breaking **on developers' machines** that might have large build contexts, e.g.
```
 => CANCELED [grapl-cdk-test internal] load build context                                                                                                                          78.1s
 => => transferring context: 3.10GB   (it was getting bigger and bigger due to node_modules not being excluded)
```
Standardizing on src/ will fulfill the initial intent of 589.

some discussion around
https://grapl-internal.slack.com/archives/C0176MS6LBY/p1613083243265500

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -560,8 +560,8 @@ services:
   grapl-engagement-view-uploader:
     image: grapl/grapl-engagement-view:${TAG:-latest}
     build: 
-      context: src/js/engagement_view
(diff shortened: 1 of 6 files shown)


d0nut-grapl (approved):
LGTM thanks!

PR #600, opened 2021-02-12
Use Pants to create Python distributions

christophermaier:
Here, we implement logic to create Python distributions for `graplinc-grapl-api`, `grapl-common`, and `grapl-analyzerlib`, the three libraries we release publicly.

Rather than maintaining separate `setup.py` files for each, we rely on Pants' built-in `python_distribution` target, helped out by a few custom plugins that are introduced here.

First, we intercept the creation of the various keyword-arguments for `setup`. This allows us to ensure consistent metadata across all our distributions (maintainer information, URLs, keywords, etc).

Secondly, we automatically pull in `README.md` files to serve as the `long_description`. If the `BUILD` file that specifies the distribution has a `README.md` as a sibling, we can automatically read its contents. Otherwise, or to be explicit, we can specify a path to the file of our choice with the `readme` key.

To further reduce boilerplate, we use a new `grapl_python_distribution` macro to define our distributions. As it is, this is a very minimal wrapper around the built-in `python_distribution`, but one that ensures that we build an `sdist` and `bdist_wheel` for all our distributions. This is purely a convenience to reduce typing and ensure consistency.

We could (and may yet) implement `grapl_python_distribution` as a completely new Pants target (basically subclassing `python_distribution`), but that isn't yet necessary.

For the time being, we will specify the version of the distributions inline in the Pants targets. However, we could also implement similar file-reading logic to pull that information from `VERSION` files if we choose to. It depends on if other things need to be able to easily access version information.

Another macro, `py_typed` is also added to facilitate the capture of `py.typed` files as Pants `resources` targets for easy inclusion in distributions.

Generation of distributions in our existing Docker build process will be a follow-on task.


jgrillo-grapl (approved):
LGTM

PR #607, opened 2021-02-18
Rust zips (just Metric Forwarder) now use Amazon Linux 2, which provides glibc

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/249

Change the runtime to Amazon Linux 2
We removed musl and now depend on glibc, which `provided` doesnt provide but `provided.al2` does

aws deploy - I saw the lambda running and i saw results showing up in cloudwatch.

### DIFF
--- a/src/js/grapl-cdk/lib/service.ts
+++ b/src/js/grapl-cdk/lib/service.ts
@@ -78,7 +78,8 @@ export class Service {
         const runtime =
             opt && opt.runtime
                 ? opt.runtime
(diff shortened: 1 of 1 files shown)


inickles-grapl (approved):

jgrillo-grapl (approved):
LGTM. IIRC this also means we've bumped python from 3.7 to 3.8, so something to be aware of for Python code.

colin-grapl (approved):

PR #609, opened 2021-02-18
Reorganize docker-compose setup

christophermaier:

https://github.com/grapl-security/issue-tracker/issues/228

This removes links from our Docker Compose setup, in favor of explicit network aliases for all our services. This begins to move us toward a local testing setup that is closer to "real". (It also gets us using more modern docker-compose practices.)

In addition to lots of reorganization, it also extracts a lot of configuration information into an `.env` file, which is automatically pulled into each `docker-compose` run. The information in these environment variables was repeated several times within and across files, so this helps consolidate things. Our `nginx` configuration has been tweaked to also source information from these environment variables.

Some of the accessory testing code still has hard-coded values for some of these variables; rooting all of that out should naturally fall out of the ongoing work to bring our local testing environment more in line with our production deployments.

``` sh
make test-integration
make test-e2e
```


inickles-grapl (changes requested):
Some of us have been using `.env` for specifying things like CDK deployment parameters or build arguments like `CARGO_PROFILE`, often specific to the developer. Having this file ignored by git is great for this.

Instead of having this `.env` file and removing the entry from .gitignore could we still use environment variables with defaults in the docker-compose.yml file? Ex:
```
AWS_DEFAULT_REGION: ${AWS_REGION:-us-east-1}
```

christophermaier:
@inickles-grapl Yeah, I can see how stepping on the toes of other `.env` files would be a pain.

One thing I was trying to achieve here is to centralize information that was duplicated across several compose files (as well as other files), so I'd rather not leave defaults in the various files, since that kind of functions as another kind of duplication.

As a compromise, I can rename the file and then include it in all our `docker-compose` runs with the `--env-file` option; how does that sound to you? Longer-term, I think it would be useful to move some of this stuff into a different directory (`infra`? `local`?) The fact that we've got a lot of "local grapl" code scattered throughout the codebase (this compose file in the root, nginx config and accessory scripts in `etc`, "local provisioning" code in `src/python`, etc.) also argues for a consolidation, IMO. Of course, the end goal would be to reduce as much of this extra "stuff" as possible; I think that's also made a bit easier if everything is obviously located in one place.

inickles-grapl:
> As a compromise, I can rename the file and then include it in all our docker-compose runs with the --env-file option; how does that sound to you?

That sounds fine to me. As an alternative to adding `--env-file` to each docker-compose invocation we could have Make load it similar to how it loads `.env`. So in the Makefile we could have something like:
```
-include .env
-include grapl-local.env
```

Either works for me.

> Longer-term, I think it would be useful to move some of this stuff into a different directory (infra? local?) The fact that we've got a lot of "local grapl" code scattered throughout the codebase (this compose file in the root, nginx config and accessory scripts in etc, "local provisioning" code in src/python, etc.) also argues for a consolidation, IMO. Of course, the end goal would be to reduce as much of this extra "stuff" as possible; I think that's also made a bit easier if everything is obviously located in one place.

Agreed, agreed, agreed.

jgrillo-grapl (approved):
LGTM

christophermaier:
@inickles-grapl I think https://github.com/grapl-security/grapl/pull/609/commits/d1f1766715fdb5dc6a9fb52f1fe96af2d4f94c9d will hopefully balance the contending forces here; let me know what you think!

inickles-grapl (approved):

PR #610, opened 2021-02-18
Ignore orphan containers in targets with multiple compose runs

christophermaier:
When we invoke `docker-compose` multiple times with the same project
name, but different files, we see warnings about the presence of
"orphan containers". Any time `docker-compose` runs and finds services
in a project, but that aren't defined in the file(s) it is currently
operating, it regards them as "orphans".

Since we run two compose runs (once to set up an environment, and once
to execute a series of tests within that same environment), we trigger
this "orphan" detection. In our case, this is expected and not an
issue for concern.

To remove potentially confusing warning messages, we can selectively
ignore orphans in these tasks by setting task-specific environment
variable overrides.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,6 +157,7 @@ test-typecheck: build-test-typecheck ## Build and run typecheck tests
 		-p grapl-typecheck_tests
 
 .PHONY: test-integration
+test-integration: export COMPOSE_IGNORE_ORPHANS=1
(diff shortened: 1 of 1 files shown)


christophermaier:
Documentation link! https://docs.docker.com/compose/reference/envvars/#compose_ignore_orphans

inickles-grapl (approved):

PR #612, opened 2021-02-19
Restore host port binding to containers for local grapl environment

inickles-grapl:

A recent diff accidentally removed the host port binding to the local grapl environment. This restores those mappings.

Additionally, I removed `ports` specs from services that seem like they don't need to be accessible from the host. This is so the host ports won't be bound.

Andrea initially identified in the issue. Making similar changes resolved that for her.

Also, I ran `make test-integration` locally with great success.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,10 +51,7 @@ services:
     tty: false
     image: dgraph/standalone:v20.07.1
     ports:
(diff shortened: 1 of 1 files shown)


jgrillo-grapl (approved):
LGTM

christophermaier (approved):

PR #613, opened 2021-02-19
Dgraph consistency issues

d0nut-grapl:

https://github.com/grapl-security/issue-tracker/issues/152

We converted all of the upsert logic into batch operations. Additionally, changes were made to support batch dynamodb and redis/cache operations. Finally, bug fixes were issued to sysmon generator to ensure that our subgraphs were correctly generated (they weren't before).

Local testing


src/rust/graph-descriptions/Cargo.toml
```diff
@@ -18,8 +18,17 @@ serde_json = "1"
 serde_derive = "1"
 
 derive_builder = "0.9"
+dgraph-tonic = "0.8"
 
 log = "0.4"
+dgraph_query_lib = { git = "https://github.com/d0nutptr/dgraph_query_lib", branch = "main" }
```
colin-grapl, line 24: Can we either vendor this into the monorepo, like in a `/vendored/` directory (ask in slack imo), or get this on crates.io? At minimum we'd want to specify a hash, but I think to get into staging even that would not really be possible.

src/rust/graph-descriptions/src/dynamic_node.rs
```diff
@@ -123,4 +98,60 @@ impl NodeT for DynamicNode {
...
+                Some(node_property::Property::Strprop(s)) => MutationPredicateValue::string(s),
+                None => panic!("Invalid property on DynamicNode: {}", self.node_key),
+            };
+
+            mutation_unit.predicate_ref(key, prop);
+        }
+    }
+
+    fn get_cache_identities_for_predicates(&self) -> Vec<Vec<u8>> {
+        let mut predicate_cache_identities = Vec::new();
```
colin-grapl, line 134: Not gonna block on it, and idc if you fix it at all, but there's `Vec::with_capacity`, which is good for cases where you know how large the Vec will get to, or at least a lower bound. So here I expect it would be `Vec::with_capacity(self.properties.len())`

Again, don't care, just a nice quick win as you write code.

src/rust/graph-descriptions/src/ip_address.rs
```diff
@@ -87,4 +69,61 @@ impl NodeT for IpAddress {
...
+                "{}:{}:{}",
+                self.get_node_key(),
+                "last_seen_timestamp",
+                self.last_seen_timestamp
+            ));
+        }
+
+        predicate_cache_identities
+            .into_iter()
+            .map(|item| item.as_bytes().to_vec())
```
colin-grapl, line 126: `into_bytes` is what you want, and it's free. The code, as written, copies the bytes to a vec but we can just move the bytes right out of the String - and since a String is just a Vec<u8> with some extra guarantees, yeah, free.

src/rust/grapl-utils/src/iter_ext/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+
+pub struct ChunkedIterator<I>
+where
+    I: Iterator + Sized,
+{
+    inner: I,
+    chunk_size: usize,
+}
+
+impl<I> ChunkedIterator<I>
```
colin-grapl, line 22: Is `.chunks` not suitable?

src/rust/grapl-utils/src/rusoto_ext/dynamodb.rs
```diff
@@ -0,0 +1,178 @@
...
+                      BatchGetItemError,
+                      BatchGetItemInput,
+                      BatchGetItemOutput,
+                      DynamoDb,
+                      KeysAndAttributes};
+
+const DYNAMODB_MAX_BATCH_GET_ITEM_SIZE: usize = 100;
+
+#[async_trait]
+pub trait GraplDynamoDbClientExt: DynamoDb {
```
colin-grapl, line 15: Possible to get some metrics in here?

colin-grapl (changes requested):
Did a first pass. Looks good. Only one issue. I'll do a full pass tonight.

src/rust/graph-descriptions/Cargo.toml
d0nut-grapl: Yea I agree with that. I'll put it on crates.io eventually so that'll probably be the move

src/rust/graph-descriptions/src/node.rs
```diff
@@ -850,4 +858,80 @@ impl NodeT for Node {
             }
         }
     }
+
+    fn attach_predicates_to_mutation_unit(&self, mutation_unit: &mut MutationUnit) {
```
ajm-grapl, line 862: My only concern with this PR is the overhead of 2k new lines that are mostly boilerplate.  Would you be OK with letting me know when this PR is out of draft, and I can write up some macros that would lower the LOC?

src/rust/graph-descriptions/src/ip_address.rs
d0nut-grapl: oh perfect

thanks!

ajm-grapl (changes requested):
No issues with correctness, but +2k LOC with no test coverage is a lot of tech debt.  Mind if I work on the derive macro or on adding some unit tests?

src/rust/grapl-utils/src/iter_ext/mod.rs
d0nut-grapl: Nah, chunks only produces an iterator of iterators where the inner iterators reference a `&T` which causes some lifetime problems when invoking the `upsert_graph` method from the graph-merger. It was much easier to just chunk and own the data.

src/rust/graph-descriptions/src/node.rs
d0nut-grapl: I would absolutely love that! Thank you for the help. A derive macro is almost certainly a must for this pattern to continue going forward.

src/rust/grapl-utils/src/rusoto_ext/dynamodb.rs
d0nut-grapl: I would, but I was really hoping to keep this as a drop-in replacement for `batch_get_item` (so, I'd like to avoid adding additional arguments and modifying the return type if possible).

Let me know what you think. Thanks!

colin-grapl: Yeah I think that's fair for now.

colin-grapl (approved):

src/rust/graph-descriptions/src/node.rs
colin-grapl: FWIW once this PR is in I'll be putting another one up that cuts out tons of the code and adds multiple integration tests.

PR #618, opened 2021-02-19
Add #[derive(Eq, PartialEq)] to sqs-excutor::cache::CacheResponse.

inickles-grapl:

Not sure if it's worth creating an issue to track time spent on this PR.

This adds `#[derive(Eq, PartialEq)]` to `sqs-excutor::cache::CacheResponse`. This should enable us to do things like:

```
resp == CacheResponse::Miss
```

I ran `make test-unit-rust`.

### DIFF
--- a/src/rust/sqs-executor/src/cache.rs
+++ b/src/rust/sqs-executor/src/cache.rs
@@ -23,7 +23,7 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


d0nut-grapl (approved):
lgtm

PR #619, opened 2021-02-20
Fix missing ports

d0nut-grapl:

https://github.com/grapl-security/issue-tracker/issues/257

Opens up ports required for SQS and Ratel interface to operate on local grapl

Ran `make up` locally to verify that:

* I was able to upload logs to local sqs using `upload-sysmon-logs.py`
* I was able to access and interact with the ratel interface on port 8000 (which uses port 8080 for running queries)


jgrillo-grapl (approved):
👍

PR #621, opened 2021-02-20
Internal 255 model plugin error

colin-grapl:
https://github.com/grapl-security/issue-tracker/issues/255

* Adds a handler to the model-plugin-deployer for prod/ prefixed urls
* Replaces the `get_s3_client` with the from_env implementation

Tested these locally by listing plugins and observing that it no longer failed.


src/python/grapl-model-plugin-deployer/src/grapl_model_plugin_deployer.py
```diff
@@ -75,6 +75,7 @@
     )["SecretString"]
 
 app = Chalice(app_name="model-plugin-deployer")
+app.debug = True
```
jgrillo-grapl, line 78: what does this do?

jgrillo-grapl (approved):

src/python/grapl-model-plugin-deployer/src/grapl_model_plugin_deployer.py
colin-grapl: Sets up debug logging, but I'll remove this - I'll add it back in a separate PR where it's env-based.

PR #625, opened 2021-02-22
Make Pants aware of Python scripts in grapl-cdk

christophermaier:
And format them too, while we're at it.


jgrillo-grapl (approved):
LGTM, thanks!

PR #628, opened 2021-02-23
Replace most `DgraphClient` / `Local/MasterGraphClient`s in python with `GraphClient`

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/85

Replace most `DgraphClient` / `Local/MasterGraphClient`s in python with `GraphClient`
Not all, because theres one case (Jupyter ipynb) where we can't inject the environment variables required.

e2e tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,6 @@ export
 
 export EVERY_COMPOSE_FILE=-f docker-compose.yml \
 	-f ./test/docker-compose.unit-tests-rust.yml \
-	-f ./test/docker-compose.unit-tests-python.yml \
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -15,7 +15,6 @@ export
 
 export EVERY_COMPOSE_FILE=-f docker-compose.yml \
 	-f ./test/docker-compose.unit-tests-rust.yml \
-	-f ./test/docker-compose.unit-tests-python.yml \
```
wimax-grapl, line 18: just something left over from an old PR I meant to do

colin-grapl (approved):

PR #635, opened 2021-02-24
Internal 280 cypress auth

andrea-grapl:

https://github.com/grapl-security/issue-tracker/issues/280

Adds testing for authentication because we previously didn't have a tests to validate that authentication is working appropriately. 

Using Cypress locally:

- Run `make up`
- `cd test` from the root of grapl
- Change baseUrl from `"baseUrl": "http://api.grapl.test:3128",` `cypress.json` to `baseUrl: http://localhost:1234,`
- Run `npx cypress open`
- Click `login.js` // cypress tests will run

Change URL back to `http://api.grapl.test:3128" & validated Grapl Build integration tests checks pass when submitting PR.

Reference: https://stackoverflow.com/questions/60031340/cypress-login-command-with-cy-request-causes-subsequent-cy-visit-to-fail


test/cypress/integration/login.js
```diff
@@ -0,0 +1,52 @@
+describe("sanity check", () => {
```
wimax-grapl, line 1: probably not needed

test/cypress/integration/login.js
```diff
@@ -0,0 +1,52 @@
...
+	it("passes", () => {
+		expect(true).to.equal(true);
+	});
+});
+
```
wimax-grapl, line 6: could we replace all the tabs here with four spaces, if possible?

test/cypress.json
```diff
@@ -1,4 +1,5 @@
 {
     "baseUrl": "http://api.grapl.test:3128",
-    "pageLoadTimeout": 180000
+    "pageLoadTimeout": 180000, 
+    "chromeWebSecurity": false
```
wimax-grapl, line 4: is this line absolutely needed?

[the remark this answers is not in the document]
ajm-grapl: I was able to get passes with `make test-integration` without this line, FWIW.

test/cypress/integration/login.js
```diff
@@ -0,0 +1,52 @@
...
+			const grapl_jwt = { user: { authenticationData: { token: body.token } } };
+			window.localStorage.setItem("grapl_jwt", JSON.stringify(grapl_jwt));
+		});
+		cy.contains("login").should("not.exist");
+	});
+});
+
+describe("checks that cookie was set after login", () => {
+	it("retrieves grapl_jwt", () => {
+		cy.getCookie("grapl_jwt");
```
ajm-grapl, line 50: I tried this as `cy.getCookie("grapl_jwt").should('exist')` and got a failure, saying the cookie was null.

AFAICT the Cypress API will only fail for unexpected conditions if you use an assertion method, like "should".

ajm-grapl (changes requested):
An explicit assertion on the jwt token causes the test to fail

test/cypress.json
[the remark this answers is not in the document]
andrea-grapl: It's for CORS

PR #640, opened 2021-02-27
change Fargate service LogGroup removal policy to DESTROY

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/285

This changes the LogGroup removal policy to DESTROY. By default the log groups persist after CloudFormation stack removal, which will cause a later deployment of the same name to fail.

Alternative to this solution, it seems we could check `fromLogGroupName` and use that if found before trying to create a new one, but I'm not sure we want these log groups around after removing a stack anyway.

I have the same code for this in a generator in external repo, where I tested this diff. With these changes the issue was resolved.


jgrillo-grapl (approved):
LGTM

inickles-grapl:
> LGTM

🙏

PR #642, opened 2021-03-02
Fix lens scores

colin-grapl:
https://github.com/grapl-security/issue-tracker/issues/229

Sets the score for a lens after calculating it

Tested locally.


wimax-grapl (approved):
would like to see tests

PR #643, opened 2021-03-02
Introduce Pulumi for infrastructure management

christophermaier:
Here, we start using Pulumi for infrastructure, beginning with our testing infrastructure.

Several other cleanups and refactorings of testing infrastructure and code were also done.

Please read the individual commit messages for additional details and background.


docker-compose.yml
```diff
@@ -76,71 +78,55 @@ services:
...
+      # things like the AWS CLI, Pulumi, etc. We'll also re-expose it
+      # for various other services that we might want to otherwise
+      # target explicitly from our workstations.
+      #
+      # You can just stick with the main localstack port, though, and
+      # should get proxied through to where you're trying to go.
+      - 127.0.0.1:${LOCALSTACK_PORT}:${LOCALSTACK_PORT}
+      # As such, these other port exposures can be considered "legacy"
+      # at this point.
+      - 127.0.0.1:${SQS_PORT_EXTERNAL}:${LOCALSTACK_PORT}
```
wimax-grapl, line 117: given this comment - can we remove this second port forward then?

docker-compose.yml
```diff
@@ -503,18 +498,48 @@ services:
...
+         pulumi login --local &&
+         pulumi stack init local-grapl --non-interactive &&
+         # See pulumi/README.md fro why we do these configuration overrides
+         pulumi config set aws:accessKey ${FAKE_AWS_ACCESS_KEY_ID} --plaintext &&
+         pulumi config set aws:secretKey ${FAKE_AWS_SECRET_ACCESS_KEY} --plaintext &&
+         pulumi config set --path 'aws:endpoints[0].s3' ${S3_ENDPOINT} &&
+         pulumi up --yes --skip-preview --stack=local-grapl &&
+         # NOTE: This is running in localstack's network namespace, so
+         # we will need to wait on localstack, NOT grapl-pulumi.
+         python -u -m http.server ${WAIT_PORT}
```
wimax-grapl, line 517: Can you tell me a bit more about how this universal WAIT_PORT works? Is it just, every container exposes :1337 and that's just the magic "wait for it" port? No conflict over the docker network? 

(If so: wonderful)

etc/local_grapl/bin/upload-sysmon-logs.py
```diff
@@ -36,22 +35,6 @@ def hack_PATH_to_include_grapl_tests_common() -> Callable:
     return upload_sysmon_logs
 
 
-def setup_env(bucket_prefix: str):
```
wimax-grapl, line 39: Have you manually tested that this script works?
(It's awful, but as it stands, it needs to work outside of any Docker context)

etc/local_grapl/bin/upload_analyzer_local.sh
```diff
@@ -3,16 +3,16 @@ set -eu
 
 ANALYZER_UPLOAD_SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 LOCAL_GRAPL_DIR=$ANALYZER_UPLOAD_SCRIPT_PATH/../
-export AWS_ACCESS_KEY_ID=minioadmin 
-export AWS_SECRET_ACCESS_KEY=minioadmin 
+export AWS_ACCESS_KEY_ID="${S3_ACCESS_KEY_ID}"
```
wimax-grapl, line 6: Again, has this been manually tested? Could you document "hey, run <some script> that will export S3_ACCESS_KEY_ID etc etc"?

pulumi/Pulumi.local-grapl.yaml
```diff
@@ -0,0 +1,13 @@
+encryptionsalt: v1:KRH5XXw/Vv4=:v1:Mi1EOttYFIuFy8Zl:u/QoTfZuh9c5m6Jyw2L3mAKC+VNs8g==
+config:
+  aws:accessKey: test
+  aws:endpoints:
+    - dynamodb: http://localhost:4566
+      s3: http://localhost:4566
+      secretsmanager: http://localhost:4566
+      sqs: http://localhost:4566
+  aws:region: us-east-1
```
wimax-grapl, line 9: Unrelated to your PR:
I actually wonder if we should have our default region, locally, be:
- a non-default region (like eu-west-1)
- a made up one (like local-region-1)

i feel like we'd do a better job of ensuring all our code is correctly region-aware

pulumi/__main__.py
```diff
@@ -0,0 +1,54 @@
...
+        "unid-subgraphs-generated-bucket",
+    )
+
+    for bucket in buckets:
+        bucket_name = f"{PREFIX}-{bucket}"
+        bucket = aws.s3.Bucket(
+            bucket,
+            bucket=bucket_name,
+        )
+        pulumi.export(f"Bucket: {bucket_name}", bucket.id)
```
wimax-grapl, line 32: shadowing the `bucket` name here with something of a different type really rubs me the wrong way 

maybe `for bucket_name_suffix in buckets:` or something

sidenote: if pulumi is well typed, would love to see typechecking on this!

src/python/engagement_edge/src/engagement_edge.py
```diff
@@ -72,10 +72,12 @@ def _retrieve_jwt_secret_local(self) -> str:
...
-                    region_name="us-east-1",
-                    aws_access_key_id="dummy_cred_aws_access_key_id",
-                    aws_secret_access_key="dummy_cred_aws_secret_access_key",
-                    endpoint_url="http://secretsmanager.us-east-1.amazonaws.com:4584",
+                    region_name=os.environ["AWS_REGION"],
+                    aws_access_key_id=os.environ["SECRETSMANAGER_ACCESS_KEY_ID"],
+                    aws_secret_access_key=os.environ[
+                        "SECRETSMANAGER_ACCESS_KEY_SECRET"
+                    ],
+                    endpoint_url=os.environ["SECRETSMANAGER_ENDPOINT"],
```
wimax-grapl, line 80: Personally I'd prefer to see this done as a new factory in 
src/python/grapl-common/grapl_common/env_helpers.py
but not a blocker on accepting

src/python/grapl-tests-common/grapl_tests_common/clients/engagement_edge_client.py
```diff
@@ -15,7 +15,7 @@ class EngagementEdgeException(Exception):
 
 class EngagementEdgeClient:
     def __init__(self, use_docker_links: bool = False) -> None:
-        hostname = "grapl-engagement-edge" if use_docker_links else "localhost"
+        hostname = "auth.grapl.test" if use_docker_links else "localhost"
```
wimax-grapl, line 18: technically we should probably consume GRAPL_AUTH_HOST here, but, eh, whatever

src/python/grapl-tests-common/grapl_tests_common/dump_dynamodb.py
```diff
@@ -16,10 +17,10 @@
 def dump_dynamodb() -> None:
     dynamodb: DynamoDBServiceResource = boto3.resource(
```
wimax-grapl, line 18: this is something that should definitely use the DynamoDBResourceFactory

src/python/grapl-ux-router/src/grapl_ux_router.py
```diff
@@ -87,9 +88,9 @@ def _retrieve_bucket_local(self) -> Bucket:
             try:
                 s3 = boto3.resource(
```
wimax-grapl, line 89: If grapl_ux_router already consumes grapl-common, would be nice to use S3ResourceFactory.
Otherwise, whatever, it's okay

src/python/grapl_provision/grapl_provision.py
```diff
@@ -444,43 +249,4 @@ def sqs_provision_loop() -> None:
...
-            if i >= 50:
-                LOGGER.error(e)
-            time.sleep(1)
-
-    LOGGER.info("Ensuring S3/SQS completed...")
-    sqs_t.join(timeout=300)
-    s3_t.join(timeout=300)
-    LOGGER.info("S3/SQS completed")
-
     LOGGER.info("Completed provisioning")
```
wimax-grapl, line 252: beautiful PR

wimax-grapl (commented):
generally looks good, but i have a few nitpicks :)

pulumi/Pulumi.local-grapl.yaml
jgrillo-grapl: IIRC localstack is very particular about regions

jgrillo-grapl (approved):
LGTM

docker-compose.yml
christophermaier: Yup, basically. Just like you can have multiple servers on the same network that expose a service on the same port, you can have multiple containers do the same thing because they each have their own network namespace by default.

(In this particular instance, however, the `grapl-pulumi` container happens to intentionally share the network namespace of the `localstack` container, which is why we wait on `localstack:${WAIT_PORT}`.)

I basically created the `WAIT_PORT` variable so we didn't have to manually track lots of different magic port numbers.

docker-compose.yml
christophermaier: If nobody is using it, sure. I'll verify that today.

pulumi/Pulumi.local-grapl.yaml
christophermaier: The documentation suggests we can set `DEFAULT_REGION` when starting Localstack and set it to whatever we want. I can try that and see if everything still works. If so, it could be a nice way to ensure that we don't accidentally bake in `us-east-1` assumptions in our services.

pulumi/__main__.py
christophermaier: Ah, yes... I forgot to add typing; I'll take care of that!

src/python/engagement_edge/src/engagement_edge.py
christophermaier: Yeah, that's a good idea.

src/python/grapl-tests-common/grapl_tests_common/clients/engagement_edge_client.py
christophermaier: Fair point.

Reading it again, the `use_docker_links` parameter name doesn't really fit well anymore, either. 🤔

src/python/grapl-tests-common/grapl_tests_common/dump_dynamodb.py
christophermaier: Good catch!

src/python/grapl-ux-router/src/grapl_ux_router.py
christophermaier: I'll check on that.

src/python/grapl_provision/grapl_provision.py
christophermaier: Deleting code is fun 🎉

pulumi/__main__.py
christophermaier: Also, this isn't directly related to your question, but the question of naming reminded me of it... what value do we get from naming a bucket (and all the other resources we deal with) with a given prefix? I ask because controlling the name that tightly (i.e., conflating logical names and physical names) can have complicating implications.

- https://www.pulumi.com/blog/infrastructure-as-code-resource-naming/
- https://www.pulumi.com/docs/intro/concepts/resources/#names

This isn't something to be addressed for this PR, but was something I was curious about.

docker-compose.yml
christophermaier: It's not required, so we can remove it entirely.

pulumi/Pulumi.local-grapl.yaml
christophermaier: I think this may work, but we do have `us-east-1` in a handful of places in the code, and not all of it is testing / local grapl code.

I'll leave the code as-is in this PR, but will file an issue to weed this out of the codebase.

etc/local_grapl/bin/upload-sysmon-logs.py
christophermaier: @wimax-grapl Can you tell me a bit more about the intended usecases for this script? The `setup_env` function here only did anything when dealing with the `local-grapl` prefix, so using it against real AWS wouldn't be affected.

Are you saying that this script needs to be able to run against local Grapl from outside the docker-compose environment?

wimax-grapl (approved):

PR #657, opened 2021-03-04
Bucket prefix is now called deployment name

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/125

- Rename every bucketPrefix BUCKET_PREFIX etc to deploymentName DEPLOYMENT_NAME etc
- Fix a screwup where we added too many http://'s to MG_ALPHAS in rust
- 

All the way up through AWS.
Messages made their way up through engagement creator but tbh i was unable to get a lens


wimax-grapl (commented):
left some guidance

inickles-grapl (approved):

PR #660, opened 2021-03-05
Refactor log levels in CDK; introduce a Deployment Name regex

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/318

- Deployment names need to be normalized. No caps, no unicode, just lowercase and numbers and - and _ 
- but for people with existing deployments, we do have a legacy flag
- In testing this, I decided to refactor the crap out of log levels and Deployment Parameters (because, at test time, it expects a process.env.STUFF)

added tests

### DIFF
--- a/src/js/grapl-cdk/lib/services/model_plugin_deployer.ts
+++ b/src/js/grapl-cdk/lib/services/model_plugin_deployer.ts
@@ -55,7 +55,7 @@ export class ModelPluginDeployer extends cdk.NestedStack {
             ),
             vpc: props.vpc,
             environment: {
-                GRAPL_LOG_LEVEL: props.defaultLogLevel,
+                GRAPL_LOG_LEVEL: props.logLevels.defaultLogLevel,
                 MG_ALPHAS: props.dgraphSwarmCluster.alphaHostPort(),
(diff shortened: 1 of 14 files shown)


src/js/grapl-cdk/lib/services/model_plugin_deployer.ts
```diff
@@ -55,7 +55,7 @@ export class ModelPluginDeployer extends cdk.NestedStack {
             ),
             vpc: props.vpc,
             environment: {
-                GRAPL_LOG_LEVEL: props.defaultLogLevel,
+                GRAPL_LOG_LEVEL: props.logLevels.defaultLogLevel,
```
colin-grapl, line 58: Should we have a specific log level for this service?

colin-grapl (approved):

src/js/grapl-cdk/lib/services/model_plugin_deployer.ts
wimax-grapl: i mean probably but that wasnt the original intent of this PR lol

d0nut-grapl (approved):
Looks good :D

PR #666, opened 2021-03-08
Manage our virtualenv better

christophermaier:

Provides a better way to manage a virtual environment of 3rd-party dependencies for Grapl. Rather than simply creating a virtualenv for the purposes of managing our `constraints.txt` file, we'll now expose it as an environment for proper usage by developers.

The first benefit of this is that we no longer need to manage Pulumi's virtualenv manually.

```sh
make populate-venv
source build-support/grapl-venv
```
OR
```sh
build-support/manage_virtualenv.sh populate
source build-support/grapl-venv
```
OR
```sh
build-support/manage_virtualenv.sh regenerate-constraints
```


Dockerfile.pulumi
```diff
@@ -0,0 +1,64 @@
...
+        grapl
+
+USER grapl
+ENV USER=grapl
+WORKDIR /home/grapl
+
+RUN curl -O https://get.pulumi.com/releases/sdk/pulumi-v2.21.2-linux-x64.tar.gz && \
+    tar -xzvf pulumi-*.tar.gz && \
+    mv pulumi pulumi-bin
+ENV PATH=/home/grapl/pulumi-bin:${PATH}
```
inickles-grapl, line 25: Any particular reason we don't use a Pulumi image from DockerHub like pulumi/pulumi:v2.21.2 ?

colin-grapl (approved):

.dockerignore
```diff
@@ -0,0 +1,12 @@
+build-support/grapl-venv
```
inickles-grapl, line 1: Looks like this moves the .dockerignore file, which I think means all other build contexts need updating, right? In order to use it that is, since it's only read from the root of a build context?

inickles-grapl (changes requested):

.dockerignore
inickles-grapl: err nevermind, for some reason i thought i saw src/.dockerignore was removed, but that's not the case.

inickles-grapl (approved):

Dockerfile.pulumi
christophermaier: Since installing Pulumi is the least of what this Dockerfile is doing, I figured it was more important to align with our other Dockerfile.

PR #671, opened 2021-03-09
Internal 286 lens hdr merg

andrea-grapl:

https://github.com/grapl-security/issue-tracker/issues/286

In addition to resolving the duplicate lens header bug, the following items were also addressed in this ticket:

Minor Restyling
Sidebar file restructuring
Resolve Material UI / react-scripts update warnings
Correct errors from improperly nested lens table header
Hide nested objects from node table data which caused the entire app to crash.

The purpose of this PR was to resolve various errors manifesting from our table structure, causing issues elsewhere. 

Locally, because this is a purely styling/bug fix issue - no functionality was changed as a part of this PR which means this shouldn't affect AWS functionality.

Ran `./build.sh` before final push to ensure files were properly built. 

Before: Run localgrapl + upload an analyzer with two lens names 

**Crashing issue:** 
1 - Click on lens name 
2 - Click on node 
3 - Click on lens name again 
4 - App should not crash 

**Lens Header Duplication:**
In `grapl/etc/local_grapl/suspicious_svchost/main.py`
1 -  We should only see the lens header once and a list of lenses. 

image


colin-grapl (approved):

3rdparty/python/constraints.txt
```diff
@@ -55,3 +55,4 @@ urllib3==1.26.3
 wcwidth==0.2.5
 wheel==0.36.2
 zstd==1.4.8.1
+
```
christophermaier, line 58: For the record, this was at my suggestion... I think we're hitting some weird caching edge-cases in Github Actions that appear to cause builds to fail. This effectively clears the cache, causing it to be regenerated anew. It sucks, but it's what we have until there's a UI to manage caches for Github Actions 😭

PR #675, opened 2021-03-10
Initial Pulumi / CDK Harmonization

christophermaier:

https://github.com/grapl-security/issue-tracker/issues/293

Brings our existing Pulumi code (which was based on code from `grapl_provision` for use with Local Grapl / testing environments) in line with what we actually create in AWS using CDK. We had a few mis-named buckets and queues, as well as some configuration options, that were out-of-sync with our real infrastructure.

A few things we're doing in AWS don't work (yet?) in our Localstack environment (like S3 bucket website configuration and SSE encryption for buckets). For now, we'll simply leave those bits out when we're deploying Local Grapl.

One notable difference is that the real infrastructure is tagged. We've now got some tooling to universally apply tags across all the infrastructure objects we create; define the tags in one place and all the infrastructure gets it.

To toggle between creating infrastructure and importing existing infrastructure (which we'll need to do at some point as we transition off of CDK), I've added a stack configuration flag that will globally toggle between "create" and "import" mode. When in "import" mode, we can run `pulumi up` against existing infrastructure and compare what exists with what we've coded.

To check all this, I deployed to my sandbox in AWS using CDK, and then performed `pulumi up` in import mode in a new personal sandbox stack pointed to my AWS instance. This gives a diff of what Pulumi would generate compared to what exists already, allowing me to iteratively bring our Pulumi code in line with the CDK infrastructure. 

In addition, I regularly ran `make test-integration` and `make test-e2e` locally to ensure the Pulumi code continued to work locally.


pulumi/infra/ui.py
```diff
@@ -0,0 +1,39 @@
...
+
+import pulumi_aws as aws
+from infra import util
+from infra.util import IS_LOCAL
+
+import pulumi
+
+
+class UI(pulumi.ComponentResource):
+    """ Represents the web GUI for Grapl."""
```
wimax-grapl, line 11: We have a lot of terms for this - Engagement Edge, Engagement UX, etc - and I'd prefer to keep it at 2 instead of 3 if possible!

pulumi/infra/util.py
```diff
@@ -0,0 +1,88 @@
...
+    import_from_existing = pulumi.Config().require_bool("import_from_existing")
+
+    given_opts = pulumi.ResourceOptions(**kwargs)
+    import_opts = pulumi.ResourceOptions(
+        import_=resource_id if import_from_existing else None
+    )
+    return pulumi.ResourceOptions.merge(given_opts, import_opts)
+
+
+def grapl_bucket(
```
wimax-grapl, line 44: Note: our CDK buckets auto-delete objects upon destroy; does that still work here?
https://github.com/grapl-security/issue-tracker/issues/135

pulumi/infra/util.py
```diff
@@ -0,0 +1,88 @@
...
+    logical_bucket_name: str,
+    prefix: str = pulumi.get_stack(),
```
wimax-grapl, line 46: again, previous comment about prefix

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -444,10 +444,13 @@ def emit_event(s3: S3ServiceResource, event: ExecutionHit, is_local: bool) -> No
     # TODO fargate: always emit manual events
     if is_local:
         # Local = manual eventing
+
+        deployment_name = os.environ["DEPLOYMENT_NAME"]
```
wimax-grapl, line 448: might be a candidate for grapl_common `env_helpers.py` perhaps

pulumi/infra/emitter.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __init__(
+        self, event_name: str, opts: Optional[pulumi.ResourceOptions] = None
+    ) -> None:
+
+        super().__init__("grapl:EventEmitter", event_name, None, opts)
+
+        logical_bucket_name = f"{event_name}-bucket"
+        self.bucket = util.grapl_bucket(logical_bucket_name, sse=True, parent=self)
+
```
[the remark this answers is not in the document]
wimax-grapl: edit: looks like grapl_bucket does this automatically

pulumi/infra/service_queue.py
```diff
@@ -19,45 +21,72 @@ def __init__(
...
-                    "maxReceiveCount": 10,
+                    "maxReceiveCount": 3,
                 }
             )
 
         # TODO: delete_before_replace is only needed if we're
         # overriding the name of the queues
 
-        dead_letter_name = f"grapl-{name}-dead_letter-queue"
+        prefix = pulumi.get_stack()
```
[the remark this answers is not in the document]
christophermaier: Ah, crap... this was code I wrote before your PR, and I forgot to rename after rebasing... I'll fix this up; thanks for catching it.

pulumi/__main__.py
```diff
@@ -1,5 +1,6 @@
 import pulumi_aws as aws
-from infra import dynamodb
+from infra import dynamodb, emitter, ui, util
```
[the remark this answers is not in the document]
christophermaier: Not my favorite either... any suggestions?

pulumi/infra/ui.py
christophermaier: I'll call it `EngagementUX`.

pulumi/infra/util.py
christophermaier: I have to add one more flag, but yes, we can do that here, too.

pulumi/infra/service_queue.py
```diff
@@ -19,45 +21,72 @@ def __init__(
...
-        queue_name = f"grapl-{name}-queue"
+        logical_queue_name = f"{name}-queue"
+        physical_queue_name = f"{prefix}-{logical_queue_name}"
         self.queue = aws.sqs.Queue(
-            queue_name,
-            name=queue_name,
+            logical_queue_name,
+            name=physical_queue_name,
             message_retention_seconds=message_retention_seconds,
+            visibility_timeout_seconds=180,
```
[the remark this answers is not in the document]
christophermaier: I'll file an issue to follow up on this, since it doesn't seem like there's a straightforward place in that code where that assumption is encoded. These numbers are basically coming straight from the CDK code as it exists now.

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
christophermaier: Good idea... I'll file an issue for this as well.

pulumi/infra/service_queue.py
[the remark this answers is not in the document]
christophermaier: https://github.com/grapl-security/issue-tracker/issues/333

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
christophermaier: https://github.com/grapl-security/issue-tracker/issues/334

pulumi/infra/ui.py
jgrillo-grapl: It wouldn't be a bad idea to revisit the Grapl UI name at some point. The word "Engagement" is pretty dated, the UI does a lot more than just Engagements. Created a ticket for this as well: https://github.com/grapl-security/issue-tracker/issues/335

wimax-grapl (approved):

PR #679, opened 2021-03-12
Remove MinIO in favor of Localstack's S3 implementation

christophermaier:

In order to more closely mimic AWS in our testing environments, we need a way to implement our bucket notifications to SNS topics. As long as we are using MinIO, we can't do that.

This PR replaces MinIO with Localstack's built-in S3 implementation, which should allow this. This removes testing infrastructure, and also allows us to start removing a bunch of special-casing in our code, starting with the website bucket configuration logic in our Pulumi code, as well as modifying certain Pulumi stack configuration values.

One interesting aspect of this was that we needed to strip the leading `/` from the S3 object keys (derived from URL paths) in our `grapl-ux-router` code. Apparently, MinIO was able to deal with finding an object named "/static/js/2.642cd0fc.chunk.js", whereas Localstack is more strict and wants this to be simply "static/js/2.642cd0fc.chunk.js" (this appears to be more proper anyway). The only code changed here was explicitly "local only"

``` sh
make test-integration
make test-e2e
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -85,24 +85,6 @@ services:
         aliases:
           - ${REDIS_HOST}
 
-  # While localstack does offer S3 capabilities, it's not actually
-  # MinIO, which is a very full-featured and faithful S3 clone. We
-  # have noticed test failures when we use localstack's s3, so we will
-  # continue to use MinIO.
-  s3:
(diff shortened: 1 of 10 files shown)


jgrillo-grapl (approved):
LGTM

PR #682, opened 2021-03-15
Internal 256 rename fetch

andrea-grapl:

grapl-security/issue-tracker#256

Rename “apiFetch” to “apiFetchWithBody” and “apiFetchWithBody” to “apiPostWithBody” per request by @colin-grapl  for clarity. 

Locally - this is just a variable name change so no functionality was impacted.


src/js/engagement_view/src/services/fetch.ts
```diff
@@ -14,7 +14,7 @@ export const apiFetch = async (urlSlug: string, method = "GET") => {
 };
 
 
-export const apiFetchWithBody = async (urlSlug: string, method = "POST", body: string) => {
+export const apiPostRequestWithBody = async (urlSlug: string, method = "POST", body: string) => {
```
wimax-grapl, line 17: Everywhere you call apiPostRequestWithBody you specify "post"; why not just remove this param? It seems encoded in the method name now.

src/js/engagement_view/src/services/fetch.ts
```diff
@@ -1,4 +1,4 @@
-export const apiFetch = async (urlSlug: string, method = "GET") => {
+export const apiFetchWithBody = async (urlSlug: string, method = "GET") => {
```
wimax-grapl, line 1: A) There's no "body" here, hence `withBody` doesnt make sense
B) There's a few places where we 
apiFetchWithBody
with POST not GET - in which case - why not just apiPostRequestWithBody in those cases?

src/js/engagement_view/src/services/fetch.ts
andrea-grapl: Good idea, I think the rationale was to use the param in case we ever wanted to call a different method, but you're totally right, this doesn't make sense for the name - especially when we only support GET + POST.

src/js/engagement_view/src/services/fetch.ts
andrea-grapl: Fixed

wimax-grapl (approved):

PR #686, opened 2021-03-17
Remove check-pypi check, since we do pypi updates manually now

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/346

Remove part of our lint workflow that doesn't provide value. 
It is an artifact of when our grapl releases did pypi automatically.

it's github actions, so, this CI


colin-grapl (approved):

jgrillo-grapl (approved):
💯

PR #688, opened 2021-03-18
Added a graphql endpoint test to e2e tests, just so we have *some* coverage on it

wimax-grapl:
None yet.

- Add a rudimentary GraphqlEndpointClient to grapl-tests-common
- Add a rudimentary test that, yes, it loads lenses okay
- Remove some `use_docker_links=` stuff that's no longer applicable

I would like to have integration tests on it within the graphql endpoint source tree, but right now, this is a low-effort/high-ROI addition that ensures the endpoint hasn't broken.

its all tests dude!!!

### DIFF
--- a/src/python/grapl-tests-common/grapl_tests_common/clients/graphql_endpoint_client.py
+++ b/src/python/grapl-tests-common/grapl_tests_common/clients/graphql_endpoint_client.py
@@ -1,4 +1,5 @@
 import os
+
 import requests
 
 
--- a/src/python/grapl-tests-common/grapl_tests_common/clients/graphql_endpoint_client.py
+++ b/src/python/grapl-tests-common/grapl_tests_common/clients/graphql_endpoint_client.py
@@ -1,4 +1,5 @@
 import os
+from typing import Any, Dict
 
 import requests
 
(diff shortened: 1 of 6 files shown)


src/python/grapl-tests-common/grapl_tests_common/clients/graphql_endpoint_client.py
```diff
@@ -0,0 +1,18 @@
+import os
+import requests
+
+
+class GraphqlEndpointClient:
+    def __init__(self, jwt: str) -> None:
+        hostname = "grapl-graphql-endpoint"
+        port = os.environ["GRAPL_GRAPHQL_PORT"]
+        self.endpoint = f"http://{hostname}:{port}"
```
christophermaier, line 10: Would it be better to just provide a `GRAPL_GRAPHQL_ENDPOINT` value, and not have to piece it together (also assuming a hard-coded hostname)?

christophermaier (approved):
Nice addition... I have a small suggestion, but otherwise 👍

colin-grapl (approved):

src/python/grapl-tests-common/grapl_tests_common/clients/graphql_endpoint_client.py
wimax-grapl: eh, i think the host + port is easy enough; + there are some libs that prefer specifying something like `port=` than adding `:${PORT}` to the endpoint url.

i will add a GRAPL_GRAPHQL_HOSTNAME though

PR #692, opened 2021-03-20
Fix 'make down' by running with the various project names that we use

inickles-grapl:

In trying to clean a local environment by `make down`, docker-compose would fail in trying to remove the docker network. This would be due to the Grapl local environment still being up after tests were done running, so containers were still running that were using the network. 

```
Removing network grapl-network
ERROR: error while removing network: network grapl-network id 6b829bd70b1b4fd4ae89bb89c37219b38039f06b4e472be72f19e0e202bbd852 has active endpoints
make: *** [Makefile:285: down] Error 1
```

These changes do a few things:
1. Remove from the test wrapper the code that would stop test containers and move that functionality into the Makefile. We still have some functionality of that wrapper in a script, but it only checks the exit codes of containers now.
2. Set .ONESHELL in the Makefile, and shell opts to preserve Make behavior. This is required for the previous and should 
3. The `make down` target now calls down for each of the compose project names we use.

I ran `make test` and ensured they all worked locally. Also, I verified a test failure still surfaces to Make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ export EVERY_COMPOSE_FILE=-f docker-compose.yml \
 DOCKER_BUILDX_BAKE := docker buildx bake $(DOCKER_BUILDX_BAKE_OPTS)
 VERBOSE_PANTS := PEX_VERBOSE=5 ./pants -ldebug
 
+COMPOSE_PROJECT_INTEGRATION_TESTS := grapl-integration_tests
+COMPOSE_PROJECT_E2E_TESTS := grapl-e2e_tests
+
 
 # Use a single shell for each of our targets, which allows us to use the 'trap'
 # built-in in our targets. We set the 'errexit' shell option to preserve
@@ -169,13 +172,13 @@ test-typecheck-build-support: ## Typecheck build-support Python code
 test-typecheck-pants: test-typecheck-pulumi test-typecheck-build-support ## Typecheck Python code with Pants
 
 .PHONY: test-integration
-test-integration: export COMPOSE_PROJECT_NAME := grapl_integration_tests
+test-integration: export COMPOSE_PROJECT_NAME := $(COMPOSE_PROJECT_INTEGRATION_TESTS)
 test-integration: export COMPOSE_FILE := ./test/docker-compose.integration-tests.yml
 test-integration: build-test-integration ## Build and run integration tests
 	$(MAKE) test-with-env
 
 .PHONY: test-e2e
-test-e2e: export COMPOSE_PROJECT_NAME := grapl_e2e_tests
+test-e2e: export COMPOSE_PROJECT_NAME := $(COMPOSE_PROJECT_E2E_TESTS)
 test-e2e: export export COMPOSE_FILE := ./test/docker-compose.e2e-tests.yml
 test-e2e: build-test-e2e ## Build and run e2e tests
 	$(MAKE) test-with-env
@@ -184,18 +187,19 @@ test-e2e: build-test-e2e ## Build and run e2e tests
 # intended for tests in docker-compose files that need the Grapl environment.
 .PHONY: test-with-env
 test-with-env: # (Do not include help text - not to be used directly)
-	function tearDown {
-		docker-compose stop
+	stopGrapl() {
+		# Unset COMPOSE_FILE to help ensure it will be ignored with use of --file
+		unset COMPOSE_FILE
+		docker-compose --file docker-compose.yml stop;
 	}
 	# Ensure we call stop even after test failure, and return exit code from 
 	# the test, not the stop command.
-	trap tearDown EXIT
+	trap stopGrapl EXIT
 	$(WITH_LOCAL_GRAPL_ENV)
 	# Bring up the Grapl environment and detach
 	$(MAKE) up-detach
 	# Run tests and check exit codes from each test container
-	COMPOSE_FILE=$(TEST_FILE) test/docker-compose-with-error.sh \
-		$(TARGET)
+	test/docker-compose-with-error.sh $(TARGET)
 
 
 .PHONY: test
@@ -284,8 +288,8 @@ up-detach: build-services ## Bring up local Grapl and detach to return control t
 down: ## docker-compose down - both stops and removes the containers
 	$(WITH_LOCAL_GRAPL_ENV)
 	docker-compose down --timeout=0
-	docker-compose --project-name "grapl-integration_tests" down --timeout=0
-	docker-compose --project-name "grapl-e2e_tests" down --timeout=0
+	docker-compose --project-name $(COMPOSE_PROJECT_INTEGRATION_TESTS) down --timeout=0
+	docker-compose --project-name $(COMPOSE_PROJECT_E2E_TESTS) down --timeout=0
 
 .PHONY: stop
 stop: ## docker-compose stop - stops (but preserves) the containers
@@ -295,7 +299,7 @@ stop: ## docker-compose stop - stops (but preserves) the containers
 .PHONY: e2e-logs
 e2e-logs: ## All docker-compose logs
 	$(WITH_LOCAL_GRAPL_ENV)
-	docker-compose $(EVERY_COMPOSE_FILE) -p grapl-e2e_tests logs -f
+	docker-compose $(EVERY_COMPOSE_FILE) --project-name $(COMPOSE_PROJECT_E2E_TESTS) logs -f
 
 .PHONY: help
 help: ## Print this help
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ DOCKER_BUILDX_BAKE_OPTS ?=
 ifneq ($(GRAPL_RUST_ENV_FILE),)
 DOCKER_BUILDX_BAKE_OPTS += --set *.secrets=id=rust_env,src="$(GRAPL_RUST_ENV_FILE)"
 endif
+COMPOSE_IGNORE_ORPHANS=1
 export
 
 export EVERY_COMPOSE_FILE=-f docker-compose.yml \
@@ -23,6 +24,15 @@ export EVERY_COMPOSE_FILE=-f docker-compose.yml \
 
 DOCKER_BUILDX_BAKE := docker buildx bake $(DOCKER_BUILDX_BAKE_OPTS)
 VERBOSE_PANTS := PEX_VERBOSE=5 ./pants -ldebug
+
+
+# Use a single shell for each of our targets, which allows us to use the 'trap'
+# built-in in our targets. We set the 'errexit' shell option to preserve
+# execution behavior, where failure from one line in a target will result in
+# Make error.
+# https://www.gnu.org/software/make/manual/html_node/One-Shell.html
+export SHELLOPTS:=$(if $(SHELLOPTS),$(SHELLOPTS):)errexit
+.ONESHELL:
 
 # Our `docker-compose.yml` file declares the setup of a "local Grapl"
 # environment, which can be used to locally exercise a Grapl system,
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -161,26 +167,33 @@ test-typecheck-build-support: ## Typecheck build-support Python code
...
-	# make test-integration TARGET=grapl-analyzerlib-integration-tests
-	$(WITH_LOCAL_GRAPL_ENV) \
-	$(MAKE) up-detach PROJECT_NAME="grapl-integration_tests" && \
-	test/docker-compose-with-error.sh \
-		-f ./test/docker-compose.integration-tests.yml \
-		-p "grapl-integration_tests" \
-		-t "$(TARGET)"
+	$(MAKE) test-with-env \
+		TEST_FILE="./test/docker-compose.integration-tests.yml" \
+		COMPOSE_PROJECT_NAME="grapl-integration_tests"
```
christophermaier, line 173: Reading this initially, it's a bit surprising that the `TEST_FILE` isn't handled by `COMPOSE_FILE`, and that it and `COMPOSE_PROJECT_NAME` aren't supplied as target-specific variables, as they are elsewhere.

That leads me to assume that `test-with-env` is doing something "special"... if that special-ness isn't truly required, it would be nice to eliminate it; if it is, it would be good to explicitly call out the reasons for it.

Makefile
```diff
@@ -247,32 +260,34 @@ push: ## Push Grapl containers to Docker Hub
...
 	# Primarily used for bringing up an environment for integration testing.
-	# Usage: `make up-detach PROJECT_NAME=asdf`
-	$(WITH_LOCAL_GRAPL_ENV) \
+	# For use with a project name consider setting COMPOSE_PROJECT_NAME env var
+	# Usage: `make up-detach`
+	$(WITH_LOCAL_GRAPL_ENV)
 	docker-compose \
-		-p $(PROJECT_NAME) \
-		-f docker-compose.yml \
+		--file docker-compose.yml \
```
christophermaier, line 273: If you explicitly supply `--file` here, any `COMPOSE_FILE` value provided is ignored. That implies that we could harmonize our usage of `TARGET_FILE`/`COMPOSE_FILE` elsewhere (though it would be good to document this intention here at this target; that it's specifically intended to spin up just the local Grapl environment, since that's not particularly evident from the help text comment).

Makefile
```diff
@@ -247,32 +260,34 @@ push: ## Push Grapl containers to Docker Hub
 
 .PHONY: up
 up: build-services ## Build Grapl services and launch docker-compose up
-	$(WITH_LOCAL_GRAPL_ENV) \
+	$(WITH_LOCAL_GRAPL_ENV)
```
christophermaier, line 263: I presume that the `ONESHELL` trick is what lets you dispense with the `\` escaping here, correct?

Makefile
```diff
@@ -161,26 +167,33 @@ test-typecheck-build-support: ## Typecheck build-support Python code
...
+test-with-env: # (Do not include help text - not to be used directly)
+	function tearDown {
+		docker-compose stop
+	}
+	# Ensure we call stop even after test failure, and return exit code from 
+	# the test, not the stop command.
+	trap tearDown EXIT
+	$(WITH_LOCAL_GRAPL_ENV)
+	$(MAKE) up-detach
+	# Run tests and heck exit codes from each test container
```
christophermaier, line 193: "heck" -> "check"

Makefile
```diff
@@ -247,32 +260,34 @@ push: ## Push Grapl containers to Docker Hub
...
 		up --detach --force-recreate
 
 .PHONY: down
 down: ## docker-compose down - both stops and removes the containers
-	$(WITH_LOCAL_GRAPL_ENV) \
-	docker-compose $(EVERY_COMPOSE_FILE) down --remove-orphans
+	$(WITH_LOCAL_GRAPL_ENV)
+	docker-compose down
+	docker-compose --project-name "grapl-integration_tests" down
+	docker-compose --project-name "grapl-e2e_tests" down
```
christophermaier, line 281: Might be worth adding `--timeout=0` to these calls, since we don't really care about state.

Extracting the particular project invocations to their own separate targets might make things more composable, and provide for better overall handling of environments. Though perhaps not, since it's currently impossible to have more than one of these environments running at the same time.

At the very least, extracting the _names_ of the environments as top-level makefile variables would be good, since they're repeated throughout a few times. After that, switching the "_" to "-" would be a nice touch.

test/docker-compose-with-error.sh
```diff
@@ -1,60 +1,38 @@
...
 SERVICES="$@"
 
 # Execute the 'up'
-docker-compose ${FILE_ARGS} --project-name "$p" up --force-recreate ${SERVICES} ${TARGET_ARGS}
+docker-compose ${FILE_ARGS} up --force-recreate ${SERVICES}
 
 # check for container exit codes other than 0
 EXIT_CODE=0
-ALL_TESTS=$(docker-compose ${FILE_ARGS} --project-name "$p" ps -q ${SERVICES})
+ALL_TESTS=$(docker-compose ps -q ${SERVICES})
```
christophermaier, line 35: Using `--quiet` instead of `-q` is more self-documenting and easier to read.

test/check-compose-exit-codes.sh
```diff
@@ -0,0 +1,19 @@
...
+ALL_TESTS=$(docker-compose ps -q ${TARGETS})
+for test in $ALL_TESTS; do
+    docker inspect -f "{{ .State.ExitCode }}" $test | grep -q ^0;
+    if [ $? -ne 0 ]; then 
+        EXIT_CODE=$?;
+        break
+    fi
+done
+
+exit $EXIT_CODE
```
christophermaier, line 19: I don't see where this file is actually used....

test/docker-compose-with-error.sh
```diff
@@ -1,60 +1,38 @@
...
+    echo 'This script calls into `docker-compose up` and checks the exit code' 1>&2
+    echo 'of each container upon exit. If any container exit code is non-zero,' 1>&2
+    echo 'this script will exit non-zero.' 1>&2
     echo
-    echo 'Options:' 1>&2
-    echo '    -f	Path to compose file. Can be passed multiple times.' 1>&2
-    echo '    -p	Project name.' 1>&2
-    echo '    -t	Target (if you only want to execute one thing).' 1>&2
+    echo 'Use Compose environment variables, such as COMPOSE_FILE and COMPOSE_PROJECT_NAME,' 1>&2
+    echo 'for driecting docker-compose.' 1>&2
```
christophermaier, line 13: "driecting" -> "directing"

test/docker-compose-with-error.sh
```diff
@@ -1,60 +1,38 @@
...
     exit 1
 }
 
 while getopts "hf:p:t:" arg; do
```
christophermaier, line 17: You don't need the config flags for `f`, `t`, or `p` anymore.

I'm not actually certain we need this `getopts` bit at all anymore.

test/docker-compose-with-error.sh
```diff
@@ -1,60 +1,38 @@
...
-    usage
-fi
-
-if [ -z "${TARGET_ARGS}" ]; then
-    TARGET_ARGS=""
-fi
-
+# We'll pass SERVICES to the end of docker-compose up to allow users to specify
+# specific services to run, instead of all of them, which is the default behavior.
 shift $(($OPTIND - 1))
```
christophermaier, line 27: I don't think this is necessary any longer, correct?

test/check-compose-exit-codes.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+shift $(($OPTIND - 1))
```
christophermaier, line 5: I don't think this is needed.

test/check-compose-exit-codes.sh
inickles-grapl: Gah, indeed, I didn't mean to commit this. It was from a route I started and decided against.

Makefile
```diff
@@ -24,6 +25,13 @@ export EVERY_COMPOSE_FILE=-f docker-compose.yml \
 DOCKER_BUILDX_BAKE := docker buildx bake $(DOCKER_BUILDX_BAKE_OPTS)
 VERBOSE_PANTS := PEX_VERBOSE=5 ./pants -ldebug
 
+# Use a single shell for each of our targets, and in doing so set
+# shell otps pipefail and errexit to preserve the behaviors of Make.
+# https://www.gnu.org/software/make/manual/html_node/One-Shell.html
+export SHELL:=/bin/bash
+export SHELLOPTS:=$(if $(SHELLOPTS),$(SHELLOPTS):)pipefail:errexit
+.ONESHELL:
```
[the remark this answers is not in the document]
inickles-grapl: Yeah, we don't even need bash, I'm not going to change the shell. We still need .ONESHELL though for using 'trap' built-in within a target.

Makefile
inickles-grapl: Can't set `COMPOSE_FILE` for the whole target here because we we're executing two docker-compose with different files.

test/docker-compose-with-error.sh
inickles-grapl: Agreed, I'll drop it.

Makefile
inickles-grapl: I've updated the help text to be more specific that this is intended only for bringing up the local environment only. I considered removing this target since it is now only used by one other target now, but I had seconds thoughts in case anyone has started to use this target directly.

Makefile
inickles-grapl: Would `--timeout=0` mean this could hang longer or indefinitely?

Note it is possible to have multiple environments _exist_ at the same time, even though not running. Running `down` will remove all the containers, even if they aren't running. We don't destroy the test containers for the tests in case a developer wants to inspect them after a run.

Makefile
inickles-grapl: I think I'll harmonize the `TARGET_FILE/COMPOSE_FILE` elsewhere and add documentation for that.

inickles-grapl: I agree it'd be nice to harmonize the `TARGET_FILE/COMPOSE_FILE` elsewhere, but it also seems a little tricky to me.

Is there documentation stating the COMPOSE_FILE env var will be ignored when `--file` is provided? I don't see it at:
https://docs.docker.com/compose/reference/overview/#specifying-a-path-to-a-single-compose-file

Makefile
inickles-grapl: For `COMPOSE_PROJECT_NAME` I'll change to set that as other targets do. For `TEST_FILE` I didn't want to set `COMPOSE_FILE` because we're executing docker-compose without that file and I wanted to be more explicit about that. I wouldn't want that value to interfere with that execution.

Makefile
christophermaier: I also went looking for documentation on the interplay between `COMPOSE_FILE` and `--file` but couldn't find it. However, I did some local testing and it looks like it works as I speculated above.

Makefile
christophermaier: Testing locally, `--timeout=0` results in a very quick shutdown of everything.

When I said "we don't really care about state", I should have said "we don't really care about cleanly shutting down", which is true, right?

Makefile
inickles-grapl: I guess we could explicitly unset `COMPOSE_FILE` in `up-detach` to avoid the possibility of that behavior changing. I'd be more comfortable harmonizing with `TEST_FILE` at that point. I'll do that.

Makefile
inickles-grapl: > Testing locally, --timeout=0 results in a very quick shutdown of everything.

Sold.

> At the very least, extracting the names of the environments as top-level makefile variables would be good, since they're repeated throughout a few times. After that, switching the "_" to "-" would be a nice touch.

I'm fine with moving to top-level variables, but since these values are used outside of the Makefile I'd rather not change them in this PR.

christophermaier (approved):
:shipit:

PR #693, opened 2021-03-20
Add vendoring to Cargo config

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/355

This adds vendoring to the Rust Cargo config and no longer adds the local Cargo registry to the Docker build cache.

This preserves the local developer experience of not having to redownload the dependency sources for each iteration during development. If the `vendor` directory exists locally, as it would in local Rust development, it'll be included in the initial Dockerfile COPY. Cargo will still download as needed if the directory doesn't exist locally.

I ran `make test-unit-rust` and `make test-integration` and verified we're not redownloading and rebuilding test dependencies. Also, I simulated the dev iteration by changing the sources and verified the dependency sources are not being redownloaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ __pycache__/
 .pytype/
 
 # Rust outputs
-target/
+src/rust/target/
 
 # Node packages
 node_modules/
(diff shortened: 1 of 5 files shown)


inickles-grapl:
Closing, this isn't behaving as expected on a clean repo.

PR #696, opened 2021-03-23
Enable lambdas to run in Localstack

christophermaier:
This adds enough Pulumi provisioning code and Localstack configuration to enable us to run both the `metric-forwarder` and `engagement-creator` as proper Lambda functions in Localstack. This includes Pulumi abstractions to replicate our CDK implementations (for the most part... we haven't yet gotten to the point where we're using VPCs in Localstack).

This doesn't yet rip out the corresponding implementations in CDK, but that's coming next. I wanted to ensure that this works in Localstack first before going much further.

This was tested by running `make test-e2e` many many times, as well as lots of `pulumi up` runs against my AWS sandbox account with CDK-deployed infrastructure.

Please read commit messages for further details.


jgrillo-grapl (approved):
LGTM, modulo that docker issue discussed in slack. If it can't be helped then so be it.

Makefile
```diff
@@ -14,13 +14,17 @@ endif
 COMPOSE_IGNORE_ORPHANS=1
 export
 
+export EVERY_LAMBDA_COMPOSE_FILE=-f docker-compose.lambda-zips.js.yml \
+	-f docker-compose.lambda-zips.python.yml \
+	-f docker-compose.lambda-zips.rust.yml
```
inickles-grapl, line 19: Should we use `--file` here instead of `-f`? I see `EVERY_COMPOSE_FILE` doesn't, but I don't think that means we shouldn't use `--file` moving forward.

Makefile
```diff
@@ -264,6 +264,7 @@ zip: build-lambdas ## Generate zips for deploying to AWS (src/js/grapl-cdk/zips/
 zip-pants: ## Generate Lambda zip artifacts using pants
 	./pants filter --filter-target-type=python_awslambda :: | xargs ./pants package
 	cp ./dist/src.python.provisioner.src/lambda.zip ./src/js/grapl-cdk/zips/provisioner-$(or $(TAG),latest).zip
+	cp ./dist/src.python.engagement-creator/engagement-creator.zip ./src/js/grapl-cdk/zips/engagement-creator-$(or $(TAG),latest).zip
```
inickles-grapl, line 267: I hadn't noticed the previous use of the default value on the line above, but doing so is redundant, as we set TAG=latest by default: https://github.com/grapl-security/grapl/blob/main/Makefile#L6

src/python/grapl-common/grapl_common/env_helpers.py
```diff
@@ -55,7 +55,9 @@ def _client_get(client_create_fn: Callable[..., Any], params: ClientGetParams) -
 
     # Unlike Rust FromEnv, we rely on boto3's built in region handling.
 
-    if all((endpoint_url, access_key_id, access_key_secret)):
+    if _running_in_localstack():
+        return _localstack_client(client_create_fn, params)
+    elif all((endpoint_url, access_key_id, access_key_secret)):
```
inickles-grapl, line 60: Is this necessary? It looks like some of this is already covered by: https://github.com/grapl-security/grapl/blob/main/local-grapl.env#L88. This adding back some "IS_LOCAL"-like code, and it seems like we don't need to, but maybe I'm missing something.

Makefile
christophermaier: Yeah, using `--file` would be the proper thing to do.

Makefile
christophermaier: Good call; I can clean that up.

src/python/grapl-common/grapl_common/env_helpers.py
christophermaier: It's adding "is localstack" code, but leaving everything else as it was. Since this is shared code, we don't want to break the other things that are using the `IS_LOCAL` flag. Once everything has been migrated over, however, we can come back and yank a bunch of this code (though the localstack-stuff would need to stay for now, absent adding some custom DNS and proxy magic to our docker setup.

inickles-grapl (approved):

PR #698, opened 2021-03-23
Fix Analyzer Executor tests 

wimax-grapl:
None, I just found it

- Fix AE bugs where we pass a Search Strategy into a str
- which required Lots Of Other Changes
- - introduced a cache delete, only used for test cleanliness

its all about tests

### DIFF
--- a/src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
+++ b/src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
@@ -9,11 +9,12 @@ import sys
 import traceback
 from collections import defaultdict
 from datetime import datetime
+from logging import Logger
 from multiprocessing import Pipe, Process
 from multiprocessing.connection import Connection
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List, Optional, Union
 
 import boto3  # type: ignore
 import redis
@@ -51,11 +52,14 @@ except Exception as e:
 
 # TODO:  move generic cache stuff into its own utility file
 class NopCache(object):
-    def set(self, key, value):
+    def set(self, key: str, value: str) -> None:
         pass
 
-    def get(self, key):
+    def get(self, key: str) -> bool:
         return False
+
+    def delete(self, key: str) -> None:
+        pass
 
 
 EitherCache = Union[NopCache, redis.Redis]
@@ -71,8 +75,14 @@ class AnalyzerExecutor:
     _singleton = None
 
     def __init__(
-        self, message_cache, hit_cache, chunk_size, is_local, logger, metric_reporter
-    ):
+        self,
+        message_cache: EitherCache,
+        hit_cache: EitherCache,
+        chunk_size: int,
+        is_local: bool,
+        logger: Logger,
+        metric_reporter: MetricReporter,
+    ) -> None:
         self.message_cache = message_cache
         self.hit_cache = hit_cache
         self.chunk_size = chunk_size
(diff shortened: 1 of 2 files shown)


src/python/analyzer_executor/tests/test_cache_integration.py
```diff
@@ -36,54 +46,76 @@ def _AnalyzerExecutorSingleton(stub_env=False, env_addr=None, env_port=None):
...
-    ae = AnalyzerExecutorSingleton(stub_env=False)
-
-    k1, k2 = st.text(min_size=3, max_size=64), st.text(min_size=3, max_size=64)
+    ae = executor_fixture(stub_env=False)
 
     assert not ae.check_hit_cache(k1, k2)
     ae.update_hit_cache(k1, k2)
     assert ae.check_hit_cache(k1, k2)
+    # Clean up, because Hypothesis provides duplicate inputs
+    ae.delete_hit_cache(k1, k2)
```
jgrillo-grapl, line 89: rather than introducing a function `delete_hit_cache()` into the analyzer-executor's public API that is only used in a test, would it make more sense to just generate hypothesis data that's not already in the cache? This can be accomplished pretty straightforwardly with a [`data()` strategy](https://hypothesis.readthedocs.io/en/latest/data.html#drawing-interactively-in-tests) or by using [`assume()` within a composite strategy](https://hypothesis.readthedocs.io/en/latest/data.html#composite-strategies) or by using the [`.filter()` combinator](https://hypothesis.readthedocs.io/en/latest/data.html#filtering)

jgrillo-grapl: A more robust method might be to isolate the cache entirely--i.e. arrange the code such that each test execution gets a completely new cache. This might be accomplished using monkeypatching and the `@pytest.mark.forked` decorator. See e.g.:
https://github.com/grapl-security/grapl/blob/internal-78-analyzer-deployer-api/src/python/analyzer-deployer/tests/test_app.py#L66-L75
and
https://github.com/grapl-security/grapl/blob/internal-78-analyzer-deployer-api/src/python/analyzer-deployer/tests/test_app.py#L318-L361
however, I'm not 100% certain this will play nice with hypothesis...

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -171,26 +181,41 @@ def check_caches(
...
         return bool(self.message_cache.get(event_hash))
 
     def update_msg_cache(self, file: str, node_key: str, msg_id: str) -> None:
-        to_hash = str(file) + str(node_key) + str(msg_id)
-        event_hash = hashlib.sha256(to_hash.encode()).hexdigest()
+        event_hash = self.to_event_hash((file, node_key, msg_id))
         self.message_cache.set(event_hash, "1")
 
+    def delete_msg_cache(self, file: str, node_key: str, msg_id: str) -> None:
+        """
```
ajm-grapl, line 198: if this is the docstring, let's just make it test utility code

ajm-grapl (approved):

src/python/analyzer_executor/tests/test_cache_integration.py
wimax-grapl: RE: the more-robust method, this is actually an integration test against the Redis service, so it has to be a real cache. No monkeypatching would work. 

RE: making a "only generate not-yet-used strings" strategy,
- that's definitely doable
- afaict it's considered non-hypothesis-ic
- quite frankly it's more effort than I want to spend on a random bug I found by accident :)

src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
wimax-grapl: Done

src/python/analyzer_executor/tests/test_cache_integration.py
jgrillo-grapl: The `analyzer-deployer` code I linked is also an integration test against real dynamodb tables (albeit a local one). The way it works is we have a function `app._analyzers_table()` which returns a `dynamdb.Table`. We monkeypatch that function in the test code to instead return a different table for each test. That way each test gets its own table and starts with fresh, uncorrupted state. I suspect we could devise a similar approach for our redis caches.

I believe you are correct about the hypothesis strategy stuff I proposed. I made a minimal example e.g.:

```python
@hypothesis.given(
    st.text(min_size=3, max_size=64).filter(lambda k: k not in DB),
    st.text(min_size=3, max_size=64),
)
def test_db(k, v):
    assert query_db(k) is None
    insert_into_db(k, v)
    assert query_db(k) == v
```

and got the following error:

```
E       hypothesis.errors.Flaky: Inconsistent data generation! Data generation behaved differently between different runs. Is your data generation depending on external state?
```

so I suspect we'd actually have to build a [stateful strategy](https://hypothesis.readthedocs.io/en/latest/stateful.html) here.

PR #701, opened 2021-03-24
Make down again

inickles-grapl:

None filed (yet?), but I noticed https://github.com/grapl-security/grapl/pull/692 doesn't quite achieve what it was meant to. The test containers, although stopped, will still exist after running make down. This is apparently due to not passing each compose file used for each project.

For running docker-compose down, run by passing `EVERY_COMPOSE_FILE` to help ensure all containers are covered.

Additionally, this explicitly sets a default value for `COMPOSE_PROJECT_NAME` to `grapl`, which should be the same as what we currently expect. docker-compose will use the directory name that holds the compose file, which is still "grapl" in most cases, but not for those under `test/`. This helps ensure all container names created by us will be prefixed with "grapl", which makes cleanup easier.

I ran `make test-e2e` locally then `make down` and checked the list of containers that still exist with `docker container ls --all`. I did similar after running `make up-detach` to ensure `make down` will still remove grapl containers that are still running.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -303,9 +303,9 @@ up-detach: build-services ## Bring up local Grapl and detach to return control t
 .PHONY: down
 down: ## docker-compose down - both stops and removes the containers
 	$(WITH_LOCAL_GRAPL_ENV)
-	docker-compose down --timeout=0
-	docker-compose --project-name $(COMPOSE_PROJECT_INTEGRATION_TESTS) down --timeout=0
-	docker-compose --project-name $(COMPOSE_PROJECT_E2E_TESTS) down --timeout=0
+	docker-compose $(EVERY_COMPOSE_FILE) down --timeout=0
+	docker-compose $(EVERY_COMPOSE_FILE) --project-name $(COMPOSE_PROJECT_INTEGRATION_TESTS) down --timeout=0
+	docker-compose $(EVERY_COMPOSE_FILE) --project-name $(COMPOSE_PROJECT_E2E_TESTS) down --timeout=0
 
 .PHONY: stop
 stop: ## docker-compose stop - stops (but preserves) the containers
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ ifneq ($(GRAPL_RUST_ENV_FILE),)
 DOCKER_BUILDX_BAKE_OPTS += --set *.secrets=id=rust_env,src="$(GRAPL_RUST_ENV_FILE)"
 endif
 COMPOSE_IGNORE_ORPHANS=1
+COMPOSE_PROJECT_NAME ?= grapl
 export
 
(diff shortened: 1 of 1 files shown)


wimax-grapl (approved):

Makefile
```diff
@@ -285,11 +285,9 @@ up-detach: build-services ## Bring up local Grapl and detach to return control t
...
 
 .PHONY: down
-down: ## docker-compose down - both stops and removes the containers
-	$(WITH_LOCAL_GRAPL_ENV)
-	docker-compose down --timeout=0
-	docker-compose --project-name $(COMPOSE_PROJECT_INTEGRATION_TESTS) down --timeout=0
-	docker-compose --project-name $(COMPOSE_PROJECT_E2E_TESTS) down --timeout=0
+down: ## Remove all Grapl containers
+	# Remove all Grapl containers - continue on error (no containers found)
+	docker rm --volumes --force $$(docker ps --filter "name=grapl*" --all --quiet) 2>/dev/null || true
```
christophermaier, line 290: Strictly speaking, you don't need the `*` in the filter; it treats the value as a prefix already.

christophermaier: A side effect of going away from `docker-compose down` is that this will leave the `grapl-network` docker network behind.

inickles-grapl: Ah yes, the network being left behind is a good point.

Makefile
```diff
@@ -12,6 +12,7 @@ ifneq ($(GRAPL_RUST_ENV_FILE),)
 DOCKER_BUILDX_BAKE_OPTS += --set *.secrets=id=rust_env,src="$(GRAPL_RUST_ENV_FILE)"
 endif
 COMPOSE_IGNORE_ORPHANS=1
+COMPOSE_PROJECT_NAME ?= grapl
```
christophermaier, line 15: Even though it's a no-op, this is important to leave in, mainly because of https://github.com/grapl-security/grapl/pull/696/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R102

(Basically, localstack needs to know the name of the container its running in for the lambda functionality to work properly for us. In `docker-compose`, that requires knowing the project name. However, the only way to get access to that information in the compose file is if it's explicitly set by environment variable. Using the default or using `--project-name` doesn't propagate that information.)

Makefile
christophermaier: A possibly-equivalent solution would be to filter on the network the container is attached to:

```
docker rm --volumes --force $(docker ps --filter "network=grapl-network" --all --quiet)
docker network rm grapl-network
```

inickles-grapl: I'm fine with using `docker-compose down` if you'd prefer. It'd mean things like networks and such would be handled automatically.

christophermaier: Yeah, I've got no problem with the `docker-compose down`-based approach. I'm cool with sticking to that.

PR #704, opened 2021-03-25
Make tests less dependent on local Grapl

inickles-grapl:

This makes our integration and E2E tests less dependent on the local Grapl environment, bringing us closer to being able to run them against an AWS deployment. There are two significant changes:

1. The tests no longer call `wait-for-it` for the local provisioner to finish executing. We move this to a new Make target `wait-for-local-provision`, which now won't return until provisioning has finished (or error after timeout). This target will be useful in general for waiting until provisioning is complete from command line.
2. The new `wait-for-local-provision` target builds a container for itself. It has a generic name 'test-utils', which conveys its purpose and makes room for future additions we may want to include in the future.
3. Hard-coded values that assumed a local environment have been parameterized via environment variables.
4. Add session token to Python AWS credential helpers to better support AWS official endpoints.

- I verified `make test-e2e test-integration` still works
- I ran these tests against an AWS deployment with VPN connectivity to back end services. Connectivity to AWS services works, though I was getting connectivity failures to DGraph, which probably speaks more to the state of my deployment than the validity of the test.


test/docker-compose.e2e-tests.yml
```diff
@@ -12,7 +12,6 @@ services:
       target: grapl-e2e-tests-build
     command: |
       bash -c "
-        wait-for-it grapl-provision:${WAIT_PORT} --timeout=250 &&
```
christophermaier, line 15: It would be worthwhile to see if we can remove `wait-for-it` from the base layer of any of these containers now.

(We might not, given how they share so much... if we can, though, we should.)

Makefile
```diff
@@ -298,6 +298,25 @@ up-detach: build-services ## Bring up local Grapl and detach to return control t
...
+	# the changes to local provisioning.
+	#
+	# It looks like docker-compose isn't honoring COMPOSE_IGNORE_ORPHANS
+	# for the 'run' command, so we're left with a bogus warning. This looks
+	# related: https://github.com/docker/compose/issues/8203.
+	docker-compose \
+		--file ./test/docker-compose.e2e-tests.yml \
+		run --rm \
+		grapl-e2e-tests \
+		wait-for-it grapl-engagement-view-uploader:$${WAIT_PORT} --timeout=250
```
christophermaier, line 319: 🤔 I don't know... I think having a container just for `wait-for-it` makes a lot of sense, in that it's very focused and the intent behind it and its usage is very clear. Nobody has to wonder why we're using the e2e-tests container for this, you don't have to explain it, and nothing mysteriously breaks if we remove `wait-for-it` from that container. It also doesn't create unnecessary dependencies between the main docker-compose file and the e2e one.

Such a container would be pretty simple, too. Here's one I found just now: https://github.com/whs/docker-wait-for-it/blob/master/Dockerfile (I'm not suggesting we directly use that one; just illustrating it's not an onerous task.)

Alternatively, we might be able to do something by listening for Docker events on the `grapl-engagement-view-uploader` container: https://docs.docker.com/engine/reference/commandline/events/

inickles-grapl: Fair enough.

Instead of pulling from some randos DockerHub, I'd prefer to pull from official debian and use apt to install, as we do elsewhere.

I'll push a change for this.

jgrillo-grapl: I agree, moving this to the make target makes it less transparent. In the long term we'll build healthchecks to accomplish the startup ordering task but for now having containers wait on their dependencies with `wait-for-it.sh` is straightforward.

test/docker-compose.integration-tests.yml
```diff
@@ -13,7 +13,6 @@ services:
       args:
         - CARGO_PROFILE=debug
     command: bash -c "
-      wait-for-it grapl-provision:${WAIT_PORT} --timeout=250 &&
```
jgrillo-grapl, line 16: why is it removed in the integration tests as well?

jgrillo-grapl (commented):
It looks like the integration tests portion of these changes is only half done.If I understand correctly, I'd expect to see some Makefile updates to account for the loss of `wait-for-it` in the compose file?

test/docker-compose.integration-tests.yml
inickles-grapl: It's redundant this point with the introduction of the `wait-for-local-provision` Make target that's used to bring up the local Grpal environment for both E2E and integration tests.

Makefile
inickles-grapl: This should be resolved now.

christophermaier (approved):

jgrillo-grapl (approved):
👍

PR #710, opened 2021-03-31
add provisioning back to docs, fix dockerfiles for aws deployments

andrea-grapl:

- Add AWS provision command to AWS documentation 

- Resolve newly introduced bugs with dockerfiles that were preventing successful deployments for `graphqlEndpoint` because we weren't including `node_modules` so we weren't able to use dependencies

Docs: N/A
Package Manager: Locally & AWS  off a different branch ( did not test this branch in AWS, but tested these changes in a different branch - wanted to add them here in case someone else runs into these issues before the PR fixing them goes up )

### DIFF
--- a/docs/setup/aws.md
+++ b/docs/setup/aws.md
@@ -144,6 +144,10 @@ see the available options by running:
 bin/graplctl aws deploy --help
 ```
 
+### Provision Grapl
+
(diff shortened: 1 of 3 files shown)


inickles-grapl (changes requested):
Looks like the inclusion of `src/js/grapl-cdk/--cloudformation-execution-policies` is a mistake?

PR #711, opened 2021-03-31
Graphql endpoint: change console.log level of most things to debug/error

wimax-grapl:
None

- Most `console.log`s in graphql_endpoint are much more acurately `.debug`s, so I made that change.
- I slightly refactored the RootQuery so it's dynamic (more on that soon)

`make build` - typescript should catch most of these problems. And of course, I now have graphql integration tests!

### DIFF
--- a/src/js/graphql_endpoint/modules/jwt.ts
+++ b/src/js/graphql_endpoint/modules/jwt.ts
@@ -18,7 +18,6 @@ const secretsmanager = new AWS.SecretsManager({
 });
 
 async function fetchJwtSecret(): Promise<string> {
-    console.log("JWT_SECRET_ID: ", JWT_SECRET_ID);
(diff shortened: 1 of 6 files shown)


src/js/graphql_endpoint/modules/jwt.ts
```diff
@@ -18,7 +18,6 @@ const secretsmanager = new AWS.SecretsManager({
 });
 
 async function fetchJwtSecret(): Promise<string> {
-    console.log("JWT_SECRET_ID: ", JWT_SECRET_ID);
```
wimax-grapl, line 21: seems semi suspicious, so, deleted

colin-grapl: The ID isn't a secret, but yeah it's fine to just remove this.

colin-grapl (approved):

PR #715, opened 2021-04-01
Fix engagement creator Lambda handler in CDK

christophermaier:
Back when we started generating the ZIP file for this service with Pants, I neglected to update CDK to take the new handler into account. This fixes that, and also removes the old container-based ZIP file build (which didn't use Pants).


jgrillo-grapl (approved):
👍

andrea-grapl (approved):

PR #718, opened 2021-04-01
Minor change to hypothesis tests - stop generating 3-char strings and assuming len > 3

wimax-grapl:
https://grapl-internal.slack.com/archives/C0176MS6LBY/p1617301403291500
`hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 4 valid examples in 1.01 seconds (7 invalid ones and 0 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.max_size or max_leaves parameters).`

- Stop assuming len > 3, when we're generating len-3 strings

it's all about tests

### DIFF
--- a/src/python/grapl_analyzerlib/grapl_analyzerlib/test_utils/strategies/misc.py
+++ b/src/python/grapl_analyzerlib/grapl_analyzerlib/test_utils/strategies/misc.py
@@ -9,9 +9,9 @@ T = TypeVar("T", bound=str)
 
 @st.composite
 def text_dgraph_compat(
-    draw: Callable[[st.SearchStrategy[str]], str], max_size: int = 128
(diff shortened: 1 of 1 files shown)


src/python/grapl_analyzerlib/grapl_analyzerlib/test_utils/strategies/misc.py
```diff
@@ -9,9 +9,9 @@
 
 @st.composite
 def text_dgraph_compat(
-    draw: Callable[[st.SearchStrategy[str]], str], max_size: int = 128
+    draw: Callable[[st.SearchStrategy[str]], str], max_size: int = 64
```
wimax-grapl, line 12: and went down to 64 because honestly it's not like the big length is gonna catch anything interesting

jgrillo-grapl (approved):
LGTM

PR #720, opened 2021-04-02
Add a DynamoDB table that defines node schemas

wimax-grapl:
we could totally reconsider this format before a commit. 

grapl-security/issue-tracker#357

- Add a new schema_properties table
- Both provisions should write something like the following to that table
- The purpose? To read those schemas at the graphql endpoint and dynamically generate GraphQL Types out of them

added a test to graphql endpoint suite


src/python/grapl-common/grapl_common/provision.py
```diff
@@ -0,0 +1,46 @@
...
+# - provisioner lambda
+# ( and grapl-analyzerlib seems like the wrong place to do that )
+Schema = Any
+
+
+class SchemaDict(TypedDict):
+    properties: List[SchemaPropertyDict]
+
+
+def store_schema_properties(table: Table, schema: Schema) -> None:
```
wimax-grapl, line 30: this function is like the crux of the PR

colin-grapl (approved):

PR #722, opened 2021-04-06
Fix (and test!) graphql endpoint lens_scope. Add a smoke test to e2e tests.

wimax-grapl:
None. Andrea reported a bug.

- I discovered a bug where I was setting stuff to `dgraph_type: undefined` and then we were checking *if that key existed* not its value
- added a smoke test to prevent this in the future

added tests

### DIFF
--- a/src/js/engagement_view/src/components/engagementView/sidebar/tables/toggleLensTable.tsx
+++ b/src/js/engagement_view/src/components/engagementView/sidebar/tables/toggleLensTable.tsx
@@ -62,7 +62,7 @@ export function ToggleLensTable({ setLens }: ToggleLensTableProps) {
 					}
 				}
 			);
-		}, 1000);
+		}, 5000);
 		return () => clearInterval(interval);
 	});
 
--- a/src/js/graphql_endpoint/server.ts
+++ b/src/js/graphql_endpoint/server.ts
@@ -7,6 +7,7 @@ import awsServerlessExpress = require("aws-serverless-express");
 import { validateJwt } from "./modules/jwt.js";
 //@ts-ignore
 import regexEscape = require("regex-escape");
+import { GraphQLError } from "graphql";
 
 const app = express();
 const PORT = process.env.PORT || 5000;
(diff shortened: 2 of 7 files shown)


src/js/engagement_view/src/components/engagementView/sidebar/tables/toggleLensTable.tsx
```diff
@@ -62,7 +62,7 @@ export function ToggleLensTable({ setLens }: ToggleLensTableProps) {
 					}
 				}
 			);
-		}, 1000);
+		}, 5000);
```
wimax-grapl, line 65: this is an annoyance I've wanted to change for a while - the constant pinging of the graphql endpoint is slowed down a bit.

src/js/graphql_endpoint/server.ts
```diff
@@ -77,6 +86,7 @@ app.use(
     return {
       schema: getRootQuerySchema(),
       graphiql: IS_LOCAL,
+      customFormatErrorFn: IS_LOCAL ? customFormatErrorFnForDebugging : undefined,
```
wimax-grapl, line 89: Gives you a bit more information in errors returned over graphql

src/python/graphql_endpoint_tests/tests.py
```diff
@@ -72,55 +72,3 @@ def _query_graphql_endpoint_for_lenses(
     """
     resp = gql_client.query(query)
     return resp["lenses"]
-
-
-def _query_graphql_endpoint_for_scope(
```
wimax-grapl, line 77: moved to the graphql_client

src/js/graphql_endpoint/server.ts
```diff
@@ -61,6 +62,14 @@ const corsDelegate = (req: cors.CorsRequest, callback: CorsCallback): void => {
   callback(null, corsOptions); // callback expects two parameters: error and options
 };
 
+function customFormatErrorFnForDebugging(error: GraphQLError) {
+  return {
```
andrea-grapl, line 66: Excellent!

andrea-grapl (approved):

PR #730, opened 2021-04-07
Add VPC infrastructure to Pulumi, Localstack

christophermaier:
This replicates (to the best of my knowledge) the infrastructure that
was generated for us by CDK's VPC construct:

- One VPC
- 2 public subnets, 2 private subnets, one each in two separate
  availability zones
- 1 internet gateway
- 1 NAT gateway
- Each lambda function gets a dedicated Security Group, as well as
  access to both private subnets.

### DIFF
--- a/pulumi/infra/network.py
+++ b/pulumi/infra/network.py
@@ -1,0 +1,138 @@
+import ipaddress
+import math
+from typing import Optional
+
+import pulumi_aws as aws
+
+import pulumi
+
+# TODO: Add policy requiring that each networking-related
+# infrastructure component have a "Name" (note capitalization) tag,
+# since this is how these objects get their names. Otherwise, they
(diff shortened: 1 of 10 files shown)


pulumi/infra/network.py
```diff
@@ -0,0 +1,139 @@
...
+# since this is how these objects get their names. Otherwise, they
+# have no name, which makes interaction via the AWS UI slightly more
+# awkward than it needs to be.
+#
+# Security groups don't seem to work this way, though. They get a name
+# independent of a Name tag (though the AWS UI shows both, somewhat
+# confusingly).
+
+
+class Network(pulumi.ComponentResource):
```
wimax-grapl, line 21: TBH I don't have much background on this sort of stuff, in particular cidr. may want second pair of eyes

wimax-grapl (approved):

PR #733, opened 2021-04-09
Rust sqs-executor cache changes

inickles-grapl:

This all started to resolve the following, but a number of other enhancements were made along the way:
- https://github.com/grapl-security/issue-tracker/issues/344

This PR not only swaps darkredis for redis-rs in the sqs-executor, it makes significant changes to the cache API:
- Remove the `get` public methods and `CacheResponse` and replace with a new method `filter_cached`. The `get` methods weren't actually "gets", and we aren't really storing values for keys. Users of the `get` methods, like generators, would partition the keys that weren't in the cache and continue to process those. Instead of requiring users to do this parsing, we now provide `filter_cached`.
- Add an `all_exists` method to the cache API, another helper that some users were previously using the `get` methods.
- The cache API now prefers slices over Vecs
- An integration test! For the changes in this PR, that is.

I included an integration test that intends to cover the changes made.

### DIFF
--- a/src/rust/generators/sysmon-subgraph-generator/src/generator.rs
+++ b/src/rust/generators/sysmon-subgraph-generator/src/generator.rs
@@ -1,11 +1,8 @@
-use std::borrow::Cow;
-
 use async_trait::async_trait;
 use grapl_graph_descriptions::graph_description::*;
 use grapl_observe::log_time;
 use log::*;
-use sqs_executor::{cache::{Cache,
-                           CacheResponse},
+use sqs_executor::{cache::Cache,
                    errors::{CheckedError,
                             Recoverable},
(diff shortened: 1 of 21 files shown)


src/rust/generators/sysmon-subgraph-generator/src/generator.rs
```diff
@@ -159,6 +150,7 @@ where
             events
                 .split(|i| &[*i][..] == &b"\n"[..])
                 .map(String::from_utf8_lossy)
+                .map(|s| s.to_string())
```
colin-grapl, line 153: `String::from_utf8_lossy` returns a String already, so calling `to_string` seems redundant.

colin-grapl (approved):

PR #735, opened 2021-04-10
Docker Compose 1.29.0 feature: wait for service to finish successfully, no more wait-for-it

inickles-grapl:

This leverages a new feature in Docker Compose 1.29.0 for one service to wait for another to finish successfully before starting, relieving our dependence on `wait-for-it`. This also leverages `heatlcheck` and `condition: service_healthy` so we don't have to depend on `wait-for-it` at all.

Additionally, this makes the Grapl services wait for the provisioning to finish before coming up, which gets rid of all the annoying logs at startup saying things like "SQS queue not found".

Running the local test suite is sufficient.


inickles-grapl:
Looks like GitHub has began rolling out version 20210412.1 of their runner, which includes Docker Compose 1.29.0: https://github.com/actions/virtual-environments (4% atm). I think we can wait for that to finish.

colin-grapl (approved):
yaaaaas

jgrillo-grapl (approved):
🔥

docker-compose.yml
```diff
@@ -179,9 +194,12 @@ services:
...
     depends_on:
-      - grapl-provision
-      - localstack
-      - redis
+      grapl-provision:
+        condition: service_completed_successfully
+      localstack:
+        condition: service_started
+      redis:
+        condition: service_started
```
inickles-grapl, line 202: I was thinking about removing the numerous copies of these we have. I put them in to better match the original, but it seems like noise at this point. Yes, these services require localstack and redis to have started, but:
1. they must have started by the time grapl-provision `service_completed_successfully` in the first place, and
2. `service_started` isn't _really_ the best criteria we want, it's that they're also healthy (though i don't want to add healthchecks for those at the moment).

christophermaier (approved):
👍

PR #740, opened 2021-04-13
GraphQL Endpoint also generates the query you need to send to `lens_scope`

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/357

- Adds a class that will dynamically generate the query we send to graphql_endpoint.
- We currently do not consume this anywhere except tests; I'm going to add a GraphQL query *to get this result* in next PR
Sample output from current schema:
https://gist.github.com/wimax-grapl/14f01759ba6160d700759107495c5711

This will eventually replace:
- the hardcoded query in src/js/engagement_view/src/services/graphQLRequests/utils_retrieveGraph/expandLensScopeQuery.tsx
- the hardcoded query in GraphqlEndpointClient

Some notes:
- PSA: We need to add a Stack Limit to expanding objects. For instance, a Process has Children with Parent has Children with Parent etc etc etc. I decided on a stack limit of 2.

i added some gr8 tests


src/js/graphql_endpoint/modules/schema_to_graphql.ts
```diff
@@ -36,7 +37,7 @@ function propertyToGraphql(
     }
 
     if (property.primitive == "Int") {
-        prim = GraphQLInt;
+        prim = GraphQLFloat;
```
wimax-grapl, line 40: GraphQLInt is only 32-bit and doesn't work with our timestamp ints

src/python/grapl_analyzerlib/grapl_analyzerlib/nodes/network_connection.py
```diff
@@ -19,7 +19,7 @@
 def default_network_connection_properties():
     return {
         "src_ip_address": PropType(PropPrimitive.Str, False),
-        "src_port": PropType(PropPrimitive.Str, False),
+        "src_port": PropType(PropPrimitive.Int, False),
         "dst_ip_address": PropType(PropPrimitive.Str, False),
         "dst_port": PropType(PropPrimitive.Int, False),
```
wimax-grapl, line 24: this seemed like a bug

src/js/graphql_endpoint/modules/schema_to_graphql.ts
colin-grapl: that's insane -_- how sad

src/python/grapl_analyzerlib/grapl_analyzerlib/nodes/network_connection.py
colin-grapl: probably yeah

colin-grapl (approved):

PR #742, opened 2021-04-14
Actually autoformat all TS files now. Apparently "**/*" is on by default in zsh, off by default in bash!

wimax-grapl:
none

we were not globbing files correctly earlier in the bin/format.sh; This fixes it.

its formatting


src/js/bin/format.sh
```diff
@@ -66,6 +67,8 @@ prettier \
     --config grapl-cdk/.prettierrc.toml \
     ${prettier_arg} \
     graphql_endpoint/**/*.ts \
-    grapl-cdk/**/*.ts \
+    grapl-cdk/bin/**/*.ts \
+    grapl-cdk/lib/**/*.ts \
+    grapl-cdk/test/**/*.ts \
     engagement_view/src/**/*.ts \
-    engagement_view/src/**/*.tsx
+    engagement_view/src/**/*.tsx \
```
wimax-grapl, line 74: # Everything below this is auto-formatting.

jgrillo-grapl (approved):

PR #743, opened 2021-04-14
Actually serving the query string from graphql-endpoint; + consume it in engagement-ux/tests

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/357

- Adds a new query to Graphql Endpoint: "give me the query I'm supposed to use for lens scope"
- Instead of that query being hardcoded in engagement_ux, it now grabs it from this endpoint.
- Some minor improvements to the `useEffect` interval things. 

I tested it with my local-grapl. There are some tests that hit the new endpoint, though engagement UX is still mostly untested.


src/js/engagement_view/Dockerfile
```diff
@@ -20,6 +20,7 @@ COPY js/engagement_view .
 FROM engagement-view-deps AS engagement-view-build
 
 # build sources
+RUN yarn add react-scripts
```
wimax-grapl, line 23: it stopped working unless i did this and i dont know why

colin-grapl (approved):
Super cool!!

PR #750, opened 2021-04-16
KEEP_TEST_ENV=1 to keep environment after running a test

wimax-grapl:
its hack week

KEEP_TEST_ENV=1 make test-e2e
will preserve the test environment
this is *super useful* for, say, manually logging in to engagement-UX afterwards and wanting to see how it renders the data (events.xml) 

locally, with and without KEEP_TEST_ENV


inickles-grapl (approved):
👍

PR #751, opened 2021-04-18
Deduplicate Rust dependencies

inickles-grapl:

This updates Rust dependencies to reduce the number of duplicates, reducing the number of build objects from 441 to 402.

This includes bringing the external sysmon library (https://crates.io/crates/sysmon), which was managed by @colin-grapl, to this repository to be its new home.

I ran the integration tests.


src/rust/endpoint-plugin/Cargo.toml
```diff
@@ -9,5 +9,5 @@ edition = "2018"
 [dependencies]
 grapl-graph-descriptions = { path="../graph-descriptions", version="*" }
 derive-dynamic-node = { path="../derive-dynamic-node" }
-uuid = "0.8.2"
-thiserror = "1.0.23"
+uuid = { version = "0", features=["v4"] }
```
colin-grapl, line 12: Version 0?

colin-grapl (approved):

src/rust/endpoint-plugin/Cargo.toml
inickles-grapl: I'd seen this in the graph merger, thought it was your doing: https://github.com/grapl-security/grapl/blob/main/src/rust/graph-merger/Cargo.toml#L49, figured I'd at least try to be more consistent. Shall we make them all "0.8"?

PR #752, opened 2021-04-19
Remove no-longer-existing file from `EVERY_COMPOSE_FILE` in Makefile.

inickles-grapl:

The Makefile was referencing a file that no longer exists. This removes that reference.

I'll rely on CI tests.


christophermaier (approved):
👍

PR #754, opened 2021-04-19
Make `TARGETS=` universal to any docker-compose-with-error.sh consumer; add documentation

wimax-grapl:
None

- Users can now specify `TARGETS="service1 service2" make test-something` in order to only run a subset of tests.
- Add documentation:
image

Ran on my machine with
no TARGETS specified
TARGETS=something real
TARGETS=something fake

### DIFF
--- a/test/docker-compose-with-error.sh
+++ b/test/docker-compose-with-error.sh
@@ -1,9 +1,13 @@
 #!/bin/bash
 
+# You can pass TARGETS to this script to specify
+# which services to run, instead of all of them, which is the default behavior.
+TARGETS="${TARGETS}"
+
 set -e
 
 usage() { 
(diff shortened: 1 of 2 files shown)


inickles-grapl (approved):
looks great 👌

test/docker-compose-with-error.sh
```diff
@@ -1,9 +1,13 @@
 #!/bin/bash
 
+# You can pass TARGETS to this script to specify
+# which services to run, instead of all of them, which is the default behavior.
+TARGETS="${TARGETS}"
```
inickles-grapl, line 5: `TARGETS="${TARGETS}"` seems redundant, I'd think we can remove but keep the comment. Your call tho.

PR #756, opened 2021-04-20
Remove a duplicate implementation of S3PayloadRetriever.

inickles-grapl:

I realized we had duplicate implementations of `S3PayloadRetriever` in the sqs-executor. This removes one.

I ran the integration tests.


inickles-grapl:
I ran a diff between the two implementations, they were the same.

colin-grapl (approved):

PR #758, opened 2021-04-20
Retry some Pants commands that usually fail in CI

wimax-grapl:
none

- add a new bash retry script to build-support
- use it in makefile with pants

I did it locally with a failing typecheck, and it worked as expected

### DIFF
--- a/.github/workflows/grapl-build.yml
+++ b/.github/workflows/grapl-build.yml
@@ -52,7 +52,7 @@ jobs:
             ~/.cache/pants/lmdb_store
 
       - name: Python unit tests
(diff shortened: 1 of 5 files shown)


christophermaier:
Any way we can push this into the CI layer itself, since that's really the only place it seems to happen?

Something like https://github.com/marketplace/actions/retry-step might be useful here.

(FWIW, this is something Buildkite supports trivially, so once we migrate, it'll be a simple matter of removing the Github Actions. It also keeps our Makefile simple)

christophermaier (approved):

PR #759, opened 2021-04-20
Put Engagement Edge service behind API Gateway

christophermaier:
This adds support for API Gateway to our Pulumi infrastructure code, and uses it to deploy the engagement-edge service as a lambda in our Local Grapl testing infrastructure.

While this moves to bring local and real infrastructures more in line, it also appears to result in slower overall test runs, which is (I think) due to Localstack implementation details. I have tried to tweak our Localstack lambda usage to minimize this as much as possible, but am still seeing it add a few minutes to the test run time locally.


jgrillo-grapl (approved):
👍

PR #760, opened 2021-04-20
Upgrade to Pulumi 3.0.0

christophermaier:
Including the Python libraries.


inickles-grapl (approved):

christophermaier:
Withdrawing this for now, as there appears to be a problem with the latest DynamoDB Table resource... looks like it no longer has a `hash_key` output, which breaks our local Grapl user creation.

PR #762, opened 2021-04-21
Allow generators to specify log compression

inickles-grapl:

This provides a means for generators (and services in general, but internally we'll probably stick with zstd) to specify the compression used for input logs.

There's a number of things cleaned up while we're at it:
- NDJSON decoder added to grapl-service lib
- OSQuery uses the generic NDJSON decoder instead of specifying its own.
- Clean up Sysmon generator parsing, which was all over the place. Create a PayloadDecoder for Sysmon generator. We can't use a generic XML parser due to how the `sysmon` is written.

Running the local test suite.


src/rust/generators/sysmon-subgraph-generator/src/generator.rs
```diff
@@ -54,121 +49,76 @@ where
...
+        for error in deserialization_errors {
+            error!("Deserialization error: {}", error);
+        }
+
+        let subgraph_results = deserialized_events
+            .into_iter()
+            .map(|event| GraphDescription::try_from(event));
+
+        // Report subgraph generation metrics
+        for result in subgraph_results.clone() {
```
colin-grapl, line 85: You can just call .iter() since yo're passing a reference in anyways - no need to clone.

src/rust/generators/sysmon-subgraph-generator/src/generator.rs
```diff
@@ -54,121 +49,76 @@ where
...
+            match error {
+                SysmonGeneratorError::UnsupportedEventType(_) => continue,
+                _ => error!("GraphDescription::try_from failed with: {:?}", error),
+            }
+        }
+
+        let final_subgraph =
+            subgraphs
+                .iter()
+                .fold(GraphDescription::new(), |mut current_graph, subgraph| {
```
colin-grapl, line 105: Is there no GraphDescription::merge? It's cool if not, nbd.

colin-grapl (approved):

src/rust/generators/sysmon-subgraph-generator/src/generator.rs
inickles-grapl: There is, it's called on the next line, `current_graph.merge(&subgraph);`

PR #765, opened 2021-04-21
Fix a bunch of typechecking in Model Plugin Deployer, and actually propagate-up errors as 500s

wimax-grapl:
…

none, but this reported bug
https://grapl-internal.slack.com/archives/C0176MS6LBY/p1619043750131600

- Fix a lot of typechecking
- fix a lot of `respond()` calls so they actually return non-200s

CI on this PR, and checking the logs

### DIFF
--- a/src/python/grapl-model-plugin-deployer/src/grapl_model_plugin_deployer.py
+++ b/src/python/grapl-model-plugin-deployer/src/grapl_model_plugin_deployer.py
@@ -12,12 +12,22 @@ from base64 import b64decode
 from hashlib import sha1
 from http import HTTPStatus
 from pathlib import Path
-from typing import TYPE_CHECKING, Dict, List, Optional, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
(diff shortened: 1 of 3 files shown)


src/python/grapl-model-plugin-deployer/src/grapl_model_plugin_deployer.py
```diff
@@ -201,7 +202,7 @@ def store_schema_properties(dynamodb: DynamoDBServiceResource, schema: Schema) -
 
 def provision_schemas(master_graph_client, raw_schemas):
```
colin-grapl, line 203: Should we add a None/Any return type here too?

colin-grapl (approved):

src/python/grapl-model-plugin-deployer/src/grapl_model_plugin_deployer.py
wimax-grapl: This is all copy pasted from the Provision service - gross - I decided to not waste too much time on this part

PR #769, opened 2021-04-22
Tweak log dumping script to account for Lambda functions, Buildkite migration

christophermaier:

With the addition of lambda functions to our Local Grapl setup, we need to adjust how we're capturing logs to make it easier to debug issues. In particular, we must capture lambda output from the Localstack instance's Cloudwatch endpoint while it is still running.

Here, we modify the `dump_artifacts.py` script to take this into account. It dynamically determines which lambda functions are running (based on the containers that Localstack spins up for them) and runs the appropriate `aws` CLI command to dump the logs.

Also added in this PR is a change to dump artifacts into a timestamped directory, rather than always using `/tmp/dumped_artifacts`. This is particularly important for our coming Buildkite rollout, because a given runner may service multiple test runs over its lifetime. If we were to use a static global path like `/tmp/dumped_artifacts`, we would end up mixing outputs from different runs. This does mean that we dump artifacts into the repository root, but we also add `.gitignore` rules and Makefile cleanup targets, so the impact should be minimal.

Additional cleanups and simplifications were added as appropriate.

Run `make test-integration` or `make test-e2e` and observe the creation of `artifacts_<COMPOSE_PROJECT>_<YYYYMMDDHHMMSS>` directories.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -355,7 +355,7 @@ clean-mount-cache: ## Prune all docker mount cache (used by sccache)
 
 .PHONY: clean-artifacts
 clean-artifacts: ## Remove all dumped artifacts from test runs (see dump_artifacts.py)
-	rm -Rf artifacts_*_[0-9]*
+	rm -Rf test_artifacts
 
 .PHONY: zip
 zip: build-lambdas ## Generate zips for deploying to AWS (src/js/grapl-cdk/zips/)
--- a/Makefile
+++ b/Makefile
@@ -250,6 +250,7 @@ test-with-env: # (Do not include help text - not to be used directly)
 		fi
 		# Unset COMPOSE_FILE to help ensure it will be ignored with use of --file
 		unset COMPOSE_FILE
+		etc/ci_scripts/dump_artifacts.py --compose-project=${COMPOSE_PROJECT_NAME}
 		docker-compose --file docker-compose.yml stop;
 	}
 	# Ensure we call stop even after test failure, and return exit code from
--- a/Makefile
+++ b/Makefile
@@ -352,6 +352,10 @@ clean: ## Prune all docker build cache and remove Grapl containers and images
 clean-mount-cache: ## Prune all docker mount cache (used by sccache)
 	docker builder prune --filter type=exec.cachemount
 
+.PHONY: clean-artifacts
+clean-artifacts: ## Remove all dumped artifacts from test runs (see dump_artifacts.py)
+	rm -Rf artifacts_*_[0-9]*
+
 .PHONY: zip
 zip: build-lambdas ## Generate zips for deploying to AWS (src/js/grapl-cdk/zips/)
 	docker-compose $(EVERY_LAMBDA_COMPOSE_FILE) up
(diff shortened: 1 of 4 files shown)


jgrillo-grapl (approved):
LGTM

Makefile
```diff
@@ -352,6 +353,10 @@ clean: ## Prune all docker build cache and remove Grapl containers and images
 clean-mount-cache: ## Prune all docker mount cache (used by sccache)
 	docker builder prune --filter type=exec.cachemount
 
+.PHONY: clean-artifacts
+clean-artifacts: ## Remove all dumped artifacts from test runs (see dump_artifacts.py)
+	rm -Rf artifacts_*_[0-9]*
+
```
wimax-grapl, line 359: I don't super love that this just keeps adding new folders to <grapl repo root>; I personally would prefer that these be in, like, `/tmp/grapl_artifacts/artifacts_yymmdd_etc`

Makefile
```diff
@@ -250,6 +250,7 @@ test-with-env: # (Do not include help text - not to be used directly)
 		fi
 		# Unset COMPOSE_FILE to help ensure it will be ignored with use of --file
 		unset COMPOSE_FILE
+		etc/ci_scripts/dump_artifacts.py --compose-project=${COMPOSE_PROJECT_NAME}
```
wimax-grapl, line 253: nice

etc/ci_scripts/dump_artifacts.py
```diff
@@ -14,77 +17,135 @@
...
-                container_name,
+                "tail",
+                f"/aws/lambda/{lambda_name}"
             ],
             stdout=out_stream,
+            env={
+                "PATH": os.environ["PATH"],
+                # "test" is the value assumed by Localstack
+                "AWS_ACCESS_KEY_ID": "test",
+                "AWS_SECRET_ACCESS_KEY": "test"
```
wimax-grapl, line 101: would be nice to `source` these from local-grapl.env, but not the biggest issue in the world

etc/ci_scripts/dump_artifacts.py
```diff
@@ -14,77 +17,135 @@
...
     # docker-compose prefixes volume names with the compose project name.
     prefix = f"{compose_project}_" if compose_project else ""
-    cmd = f"docker run -d --volume {prefix}{volume_name}:/{volume_name} alpine true"
     container_id = (
-        subprocess.run(cmd.split(" "), capture_output=True)
-        .stdout.decode("utf-8")
-        .strip()
+        subprocess.run(
+            f"docker run -d --volume {prefix}{volume_name}:/{volume_name} alpine true",
+            shell=True,
```
wimax-grapl, line 135: i finally learned about what `shell=True` and how that interacts with spaces-in-a-cmd last night :+1:

wimax-grapl (approved):

Makefile
wimax-grapl: or maybe jsut <grapl_repo_root>/artifacts/yymmdd_etc

PR #774, opened 2021-04-26
Put UX Router behind the API Gateway

christophermaier:
Additionally, build the ZIP with Pants, remove some Docker
infrastructure for building ux-router images, and also simplify the
ux-router code a bit (mainly by moving the content of `env_vars` into
the main module).

### DIFF
--- a/pulumi/infra/ux_router.py
+++ b/pulumi/infra/ux_router.py
@@ -1,0 +1,50 @@
+from typing import Optional
+
+import pulumi_aws as aws
+from infra.bucket import Bucket
+from infra.config import GLOBAL_LAMBDA_ZIP_TAG
+from infra.lambda_ import Lambda, LambdaExecutionRole, PythonLambdaArgs, code_path_for
+from infra.metric_forwarder import MetricForwarder
+from infra.network import Network
+from infra.secret import JWTSecret
+
+import pulumi
+
+
+class UxRouter(pulumi.ComponentResource):
+    def __init__(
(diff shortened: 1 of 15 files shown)


pulumi/infra/ux_router.py
```diff
@@ -0,0 +1,50 @@
...
+            args=PythonLambdaArgs(
+                execution_role=self.role,
+                description=GLOBAL_LAMBDA_ZIP_TAG,
+                handler="lambdex_handler.handler",
+                code_path=code_path_for(name),
+                env={"GRAPL_LOG_LEVEL": "INFO", "UX_BUCKET_NAME": ux_bucket.bucket},
+                timeout=5,
+                memory_size=128,
+            ),
+            # TODO: I don't think we need a network, because I don't
```
colin-grapl, line 40: ux-router only acts as a (bad) proxy to S3. Not sure if this mean sit does or doesn't need `network`, but yeah.

colin-grapl (approved):

jgrillo-grapl (approved):
LGTM

pulumi/infra/ux_router.py
christophermaier: Yeah, I'm keeping consistent with what's in the CDK code as much as possible. There are a handful of places like this and a few others where it looks like we're doing more than is strictly necessary. Once we're transitioned, I'm going to make another pass for those kinds of things.

PR #776, opened 2021-04-27
Benchmarking

d0nut-grapl:

Benchmark issue: 416

Adds basic benchmarks to our generators (aside from generic generators).

Tested locally and verified benchmarks work.


src/rust/generators/sysmon-subgraph-generator/benches/generator_bench.rs
```diff
@@ -0,0 +1,62 @@
...
+use grapl_test::cache::EmptyCache;
+use sqs_executor::{event_decoder::PayloadDecoder,
+                   event_handler::CompletedEvents};
+use sysmon_subgraph_generator_lib::{generator::SysmonSubgraphGenerator,
+                                    metrics::SysmonSubgraphGeneratorMetrics,
+                                    serialization::SysmonDecoder};
+use test::Bencher;
+use tokio::runtime::Runtime;
+
+const SYSMON_SAMPLE_DATA_FILE: &'static str = "../../../../etc/sample_data/events6.xml";
```
inickles-grapl, line 17: I know we're mono repo, but in the future I'd like to see this kind of thing be more self contained. In this case, that'd be moving to events6.xml to the sysmon generator directory. It's not like that file is useful without that generator.

Not to say I wanna require that for this PR, but  if you agree I'll bring it up for discussion.

src/rust/rust-toolchain
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.51.0"
+channel = "nightly"
```
inickles-grapl, line 2: We don't wanna do this.

src/rust/generators/sysmon-subgraph-generator/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+#![type_length_limit = "1334469"]
```
inickles-grapl, line 1: Are these needed anymore? As I recall they were some workaround to a weird issue in Rust a while back.

inickles-grapl (changes requested):
Considering we should use `sqs-executor::cache::NopCache` instead of the new `EmptyCache`, seems we don't need the new grapl-test crate.

Also, we shouldn't change the rust toolchain from 1.51 to nightly.

src/rust/generators/sysmon-subgraph-generator/benches/generator_bench.rs
colin-grapl: Yeah, I think we should use a dedicated benchmark file - events6.xml could move, or change in the future.

colin-grapl (changes requested):

src/rust/grapl-test/src/cache.rs
```diff
@@ -0,0 +1,33 @@
...
+use sqs_executor::cache::{Cache,
+                          Cacheable};
+
+use crate::error::GenericError;
+
+#[derive(Clone)]
+pub struct EmptyCache {}
+
+#[async_trait]
+impl Cache for EmptyCache {
```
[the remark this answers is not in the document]
d0nut-grapl: Oh I didn't know about that! Thanks!

src/rust/rust-toolchain
d0nut-grapl: Ahh yea I forgot to revert this. I had to fix a bug that was caused by this file actually

src/rust/generators/sysmon-subgraph-generator/benches/generator_bench.rs
d0nut-grapl: Will move a copy of the test data into the repos specifically! Sorry; got lazy during hack week :)

The code limits the number of parsed events that each generator is processing so the results should be relatively consistent and comparable.

inickles-grapl (approved):

colin-grapl (approved):

PR #782, opened 2021-04-29
Deduplicate provisioning code

wimax-grapl:
Which issue does this PR correspond to?
none

What changes does this PR make to Grapl? Why?
Lots of our provisioning code is duplicated (actually, triplicated but we are removing the old grapl_provision soon).
This puts it all together in provision_common - which should probably eventually be a lib or something - but whatever.

How were these changes tested?
Only locally because I am a monster. I'm gonna fire off a new aws provision though just in case.


src/python/grapl-common/grapl_common/provision.py
```diff
@@ -1,64 +0,0 @@
-from __future__ import annotations
```
wimax-grapl, line 1: This file was moved from grapl-common to grapl_analyzerlib.
Pros: actually typed Schema, GraphClient, etc
Cons: can't typeddict

jgrillo-grapl (approved):
LGTM

PR #783, opened 2021-04-30
[draft do not review] experiment with j8

wimax-grapl:


wimax-grapl:
doesnt seem that much quicker

PR #784, opened 2021-04-30
Internal 397 eng view unit tests

andrea-grapl:
https://github.com/grapl-security/issue-tracker/issues/397

Initial JS Unit tests for Engagement View. Still have more tests to write, will take out another PR. 

`yarn run test` 
`yarn run test:coverage`


src/python/grapl_e2e_tests/tests.py
```diff
@@ -232,7 +232,7 @@ def check_login() -> None:
 def check_invalid_creds() -> None:
     resp = EngagementEdgeClient().invalid_creds()
     if resp.status_code == 403:
-        assert "Provided invalid creds & was unauthorized"
+        assert resp.status_code == 403, "Provided invalid creds which were unauthorized"
```
wimax-grapl, line 235: `if X: assert X` smells fishy.

Sounds like you just want to tell the end-user "hey, there is an error here", which is not what `assert` is for. 
Assert is for, well, asserting "the following must be true." 

The following seems more correct:
```raise Exception("Provided invalid creds, which were unauthorized")```

Comment (andrea-grapl) [reply src/python/grapl_e2e_tests/tests.py]: I think I wanted assert here because my `invalid_credentials` method is expecting a 403 to be returned, I don't  think I want to raise an exception here because I wanted to test if we provide invalid credentials, we should get an response noting that. I don't think we're testing for an exception here, what do you think. Basically, if we get a 403 with invalid creds, I want that test to pass because that's the expected behavior for this test. I think I'd want an exception if we provided valid creds with and got a 403

File: src/python/grapl_e2e_tests/tests.py [hunk 2]
```diff
@@ -215,25 +219,18 @@ def get_notebook_url() -> None:
...
-    else:
-        raise TestException(
-            "Unexpected authorization with invalid credentials - major issues!"
-        )
+    assert resp.status_code == 403, "We expected a 403 forbidden"
+
+
+def check_empty_creds() -> None:
+    resp = EngagementEdgeClient().empty_creds()
+    assert resp.status_code == 500, "Expected 500 permissions error"
```
colin-grapl, line 236: Open a bug for this - it should return 403, not 500.

colin-grapl (approved):

PR #785, opened 2021-04-30
Add make lint-js to the buildkite pipeline

wimax-grapl:
this got lost in the move to buildkite


christophermaier (approved):
Emoji compliant: 👍

wimax-grapl:
actually let's hold off landing this, gonna try dockerizing our formatting stuff

christophermaier:
🤔 I'm a bit torn on this.

On the one hand, containerizing our requirements is a good thing, and allows us to more easily do the same thing in CI as we do locally. It also makes it easier for new contributors to get up and running because they don't have to do a bunch of fiddly setup tasks.

On the other hand, adding more docker-compose files, with a specialized Makefile to run inside those containers, feels like a lot of additional overhead and indirection. Perhaps this will be moderated in the future once we have a set of blessed base container images to work from.

I'm afraid I don't have a whole bunch in the way of constructive alternatives, and don't want to block this work, but this triggered my spidey sense.

wimax-grapl:
Agree re: spidey sense; removed the makefile complexity

wimax-grapl:
I'm gonna land this for now since it ensures continued future code quality; we can revisit the specifics at a later date.

PR #788, opened 2021-05-04
statically identifiable nodes use hashing instead of dynamodb

d0nut-grapl:

Part of Node Identifier Improvements

Changes statically identifiable nodes to use hashing instead of tracking node keys in dynamodb. As nodes with static identification strategies *must* be unique for a given set of static properties, we can hash these properties in a predictable order to avoid reaching out to dynamodb.

Integration tests + Unit Tests


wimax-grapl (approved):
code seems good. filed a bug for the e2e failure https://github.com/grapl-security/issue-tracker/issues/420

PR #789, opened 2021-05-04
Use Docker mount cache for Rust builds

inickles-grapl:

This changes the Rust Dockerfile to better leverage cache mounts by using them for the following:

1. The Cargo `target` directory, instead of using sccache and its local directory
2. The Cargo local registry. This prevents from having to redownload Rust sources on subsequent builds.
3. The Apt local repository. This prevents from having to redownload OS packages on subsequent builds.

To make this work, we need to save the `target` and Cargo local registry directories in the final image used for Rust tests in containers, so we have steps to copy them to temp directory and move them back before finishing.

This speeds up Rust builds for both clean builds and subsequent builds with changes (measured on my local machine):
| | Main  | This PR |
| ------------- | ------------- | ------------- |
| Clean build | 6:52 | 5:46 |
| Subseq build | 3:47 | 1:25 |

`make test`

### DIFF
--- a/.buildkite/scripts/integration.sh
+++ b/.buildkite/scripts/integration.sh
@@ -2,10 +2,6 @@
 
 set -euo pipefail
 
-echo "unset RUSTC_WRAPPER" > rust_env.sh
-chmod 777 rust_env.sh
-
 export GRAPL_LOG_LEVEL=DEBUG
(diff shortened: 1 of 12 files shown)


colin-grapl (approved):

jgrillo-grapl (approved):
LGTM! 👍

PR #792, opened 2021-05-04
Fix the dockerized JS formatter (had permissions issues from 'user nobody')

wimax-grapl:
none

Remove 1 line from a Dockerfile that prevented us from writing to a bind mount

locally.
if you're curious how this all got committed, somebody landed JS changes that were not rebased on my formatter change. woops.


etc/formatter/Dockerfile
```diff
@@ -13,6 +13,4 @@ RUN npm install -g prettier@2.2.1
 
 # TODO:
 # In later iterations, we can also install all the Rustup cargofmt stuff here
-# as well as Pants to perform containerized linting.
-
-USER nobody
```
wimax-grapl, line 18: this is the main change here. rest is just Prettier working its magic.

andrea-grapl (approved):

etc/formatter/Dockerfile
inickles-grapl: Looks like this will cause it to run as root. Would that make the files be owned as root after running the formatter? Maybe not since all files changes already exist, so maybe ownership is retained.

However, I'd like if we ran the container as the current uid:gid from the host. The Makefile already sets UID and GID as env vars, so in the docker-compose you should be able to set `user: "${UID}:${GID}"`. How does that sound?

wimax-grapl: Done, good idea

PR #793, opened 2021-05-04
Bump up Localstack startup period; write docker ps results

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/420

- Increase localstack startup period before it starts failing healthcheck
- while debugging the above issue I had to make an assumption about the identity of a contianer hash; to avoid that in the future we're writing `docker ps` to the artifacts.

throw it on CI and see what happens

### DIFF
--- a/etc/ci_scripts/dump_artifacts.py
+++ b/etc/ci_scripts/dump_artifacts.py
@@ -117,6 +117,19 @@ def _dump_docker_log(container_name: str, dir: Path) -> None:
             shell=True,
         )
 
+def dump_docker_ps(dir: Path) -> None:
+    """
+    run `docker ps` and dump to $DIR/docker_ps.log
+    """
+    destination = dir / "docker_ps.log"
+    logging.info(f"Dumping 'docker ps' to '{destination}'")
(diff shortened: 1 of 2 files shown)


etc/ci_scripts/dump_artifacts.py
```diff
@@ -117,6 +117,19 @@ def _dump_docker_log(container_name: str, dir: Path) -> None:
...
+    """
+    run `docker ps` and dump to $DIR/docker_ps.log
+    """
+    destination = dir / "docker_ps.log"
+    logging.info(f"Dumping 'docker ps' to '{destination}'")
+    with open(destination, "wb") as out_stream:
+        subprocess.run(
+            f"docker ps",
+            stdout=out_stream,
+            stderr=subprocess.DEVNULL,
```
christophermaier, line 130: Might be useful to capture stderr together with stdout, no?

wimax-grapl: done

inickles-grapl (approved):
👍

PR #799, opened 2021-05-07
Node identifier cleanup and preparing for batch identification

d0nut-grapl:

Node Identifier cleanup and batch identification work 

This change performs general cleanup, documentation, and reorganization of the node identifier.

The reorganization is to help support the move to batch node identification.

Tested locally


src/rust/node-identifier/src/lib.rs
```diff
@@ -61,23 +44,29 @@ use sqs_executor::{
...
+    * [Session](`grapl_graph_descriptions::graph_description::Session`) - strategy used for nodes with lifetimes (e.g. process, network connection)
+    * [Static](`grapl_graph_descriptions::graph_description::Static`) - strategy used for nodes with canonical and unique identifiers (e.g. aws events)
+*/
 #[derive(Clone)]
 pub struct NodeIdentifier<D, CacheT>
 where
-    D: DynamoDb + Clone + Send + Sync + 'static,
-    CacheT: Cache + Clone + Send + Sync + 'static,
+    D: DynamoDb + Clone + Send + Sync,
+    CacheT: Cache + Clone + Send + Sync,
```
wimax-grapl, line 69: A) I believe there's some way to deduplicate all these constraints throughout this file
B) If so - should we?

wimax-grapl (approved):

src/rust/node-identifier/src/lib.rs
d0nut-grapl: Good call. That's probably a good idea; i forgot I could use `type <some_name> = <trait bounds>` to simply this.

d0nut-grapl: err, not type aliases. I think I can do this with a new Trait that requires the appropriate trait bounds. Hmm Let me experiment a bit more.

d0nut-grapl:
```
pub trait EventHandlerDynamoDb: DynamoDb + Clone + Send + Sync {}
pub trait EventHandlerCache: Cache + Clone + Send + Sync {}

impl<D> EventHandlerDynamoDb for D where D: DynamoDb + Clone + Send + Sync {}
impl<C> EventHandlerCache for C where C: Cache + Clone + Send + Sync {}
```

Seems to work. I'll add this down into the sqs-executor so we get this everywhere.

PR #801, opened 2021-05-07
Specify Pulumi lambda log retention (still needs dgraph, fargate)

wimax-grapl:
grapl-security/issue-tracker#427

Make our service log retention 1 month.

CI

### DIFF
--- a/pulumi/infra/lambda_.py
+++ b/pulumi/infra/lambda_.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass, field
 from typing import Mapping, Optional, Union
 
 import pulumi_aws as aws
+from infra import service_shared
 from infra.config import (
     DEPLOYMENT_NAME,
     GLOBAL_LAMBDA_ZIP_TAG,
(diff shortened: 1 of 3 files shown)


pulumi/infra/service_shared.py
```diff
@@ -0,0 +1,8 @@
+"""
+Things like consts around cost tuning that should be shared by both
+Lambda and Fargate services.
+"""
+
+
+def get_service_log_retention_days() -> int:
+    return 31
```
christophermaier, line 8: Might be better to put this into https://github.com/grapl-security/grapl/blob/main/pulumi/infra/config.py

pulumi/infra/lambda_.py
```diff
@@ -178,17 +179,17 @@ def __init__(
...
-                f"{name}-log-group",
-                name=f"/aws/lambda/{lambda_name}",
-                opts=pulumi.ResourceOptions(parent=self),
-            )
+        # Lambda function output is automatically sent to log
+        # groups named like this; we create one explicitly so we can
+        # specify retention rules.
+        self.log_group = aws.cloudwatch.LogGroup(
+            f"{name}-log-group",
+            name=f"/aws/lambda/{lambda_name}",
```
colin-grapl, line 187: Would it make sense to do `/aws/lambda/deployname`? AWS APIs often take prefixes, so it would make some querying easier. NBD

colin-grapl (approved):

pulumi/infra/lambda_.py
christophermaier: I believe that AWS automatically creates these log groups with this naming pattern. We're only creating it because we need a handle to pass it to the forwarder.

christophermaier (approved):
👍

PR #804, opened 2021-05-09
Dgraph Swarm Cluster in Pulumi

wimax-grapl:

Adds a new pulumi resource for a Dgraph swarm cluster.
Doesn't happen locally, only in AWS, since we're killing localgrapl.

The deploy to AWS works fine, but I haven't tested it against the provisioner lambda (that's not in pulumi yet, but it's up next)


docker-compose.yml
```diff
@@ -470,7 +470,7 @@ services:
       /bin/bash -c "
          pulumi login --local &&
          pulumi stack init local-grapl --non-interactive &&
-         pulumi up --yes --skip-preview --stack=local-grapl &&
+         pulumi up --yes --skip-preview --stack=local-grapl --verbose 5 &&
```
wimax-grapl, line 473: just for the sake of attempting to debug it. does nothing

pulumi/infra/swarm.py
```diff
@@ -0,0 +1,328 @@
...
+                        ],
+                        "Resource": "*",
+                    },
+                ],
+            }
+        ),
+        opts=opts,
+    )
+    return aws.iam.RolePolicyAttachment(
+        f"attach_{policy_name}",
```
wimax-grapl, line 102: (this is me manually attempting to get around an un-needed `apply`; god i'd so much prefer some sort of await type syntax here)

pulumi/infra/swarm.py
```diff
@@ -0,0 +1,209 @@
...
+        # InstanceProfile for swarm instances
+        aws.iam.InstanceProfile(
+            "SwarmInstanceProfile",
+            opts=child_opts,
+            role=self.role.name,
+            name=f"{DEPLOYMENT_NAME}-swarm-instance-profile",
+        )
+
+    def cluster_host_port(self) -> pulumi.Output[str]:
+        return Output.concat(self.swarm_hosted_zone.name, ":9080")
```
christophermaier, line 170: Would it be worth extracting this port as a variable? It's repeated 3 times in various places in this code.

pulumi/infra/swarm.py
```diff
@@ -0,0 +1,209 @@
...
+                "_to_",
+                self.security_group._name,
+                "_for_",
+                str(port_range),
+            ]
+        )
+
+        # We'll accept connections from Other into SecurityGroup
+        aws.ec2.SecurityGroupRule(
+            f"ingress_{descriptor}",
```
christophermaier, line 188: Naming and casing.

pulumi/infra/swarm.py
```diff
@@ -0,0 +1,209 @@
...
+            opts=opts,
```
christophermaier, line 189: I think it might make sense to have the SecurityGroup be the parent resource:

```python
...
opts=pulumi.ResourceOptions(parent=self.security_group)
...
```
Incidentally, that might allow you to simplify the name of the `SecurityGroupRule`; you wouldn't necessarily need the name of `self.security_group` because that could then be understood from the parent/child relationship.

(Same applies for the egress rules below.)

pulumi/infra/swarm.py
```diff
@@ -0,0 +1,209 @@
...
+        self, other: aws.ec2.SecurityGroup, port_range: Ec2Port, opts: ResourceOptions
+    ) -> None:
+        descriptor = "".join(
+            [
+                "from_",
+                other._name,
+                "_to_",
+                self.security_group._name,
+                "_for_",
+                str(port_range),
```
christophermaier, line 182: Might want a custom string method for this; as it is, this generates a name with content like `Ec2Port(protocol='tcp', port=9080)`; something like `TCP:9080` would be just as informative, but less verbose.

pulumi/infra/swarm.py
```diff
@@ -0,0 +1,209 @@
...
+            to_port=port_range.port,
+            protocol=port_range.protocol,
+        )
+
+        # Allow connections out of Other to Self
+        aws.ec2.SecurityGroupRule(
+            f"egress_{descriptor}",
+            opts=opts,
+            type="egress",
+            # Perhaps this is wrong
```
christophermaier, line 203: Why?

pulumi/__main__.py
```diff
@@ -3,22 +3,35 @@
...
 from infra.service_queue import ServiceQueue
 
-if __name__ == "__main__":
 
+def _get_dgraph_cluster(network: Network) -> DgraphCluster:
+    if LOCAL_GRAPL:
+        return LocalStandInDgraphCluster()
+    else:
+        return DgraphCluster(
+            name="swarm",
```
christophermaier, line 20: "dgraph" might be a better name for this than "swarm", since the latter is an implementation detail.

pulumi/__main__.py
```diff
@@ -3,22 +3,35 @@
...
 from infra.dgraph_ttl import DGraphTTL
 from infra.engagement_creator import EngagementCreator
 from infra.metric_forwarder import MetricForwarder
 from infra.network import Network
 from infra.secret import JWTSecret
 from infra.service_queue import ServiceQueue
 
-if __name__ == "__main__":
 
+def _get_dgraph_cluster(network: Network) -> DgraphCluster:
```
christophermaier, line 15: For overall clarity, I'd recommend calling this something like `_create_dgraph_cluster`. In Pulumi, getter functions are pretty exclusively associated with retrieving a reference to already existing infrastructure, which this is definitely not doing.

pulumi/__main__.py
```diff
@@ -89,3 +107,7 @@
         user.local_grapl_user(
             dynamodb_tables.user_auth_table, "grapluser", "graplpassword"
         )
+
+
+if __name__ == "__main__":
+    main()
```
christophermaier, line 113: Nice 👍

pulumi/infra/bucket.py
```diff
@@ -28,13 +30,14 @@ def __init__(
         opts: `pulumi.ResourceOptions` for this resource.
 
         """
-        physical_bucket_name = bucket_physical_name(logical_bucket_name)
+        self.logical_bucket_name = logical_bucket_name
```
christophermaier, line 33: Why was this needed?

pulumi/infra/connectable.py
```diff
@@ -0,0 +1,14 @@
...
+from typing import Optional, Sequence
+
+import pulumi_aws as aws
+from pulumi_aws.ec2 import outputs
+from typing_extensions import Protocol
+
+import pulumi
+
+
+class SwarmConnectable(Protocol):
```
christophermaier, line 12: Is this used anywhere?

pulumi/infra/policies.py
```diff
@@ -0,0 +1,156 @@
...
+                            "Effect": "Allow",
+                            "Action": [
+                                "logs:CreateLogGroup",
+                                "logs:CreateLogStream",
+                                "logs:PutLogEvents",
+                                "logs:DescribeLogStreams",
+                            ],
+                            "Resource": f"{arn}:*",
+                        }
+                    ],
```
christophermaier, line 145: Looking closer, this policy appears to be a strict subset of the managed policy created above, and we attach it to the same Role. I don't think this is actually needed.

pulumi/infra/swarm.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+
+
+class Swarm(pulumi.ComponentResource):
+    def __init__(
+        self,
+        name: str,
+        vpc: Vpc,
+        internal_service_ports: Sequence[Ec2Port],
+        log_group: aws.cloudwatch.LogGroup,
```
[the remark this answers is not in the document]
wimax-grapl: blindly cloning existing infra instead of thinking about it :)

pulumi/infra/swarm.py
```diff
@@ -0,0 +1,209 @@
...
+            vpcs=[
+                aws.route53.ZoneVpcArgs(
+                    vpc_id=vpc.id,
+                )
+            ],
+            opts=child_opts,
+        )
+
+        self.swarm_config_bucket = Bucket(
+            logical_bucket_name="swarm-config-bucket",
```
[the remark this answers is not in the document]
wimax-grapl: this one has to be fixed, for now, for graplctl. May change in followup PR.

pulumi/infra/swarm.py
wimax-grapl: Ehhhhh it's repeated 3 times because of throwaway localgrapl stuff

pulumi/infra/swarm.py
wimax-grapl: yeah, sure

pulumi/infra/swarm.py
wimax-grapl: i like it

pulumi/infra/swarm.py
wimax-grapl: adding a comment

pulumi/infra/swarm.py
jgrillo-grapl: probably both `__str__` and `__repr__` should be like `f"{protocol}:{port}"`?

jgrillo-grapl (approved):
LGTM

pulumi/infra/swarm.py
christophermaier: Yeah, that's more what I was thinking, too.

`egress-from-graphql-endpoint-security-group-for-tcp:9080`
is nicer than
`egress-from-graphql-endpoint-security-group-for-Ec2Port(tcp:9080)`

christophermaier (approved):
Nice stuff!

wimax-grapl:
i feel like 9/10 times you don't need to update the repr,  shrug

PR #814, opened 2021-05-11
Run integration tests from Pants

christophermaier:
This removes a handful of custom containers that were used only to run pytest in our integration tests. Since this is all something that Pants can handle, we instead will let Pants run the tests for us.

It also removes several containers that are no longer needed because their usage was replaced by Pants at some point in the past.


jgrillo-grapl (approved):
👏

christophermaier:
Converted back to draft for now, since execution time is actually a bit longer.

wimax-grapl:
(cm is this actively being worked on?)

christophermaier:
@wimax-grapl kind of, but not really... I'll just close this to reduce the number of things people have to look at.

PR #818, opened 2021-05-13
Add GraphQL endpoint to Pulumi

christophermaier:
For some reason, this does not appear to work in LocalStack. As we're
moving away from LocalStack, we'll just preserve the existing
functionality we have in Local Grapl.


jgrillo-grapl (approved):
LGTM

PR #824, opened 2021-05-21
Add FargateService abstraction in Pulumi

christophermaier:
Interesting points:

- Adds a ECR Repository abstraction for container images
- Builds container images for Fargate services during the Pulumi run,
  uploading them to their own ECR repositories.

  This requires running `pulumi up` with `DOCKER_BUILDKIT=1` for some
  reason, despite allegedly being able to control this from within the
  code.

  Images are tagged with a checksum, and that specific tag is used in
  the Fargate service definitions, so we don't need to bother with
  tags beyond that (just as with CDK).
- Logging environment variables are able to be overridden in Pulumi
  stack configuration files on a per-service, per-variable basis.
- Fargate autoscaling has *not* been ported over from CDK; the details
  still need to be refined a bit.
- Since service queues are created inside the Fargate services, and
  since we're not running Fargate services in Local Grapl, we have to
  take care to explicitly create the queues independently in Local Grapl.
- Assorted IAM-related helper APIs have been added where appropriate.


pulumi/__main__.py
```diff
@@ -33,48 +38,106 @@ def main() -> None:
...
-    for service in services:
-        ServiceQueue(service)
 
-    forwarder = MetricForwarder(network=network)
+    if LOCAL_GRAPL:
+        # We need to create these queues in Local Grapl, because they
+        # are otherwise created in the FargateService instances below;
+        # we don't run Fargate services in Local Grapl.
+        #
+        # T_T
```
wimax-grapl, line 68: lol

pulumi/infra/emitter.py
```diff
@@ -24,6 +25,59 @@ def __init__(
             logical_bucket_name, sse=True, opts=pulumi.ResourceOptions(parent=self)
         )
 
+        # TODO: Unclear if this should be attached to our Bucket
+        # abstraction or here. Will keep it here for the time being.
+        #
+        # One nice thing about keeping it here is that we aren't
+        # creating policies for buckets that won't need them.
+        self.bucket_object_read_policy = aws.iam.Policy(
```
wimax-grapl, line 33: Wouldn't `grant_get_to` work for this?

pulumi/infra/fargate_service.py
```diff
@@ -0,0 +1,344 @@
...
+        #
+        # TODO: Consider a helper function for generating log group
+        # names that adheres to this convention for all our services
+        # (though this will be less of an issue once we migrate to
+        # Kafka)
+        self.log_group = aws.cloudwatch.LogGroup(
+            f"{name}-log-group",
+            name=f"/grapl/{DEPLOYMENT_NAME}/{name}",
+            opts=pulumi.ResourceOptions(parent=self),
+        )
```
wimax-grapl, line 136: Please specify 
`            retention_in_days=SERVICE_LOG_RETENTION_DAYS,`

pulumi/infra/fargate_service.py
```diff
@@ -0,0 +1,344 @@
...
+                            "Principal": {"Service": "ecs-tasks.amazonaws.com"},
+                        }
+                    ],
+                }
+            ),
+            opts=opts,
+        )
+
+
+class AWSFargateService(pulumi.ComponentResource):
```
wimax-grapl, line 64: Can we maybe make this `_AWSFargateService` just so nobody tries to consume  it instead of the subclass?

pulumi/infra/graph_merger.py
```diff
@@ -0,0 +1,66 @@
...
+        db.schema_properties_table.grant_read_permissions_to(
+            self.default_service.task_role
+        )
+        db.schema_properties_table.grant_read_permissions_to(
+            self.retry_service.task_role
+        )
+
+        # TODO: Interestingly, the CDK code doesn't have this, even though
+        # the other services do.
+        # self.allow_egress_to_cache(cache)
```
wimax-grapl, line 66: delete or no?

pulumi/infra/node_identifier.py
```diff
@@ -0,0 +1,82 @@
...
+            db.static_mapping_table,
+            db.dynamic_session_table,
+            db.process_history_table,
+            db.file_history_table,
+            db.inbound_connection_history_table,
+            db.outbound_connection_history_table,
+            db.network_connection_history_table,
+            db.ip_connection_history_table,
+            db.asset_id_mappings,
+        ):
```
wimax-grapl, line 74: I think I'd prefer if the dynamodb object had some `self.all: List[tables]`

wimax-grapl: just worried about at some point having to upkeep multiple lists of all

pulumi/infra/service_queue.py
```diff
@@ -83,6 +84,140 @@ def redrive_policy(arn: pulumi.Output) -> str:
...
+
+
+def _queue_consumption_policy(queue_name: str, queue: aws.sqs.Queue) -> aws.iam.Policy:
+    """Create an IAM Policy to consume messages from the given SQS queue.
+
+    queue_name is folded into the Policy resource name, and has to be a string (rather than a Pulumi Output).
+    """
+    return aws.iam.Policy(
+        f"consume-from-{queue_name}",
+        # We interpolate on queue.name to get the physical name
```
wimax-grapl, line 173: (we probably will never have a situation where many things pull from same queue, right?)

pulumi/infra/graph_merger.py
christophermaier: I'd prefer to keep it for now as a note for when we go through to revise our IAM and networking setup.

pulumi/infra/node_identifier.py
christophermaier: Unfortunately, it's actually not all the tables 😦 

This also seems to be the only service with access to this many tables.

pulumi/infra/service_queue.py
christophermaier: Probably not... I don't think this code is relevant to that, though.

pulumi/infra/emitter.py
christophermaier: It's the same permissions, yes, but that is method is attached to a bucket directly, whereas this is attached to an EventEmitter (which I now think needs to be recast as a kind of "notifying bucket"). The former is also a `RolePolicy` while this is a `Policy`. 

When we end up reviewing the IAM roles and policies, I think these two definitely need to be harmonized, but I'd like to hold off on this for now, since it's a bit of a bigger change.

pulumi/infra/node_identifier.py
wimax-grapl: kk

wimax-grapl (approved):

PR #826, opened 2021-05-22
Upgrade to Pants 2.5

christophermaier:
Pants 2.5 adds the ability to lint and format shell scripts, so that's what we'll do.

To help find them, the `pants tailor` command was run to generate new `BUILD` files. As a result, we are also linting and formatting some additional Python scripts.

The auto-generation of `BUILD` files also pointed out the existence of several unused files, so those have been deleted.


jgrillo-grapl (approved):
LGTM

PR #827, opened 2021-05-22
Remove import_aware_opts abstraction

christophermaier:
Now that we're basically done with the CDK-to-Pulumi migration, this
is no longer needed.


jgrillo-grapl (approved):
LGTM

PR #828, opened 2021-05-24
Use RolePolicies, rather than RolePolicyAttachments

christophermaier:
You can only have so many IAM policies attached to any given role, which is now causing issues due to how we have been modeling our policies in a relatively granular way. Basically, we have too many policies to attach, and we're exceeding the limit.

Inline policies, using `RolePolicy`, don't have a numerical limit, which makes them a bit more flexible for our purposes. They do have overall size limits, though, which calls for a bit of rethinking. The main change made here in that regard is to grant permissions on multiple DynamoDB tables in a single policy, as opposed to a policy per table as we had been doing before. This gives us the most savings in the case of the node identifier service, which needs access to nine tables.

### DIFF
--- a/pulumi/__main__.py
+++ b/pulumi/__main__.py
@@ -153,10 +153,10 @@ def main() -> None:
         dgraph_cluster=dgraph_cluster,
     )
 
-    provisioner = Provisioner(
+    Provisioner(
         network=network,
         secret=secret,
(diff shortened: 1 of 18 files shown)


jgrillo-grapl (approved):
👍 this looks great

PR #830, opened 2021-05-24
IAM Permission "Expansion"

christophermaier:
Here, we consolidate a couple permission granting calls (one to eliminate a duplicated method, the other to eliminate a redundant grant).

Additionally, the wildcarded IAM activities for buckets have been expanded as a preliminary step to limiting them. Additional PRs will follow that eliminate unnecessary permissions; it is hoped that seeing them expanded first will make the reasoning more apparent.


jgrillo-grapl (approved):
LGTM

PR #835, opened 2021-05-24
Better tagged Name stuff in Pulumi

wimax-grapl:
None

Some things had repeated names, like "grapl-network", that show up in the Name column in ec2.
I've added deployment name there to make it easier to reason about these.

Did a pulumi up against aws


christophermaier (approved):
👍

PR #862, opened 2021-05-25
Node identifier cleanup and batch identification

d0nut-grapl:

https://github.com/grapl-security/issue-tracker/issues/324

Adds batch operations for a majority of the work that the node-identifier needs to process.

Integration tests


docker-compose.yml
```diff
@@ -30,7 +30,6 @@ x-common-variables:
     NETWORK_CONNECTION_HISTORY_TABLE: ${DEPLOYMENT_NAME}-network_connection_history_table
     OUTBOUND_CONNECTION_HISTORY_TABLE: ${DEPLOYMENT_NAME}-outbound_connection_history_table
     PROCESS_HISTORY_TABLE: ${DEPLOYMENT_NAME}-process_history_table
```
colin-grapl, line 32: We should be able to remove all of the tables except for the static/session ones.

src/js/grapl-cdk/lib/historydb.ts
```diff
@@ -125,7 +109,6 @@ export class HistoryDb extends cdk.Construct {
         );
         this.ip_connection_history.grantReadWriteData(service.event_handler);
         this.asset_history.grantReadWriteData(service.event_handler);
```
colin-grapl, line 111: We can remove all of these tables like 'network connection' etc. They are unused, as we are now general over Static and Session identified nodes.

src/rust/grapl-utils/src/rusoto_ext/dynamodb.rs
```diff
@@ -125,55 +121,175 @@ pub trait GraplDynamoDbClientExt: DynamoDb + Send + Sync {
...
+                            .into_iter()
+                            .map(|row_keys| (table_name.clone(), row_keys))
+                            .collect::<Vec<_>>()
+                    })
+                    .collect();
 
-                    pending_items.extend(unprocessed_keys);
-                }
-                _ => {}
+                pending_items.extend(unprocessed_keys);
```
colin-grapl, line 152: Instead of collecting into a Vec just to extend another Vec, you can extend from the iter, which should avoid the intermediary allocation.

src/rust/grapl-utils/src/rusoto_ext/dynamodb.rs
```diff
@@ -125,55 +121,175 @@ pub trait GraplDynamoDbClientExt: DynamoDb + Send + Sync {
...
+            request_items,
+            return_consumed_capacity,
+            return_item_collection_metrics
+        } = batch_write_item;
+
+        let mut pending_writes: Vec<_> = request_items.into_iter()
+            .flat_map(|(table_name, write_requests)| {
+                write_requests.into_iter()
+                    .map(|write_request| (table_name.clone(), write_request))
+                    .collect::<Vec<_>>()
```
colin-grapl, line 183: I think '.collect' is unnecessary. `flat_map` takes IntoIterator, so we can avoid allocating to a Vec just to turn it back into an iterator.

src/rust/grapl-utils/src/rusoto_ext/dynamodb.rs
```diff
@@ -125,55 +121,175 @@ pub trait GraplDynamoDbClientExt: DynamoDb + Send + Sync {
...
+            }).collect();
+
+        let mut total_consumed_capacity = Vec::new();
+        let mut item_collection_metrics = HashMap::new();
+
+        while !pending_writes.is_empty() {
+            let mut writes_to_process = HashMap::new();
+
+            pending_writes
```
colin-grapl, line 192: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.array_chunks

nightly only T_T

src/rust/grapl-utils/src/rusoto_ext/dynamodb.rs
```diff
@@ -125,55 +121,175 @@ pub trait GraplDynamoDbClientExt: DynamoDb + Send + Sync {
...
+            }
+
+            if let Some(metrics) = batch_write_response.item_collection_metrics {
+                metrics.into_iter()
+                    .for_each(|(table_name, item_metrics)| {
+                        item_collection_metrics.entry(table_name)
+                            .or_insert(Vec::new())
+                            .extend(item_metrics);
+                    });
+            }
```
colin-grapl, line 221: Feels like maybe what we want is some sort of BatchWriteMetricsWriter or something, which would just take a response and handle all of this stuff. That way our metrics / capacity code isn't clogging up our business logic.

src/rust/grapl-utils/src/rusoto_ext/dynamodb.rs
```diff
@@ -125,55 +121,175 @@ pub trait GraplDynamoDbClientExt: DynamoDb + Send + Sync {
...
+
+            if let Some(unprocessed_items) = batch_write_response.unprocessed_items {
+                let flattened_unprocessed_items: Vec<_> = unprocessed_items.into_iter()
+                    .flat_map(|(table_name, write_requests)| {
+                        write_requests.into_iter()
+                            .map(|write_request| (table_name.clone(), write_request))
+                            .collect::<Vec<_>>()
+                    }).collect();
+
+                pending_writes.extend(flattened_unprocessed_items);
```
colin-grapl, line 231: No need to collect just to extend. `extend` takes IntoIterator

colin-grapl (commented):
Just a quick first pass. I'll check it out more.

docker-compose.yml
d0nut-grapl: Awesome; I should've dug into this more but glad to hear we can delete these.

src/rust/grapl-utils/src/rusoto_ext/dynamodb.rs
d0nut-grapl: I don't understand this point. This isn't business logic; it's specifically for scaling up and handling dynamodb's request limits transparently.

d0nut-grapl:
Moving this to draft as, despite the passing tests, this is actually broken likely do to issues with writing to DynamoDB with duplicate keys in the batch requests.

PR #864, opened 2021-05-25
Add route tables to Pulumi network subnets. Still doesn't work, but it's an improvement

wimax-grapl:

Associate each subnet with a route table.
SSM still doesn't work, but this looks closer to correct than it was previously.

pulumi up


christophermaier (approved):
This seems OK for what it does, and based on my understanding of the relevant primitives.

So, yay? :tada:

PR #868, opened 2021-05-25
Provision pulumi's dgraph with graplctl. Doesn't work completely but gets up to making disk alarms.

wimax-grapl:
grapl-security/issue-tracker#441

- Lots of tagging stuff with a Name that makes it easier to reason about.
- Nesting with resource options

against my pulumi aws stack with graplctl!

### DIFF
--- a/pulumi/infra/swarm.py
+++ b/pulumi/infra/swarm.py
@@ -24,28 +24,46 @@ class Ec2Port:
 
     def allow_internally(
         self,
-    ) -> Tuple[aws.ec2.SecurityGroupIngressArgs, aws.ec2.SecurityGroupEgressArgs]:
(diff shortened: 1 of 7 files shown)


pulumi/infra/swarm.py
```diff
@@ -24,28 +24,46 @@ class Ec2Port:
...
             protocol=self.protocol,
             self=True,
+            opts=pulumi.ResourceOptions(parent=sg),
         )
+
         return (ingress, egress)
 
-    def allow_outbound_any_ip(self) -> aws.ec2.SecurityGroupEgressArgs:
-        return aws.ec2.SecurityGroupEgressArgs(
+    def allow_outbound_any_ip(
```
christophermaier, line 53: We don't actually need to return anything from this function, or from `allow_internally`

PR #888, opened 2021-05-27
Rework an e2e assertion to account for multiple files

christophermaier:
Originally, the directory in question only had one file. Recently, a
`BUILD` file was added, which violated the assumptions of this
particular assertion. It has now been rewritten.


jgrillo-grapl (approved):
👍

PR #889, opened 2021-05-27
Expose the API invocation URL as a property on the Api class

christophermaier:
Previously this information was only available as a stack output, but
we now have need of it within the Pulumi run, so it needs to be
exposed as a property.


jgrillo-grapl (approved):
👏

jgrillo-grapl (approved):

PR #901, opened 2021-06-03
Better waiting for disk metrics in graplctl

wimax-grapl:
None

There's a step where we try to create Cloudwatch alarms based off EC2 metrics.
Previously we waited a straight 5 minutes.
Now, we check for valid state every 30 seconds, up to 5 minutes.

against aws

### DIFF
--- a/src/python/grapl-common/grapl_common/retry.py
+++ b/src/python/grapl-common/grapl_common/retry.py
@@ -1,6 +1,6 @@
 import time
 from functools import wraps
-from typing import Callable, Type, TypeVar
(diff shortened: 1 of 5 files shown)


jgrillo-grapl (approved):

src/python/grapl-common/grapl_common/retry.py
```diff
@@ -1,6 +1,6 @@
 import time
```
wimax-grapl, line 1: I was going to use `retry` - ended up not doing so - but moving it to `-common` and giving it better typing is a worthy addition

PR #902, opened 2021-06-04
Dgraph in Pulumi: IAM fixes so it actually deploys. + Graplctl SSM sane-ification

wimax-grapl:

- Graplctl now retries when it encounters InvalidPluginName; this was stolen from jgrillo's ec2 PR (and pulled out to a function).
- When we use a bucket as source for an SSM command, we need to grant different perms to it.

make graplctl && graplctl dgraph create -t i3.large

### DIFF
--- a/src/python/graplctl/graplctl/common.py
+++ b/src/python/graplctl/graplctl/common.py
@@ -110,10 +110,10 @@ def get_command_results(
             break
 
     for instance_id in instance_ids:
-        invocation = ssm.get_command_invocation(
-            CommandId=command_id,
-            InstanceId=instance_id,
-            PluginName="runShellScript",
+        invocation = get_command_invocation(
+            ssm=ssm,
+            command_id=command_id,
+            instance_id=instance_id,
         )
 
         if invocation["Status"] == "Success":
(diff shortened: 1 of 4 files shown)


src/python/graplctl/graplctl/common.py
```diff
@@ -122,6 +122,37 @@ def get_command_results(
...
+    ssm: SSMClient,
+    command_id: str,
+    instance_id: str,
+) -> GetCommandInvocationResultTypeDef:
+    """
+    get-command-invocation commonly throws an awfully named exception: InvalidPluginName.
+    The actual meaning? This invocation hasn't become available.
+    This wrapper makes it a bit more sane to use.
+    """
+    LOGGER.info(
```
wimax-grapl, line 135: these should probably be .debug

src/python/graplctl/graplctl/common.py
```diff
@@ -110,17 +110,46 @@ def get_command_results(
...
+        f"retrieving invocation metadata for command {command_id} on instance {instance_id}"
+    )
+    while True:
```
christophermaier, line 138: Should there be some kind of maximum amount of time to wait before finally erroring out?

christophermaier (approved):
👍

jgrillo-grapl (approved):
LGTM

PR #909, opened 2021-06-07
Copy over colin's S3 fix; pull Ec2port out to its own file

wimax-grapl:
none

Steal two useful fixes/refactors from #800

colin said s3 behavior was better after this

### DIFF
--- a/pulumi/infra/bucket.py
+++ b/pulumi/infra/bucket.py
@@ -4,7 +4,6 @@ from typing import List, Optional
 
 import pulumi_aws as aws
 from infra.config import DEPLOYMENT_NAME
-from infra.ec2 import Ec2Port
 
 import pulumi
 from pulumi.resource import ResourceOptions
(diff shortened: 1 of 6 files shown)


jgrillo-grapl (approved):

pulumi/infra/bucket.py
```diff
@@ -182,13 +185,19 @@ def upload_to_bucket(
         "some_dir/subdir/c.txt"
         """
         if file_path.is_file():
-            return [self._upload_file_to_bucket(file_path, root_path=file_path.parent)]
+            if preserve_path:
+                root_path = preserve_path
+            else:
+                root_path = file_path.parent
```
christophermaier, line 191: Maybe `root_path = preserve_path or file_path.parent` instead?

christophermaier (approved):
👍 with a small suggestion.

PR #910, opened 2021-06-08
Use 443 outbound by default on all Lambda Services, Fargate Services

wimax-grapl:
none, general pulumi infra

443 outbound is enabled by default for lambda and fargate services.
also, I fixed a few "oh this is enabled for Default but not Retry" bugs.

pulumi up --preview

### DIFF
--- a/pulumi/infra/ec2.py
+++ b/pulumi/infra/ec2.py
@@ -40,7 +40,7 @@ class Ec2Port:
 
     def allow_outbound_any_ip(self, sg: aws.ec2.SecurityGroup) -> None:
         aws.ec2.SecurityGroupRule(
-            f"outbound-any-ip-egress-{self}",
+            f"outbound-any-ip-egress-{self}-to-{sg._name}",
             type="egress",
             security_group_id=sg.id,
             from_port=self.port,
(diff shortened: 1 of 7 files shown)


jgrillo-grapl (approved):

pulumi/infra/ec2.py
```diff
@@ -40,7 +40,7 @@ def allow_internally(
 
     def allow_outbound_any_ip(self, sg: aws.ec2.SecurityGroup) -> None:
         aws.ec2.SecurityGroupRule(
-            f"outbound-any-ip-egress-{self}",
+            f"outbound-any-ip-egress-{self}-to-{sg._name}",
```
christophermaier, line 43: Can't comment on the actual code because it's outside Github's commenting window, but I think what is currently the `__str__` implementation for `Ec2Port` is closer to (but not quite) a `__repr__` implementation. I think having a `__str__` of

``` python
def __str__(self) -> str:
    return f"{self.protocol}:{self.port}"
```
would yield a more readable (and shorter) Pulumi resource name. 

Compare:
```
outbound-any-ip-egress-tcp:443-to-my-security-group
```
with
```
outbound-any-ip-egress-Ec2Port(tcp:443)-to-my-security-group
```

christophermaier (approved):
👍 with a small suggestion.

PR #912, opened 2021-06-09
fix typecheck error in main

wimax-grapl:
```
src/python/graplctl/graplctl/cli.py:27: error: Argument 1 to "Choice" has incompatible type "KeysView[str]"; expected "Sequence[str]"  [arg-type]
```

Sequence implies you can `seq[idx]` on it


jgrillo-grapl (approved):
Could we pin our click dep as well?

jgrillo-grapl (approved):
yolo

3rdparty/python/README.md
```diff
@@ -0,0 +1,9 @@
+# requirements.txt
+Declare all your dependencies.
+Ideally, we'd pin everything in here, but it's not the end of the universe if a few are unpinned.
```
christophermaier, line 3: Things should probably only be pinned in here if they need to be for some reason (and that reason should be documented in a comment).

wimax-grapl: updating readme

PR #928, opened 2021-06-17
Minimize eventlog

colin-grapl:
This is just a stripped down version of the eventlog.xml .

``` Generated like this
❯ rg 'EventID>2<|EventID>1<|EventID>11' ./etc/sample_data/eventlog.xml | rg 'chrome|svchost|dropper|cmd.exe' > ./min_eventlog.xml 
```
Basically keep Process and File creation events, and Process Termination events, and then only the ones for chrome, svchost, dropper, cmd. We can almost certainly strip down another 20+ events, but this was just a trivial first pass.

This takes us from >700 lines to 36.


wimax-grapl (commented):
Some thoughts:
1. Keep the original `eventlog.xml` around, at least for some stress testing. Also, it helps exercise the chopping-up-of-uploads we do in `graplctl sysmon upload`; that's why your unit test is failing (it expects 6 calls of s3 upload). 
2. maybe we just rename these to how many events there are in them (37_events.sysmon.xml) or something

inickles-grapl:
> 1. Keep the original `eventlog.xml` around, at least for some stress testing. Also, it helps exercise the chopping-up-of-uploads we do in `graplctl sysmon upload`; that's why your unit test is failing (it expects 6 calls of s3 upload).
> 2. maybe we just rename these to how many events there are in them (37_events.sysmon.xml) or something

I agree we should keep a copy of the original for other kinds of tests like you mention 👍 . My preference would be to use files name that speak more to what they're used for, so something like "sysmon_dropper_events.xml", "sysmon_stress_events.xml".

Additionally, and for a later PR, I am of these opinions: 
1. Test data for an analyzer should live in the same src directory as the analyzer.
2. Analyzers src should live in the plugins/generator src directory that build the model that's queried in the analyzer.

colin-grapl:
Didn't realize I'd deleted the original.

jgrillo-grapl (approved):

PR #931, opened 2021-06-17
Adding missing GRAPL_SCHEMA_PROPERTIES_TABLE env var to graphql lambda.

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/578

This sets GRAPL_SCHEMA_PROPERTIES_TABLE for the GraphQL Endpoint lambda, which is needed.

I updated a deployment with these changes and eye-balled the env var in the Lambda console was set with appropriate value. I was unable to use pulumi to update the deployment where https://github.com/grapl-security/issue-tracker/issues/578 was originally found, but I verified a similar fix applied manually resolved the issue.

### DIFF
--- a/pulumi/infra/graphql.py
+++ b/pulumi/infra/graphql.py
@@ -55,6 +55,7 @@ class GraphQL(pulumi.ComponentResource):
                     "UX_BUCKET_URL": pulumi.Output.concat(
                         "https://", ux_bucket.bucket_regional_domain_name
                     ),
+                    "GRAPL_SCHEMA_PROPERTIES_TABLE": db.schema_properties_table.name,
                 },
                 timeout=30,
                 memory_size=128,


jgrillo-grapl (approved):

pulumi/infra/graphql.py
```diff
@@ -55,6 +55,7 @@ def __init__(
                     "UX_BUCKET_URL": pulumi.Output.concat(
                         "https://", ux_bucket.bucket_regional_domain_name
                     ),
+                    "GRAPL_SCHEMA_PROPERTIES_TABLE": db.schema_properties_table,
```
jgrillo-grapl, line 58: this appears to be failing the typecheck, is there a `.name` property we can dereference here?

jgrillo-grapl:
link https://github.com/grapl-security/issue-tracker/issues/576

PR #934, opened 2021-06-18
mark the expected data in dgraph test as pytest.xfail

jgrillo-grapl:
### Which issue does this PR correspond to?
https://github.com/grapl-security/issue-tracker/issues/576

### What changes does this PR make to Grapl? Why?
marks the `test_expected_data_in_dgraph` test with `@pytest.mark.xfail` s.t. the test suite passes in AWS

### How were these changes tested?
`graplctl aws test`
logs from e2e-test-runner lambda:
```
START RequestId: 48c00858-0d12-4322-a6d2-c1f9d2a1da90 Version: $LATEST
--
============================= test session starts ==============================
platform linux -- Python 3.7.10, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /var/task
collected 8 items
e2e_test_runner/test_main.py [DEBUG]	2021-06-18T19:47:07.565Z	48c00858-0d12-4322-a6d2-c1f9d2a1da90	created EngagementEdgeClient for endpoint https://m4h6fr431j.execute-api.us-east-1.amazonaws.com:443/prod/auth
.[DEBUG]	2021-06-18T19:47:09.741Z	48c00858-0d12-4322-a6d2-c1f9d2a1da90	created EngagementEdgeClient for endpoint https://m4h6fr431j.execute-api.us-east-1.amazonaws.com:443/prod/auth
.[DEBUG]	2021-06-18T19:47:09.861Z	48c00858-0d12-4322-a6d2-c1f9d2a1da90	created EngagementEdgeClient for endpoint https://m4h6fr431j.execute-api.us-east-1.amazonaws.com:443/prod/auth
[DEBUG]	2021-06-18T19:47:09.861Z	48c00858-0d12-4322-a6d2-c1f9d2a1da90	retrieving jwt cookie
[DEBUG]	2021-06-18T19:47:10.523Z	48c00858-0d12-4322-a6d2-c1f9d2a1da90	retrieving jgrillo-sandbox-TestUserPassword
[DEBUG]	2021-06-18T19:47:10.685Z	48c00858-0d12-4322-a6d2-c1f9d2a1da90	logging in with user jgrillo-sandbox-grapl-test-user at https://m4h6fr431j.execute-api.us-east-1.amazonaws.com:443/prod/auth/login
[DEBUG]	2021-06-18T19:47:13.791Z	48c00858-0d12-4322-a6d2-c1f9d2a1da90	retrieved jwt cookie
[...]

PR #936, opened 2021-06-20
e2e (or rather, test-with-env) exit codes seem to work now

wimax-grapl:
Sorta has to do with https://github.com/grapl-security/grapl/pull/934.

Jesse discovered that despite e2e failing, our CI reports success.
Heads up: I think this PR *will* fail e2e in CI; we need to land #934, #935 in some order

- More robust propagation of exit codes from docker -> docker-compose-with-error -> makefile

- introduce an `exit(42)` into e2e lambda; it fails.
- undo it; it succeeds.
- this was not the case before this change.
- (also, i tested it with https://github.com/grapl-security/grapl/pull/935 )

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,7 +247,6 @@ test-e2e: build-test-e2e ## Build and run e2e tests
 .PHONY: test-with-env
 test-with-env: # (Do not include help text - not to be used directly)
 	stopGrapl() {
-		EXIT_CODE=$$?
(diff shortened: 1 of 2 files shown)


inickles-grapl (approved):

jgrillo-grapl (approved):

jgrillo-grapl:
linking this https://github.com/grapl-security/issue-tracker/issues/576

PR #938, opened 2021-06-22
Internal 576 fix graphql test client

jgrillo-grapl:
https://github.com/grapl-security/issue-tracker/issues/576

some updates to make the e2e-test-runner work, fix some logging issues, make `graplctl aws test` exit 0 upon success

`graplctl aws test`, `make test-e2e`

### DIFF
--- a/src/python/e2e-test-runner/e2e_test_runner/main.py
+++ b/src/python/e2e-test-runner/e2e_test_runner/main.py
@@ -1,7 +1,6 @@
 import os
 import shutil
 import sys
-
 from typing import Any
 
 from grapl_common.debugger.vsc_debugger import wait_for_vsc_debugger
--- a/src/python/e2e-test-runner/e2e_test_runner/main.py
+++ b/src/python/e2e-test-runner/e2e_test_runner/main.py
@@ -1,15 +1,33 @@
 import os
 import shutil
+import sys
+
 from typing import Any
 
 from grapl_common.debugger.vsc_debugger import wait_for_vsc_debugger
-from grapl_tests_common import setup_tests
+from grapl_common.grapl_logger import get_module_grapl_logger
+
+LOGGER = get_module_grapl_logger()
 
 
 def lambda_handler(event: Any, context: Any) -> None:
     wait_for_vsc_debugger("grapl_e2e_tests")
+    LOGGER.info("staging schemas")
     if os.path.exists("/tmp/schemas"):
         shutil.rmtree("/tmp/schemas")
     os.mkdir("/tmp/schemas")
     shutil.copyfile("e2e_test_runner/schemas.py", "/tmp/schemas/schemas.py")
-    setup_tests.exec_pytest()
+    LOGGER.info("staged schemas")
+
+    LOGGER.info("executing pytest")
+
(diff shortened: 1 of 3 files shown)


src/python/grapl-tests-common/grapl_tests_common/setup_tests.py
```diff
@@ -13,6 +13,7 @@
 DUMP_ARTIFACTS = bool(environ.get("DUMP_ARTIFACTS", False))
 
 LOGGER = get_module_grapl_logger()
+GRAPL_LOG_LEVEL = os.getenv("GRAPL_LOG_LEVEL", "INFO")
```
wimax-grapl, line 16: ?

src/python/grapl-common/grapl_common/grapl_logger.py
```diff
@@ -4,8 +4,8 @@
 import sys
 
 
-def get_module_grapl_logger(
-    default_log_level: str = "INFO", log_to_stdout: bool = False
+def get_module_grapl_logger(  # TODO: delete this function
```
wimax-grapl, line 7: can we get a proposed solution and/or a stub ticket for this?

i am not attached to this particular solution; I _am_ attached to a mostly-universal way to grab a logger that knows about GRAPL_LOG_LEVEL.

src/python/e2e-test-runner/e2e_test_runner/main.py
```diff
@@ -1,15 +1,32 @@
...
+    LOGGER.info("executing pytest")
+
+    from grapl_tests_common import setup_tests  # import here to limit monkeypatch
+
+    result = setup_tests.exec_pytest()
+
+    LOGGER.info(f"tests completed with status code {result}")
+
+    if result == 0:
+        return  # Lambda runtime will flip out if sys.exit(0) -_-
```
wimax-grapl, line 30: LMFAO WHAT

src/python/grapl-common/grapl_common/grapl_logger.py
jgrillo-grapl: The proposed solution is to adhere as closely as possible to the python documentation. In particular:

> Note that Loggers should NEVER be instantiated directly, but always through the module-level function logging.getLogger(name).

https://docs.python.org/3/library/logging.html#logger-objects

So, everywhere we need a logger, we should do the following:
```python
GRAPL_LOG_LEVEL = os.getenv("GRAPL_LOG_LEVEL", "INFO")
LOGGER = logging.getLogger(__name__)
LOGGER.setLevel(GRAPL_LOG_LEVEL)
LOGGER.addHandler(logging.StreamHandler(stream=sys.stdout))
```
there's nothing wrong with a little repetition here. Making this simple and explicit is much more important than making it convenient and clever.

src/python/grapl-tests-common/grapl_tests_common/setup_tests.py
jgrillo-grapl: Is there a question? I don't understand.

wimax-grapl: oh duh you do actually use this, okay

wimax-grapl (approved):

PR #945, opened 2021-06-24
Resolve an issue where Sysmon generator was not adding to `CompletedEvents`

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/590

This has the Sysmon generator update the `impl EventHandler` `CompleteEvents` parameter, which means it'll now start caching successfully processed events, and it'll fix a log line that reported number of processed events.

While I was at it, I figured I'd:
- update the logging to use tracing
- clean up project dependencies in Cargo.toml

I'm mostly relying on CI tests.


wimax-grapl (approved):

src/rust/generators/sysmon-subgraph-generator/src/models/mod.rs
```diff
@@ -71,20 +71,20 @@ impl SysmonTryFrom<Event> for GraphDescription {
...
+                    tracing::warn!(message="Failed to process outbound network event.", error=?e);
                 }
 
                 result
             }
             unsupported_event => {
                 let message = format!("Unsupported event_type: {:?}", unsupported_event);
 
-                warn!("{}", message);
+                tracing::warn!(message = message.as_str());
```
colin-grapl, line 87: Just fwiw you can do %message I believe and it'll work, but not important or worth changing.

colin-grapl (approved):

src/rust/generators/sysmon-subgraph-generator/src/models/mod.rs
inickles-grapl: I like that. Resolved.

PR #946, opened 2021-06-24
Refactor OSQuery generator

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/590

What started as a fix to https://github.com/grapl-security/issue-tracker/issues/590 turned into a significant refactor of this generator. Instead of multi-stage and lazy event parsing, we do deserialize at once, which not only allows us to resolve the issue (by allowing us to derive a cache key from hashing fields) it also significantly simplifies the code.

While I was at it, I cleaned up a number of things:
- changed the Ndjson deserializer to log deserialization failures and drop, so event handlers just need to handle events.
- switched logging to use `tracing`
- switched to `thiserror`
- cleaned up dependencies in Cargo.toml

I'm relying on CI tests.


src/rust/generators/osquery-subgraph-generator/src/parsers/grapl_pack/files.rs
```diff
@@ -94,6 +91,18 @@ impl TryFrom<OSQueryResponse<OSQueryFileQuery>> for GraphDescription {
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::parsers::OSQueryEvent;
+
+    #[test]
+    fn parse_pack_grapl_files_json() {
+        let test_json = r#"{"name":"pack_grapl_files","hostIdentifier":"ip-172-31-46-189.us-east-2.compute.internal","calendarTime":"Thu Oct 22 21:05:26 2020 UTC","unixTime":1603400726,"epoch":0,"counter":0,"numerics":false,"columns":{"action":"ACCESSED","atime":"1599180507","category":"temp","ctime":"1603395939","gid":"0","hashed":"0","inode":"101","md5":"","mode":"1777","mtime":"1603395939","sha1":"","sha256":"","size":"4096","target_path":"/tmp/","time":"1603395948","transaction_id":"0","uid":"0"},"action":"removed"}"#;
```
wimax-grapl, line 104: minor nitpick: feels like something like this (many characters-long line) might be better off as a fixture file that gets read

src/rust/generators/osquery-subgraph-generator/src/parsers/grapl_pack/files.rs
```diff
@@ -94,6 +91,18 @@ impl TryFrom<OSQueryResponse<OSQueryFileQuery>> for GraphDescription {
         graph.add_node(asset);
         graph.add_node(subject_file);
 
-        Ok(graph)
+        graph
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
wimax-grapl, line 99: hell yeah

src/rust/generators/osquery-subgraph-generator/src/parsers/grapl_pack/files.rs
wimax-grapl: this might be what read_osquery_test_data is for? though that seems to take in zstd'd stuff

wimax-grapl (approved):

src/rust/generators/osquery-subgraph-generator/src/parsers/grapl_pack/files.rs
inickles-grapl: I'm into that idea, putting this into a file. I'll do that, thanks.

colin-grapl: An easy way to do this is to put it in a file and then use the `include_str!` macro:
https://doc.rust-lang.org/std/macro.include_str.html

src/rust/generators/osquery-subgraph-generator/src/generator.rs
```diff
@@ -20,89 +10,78 @@ use sqs_executor::{
...
-        let final_subgraph = subgraphs
+        let final_subgraph = events
             .into_iter()
-            .filter_map(|subgraph| subgraph.ok())
+            .map(|event| {
+                completed.add_identity(&event, EventStatus::Success);
+                GraphDescription::from(event)
+            })
             .fold(GraphDescription::new(), |mut current_graph, subgraph| {
                 current_graph.merge(&subgraph);
```
colin-grapl, line 75: Minor, and I'll approve regardless, but since `subgraph` is being moved you can call the more efficient `merge_into`.

src/rust/generators/osquery-subgraph-generator/src/generator.rs
```diff
@@ -20,89 +10,78 @@ use sqs_executor::{
...
-            Ok(final_subgraph)
-        } else {
-            let sqs_executor_error = errors
-                .into_iter()
-                .map(|err| OSQuerySubgraphGeneratorError::Unexpected(err))
-                .next()
-                .unwrap();
+        tracing::info!(
+            message = "Completed mapping subgraphs",
+            num_completed = completed.len()
```
colin-grapl, line 81: Might be nice to add the input count to this as well, nbd.

src/rust/generators/osquery-subgraph-generator/src/parsers/grapl_pack/files.rs
```diff
@@ -94,6 +91,18 @@ impl TryFrom<OSQueryResponse<OSQueryFileQuery>> for GraphDescription {
...
+    use crate::parsers::OSQueryEvent;
+
+    #[test]
+    fn parse_pack_grapl_files_json() {
+        let test_json = r#"{"name":"pack_grapl_files","hostIdentifier":"ip-172-31-46-189.us-east-2.compute.internal","calendarTime":"Thu Oct 22 21:05:26 2020 UTC","unixTime":1603400726,"epoch":0,"counter":0,"numerics":false,"columns":{"action":"ACCESSED","atime":"1599180507","category":"temp","ctime":"1603395939","gid":"0","hashed":"0","inode":"101","md5":"","mode":"1777","mtime":"1603395939","sha1":"","sha256":"","size":"4096","target_path":"/tmp/","time":"1603395948","transaction_id":"0","uid":"0"},"action":"removed"}"#;
+
+        let _: OSQueryEvent = serde_json::from_str(&test_json).unwrap();
```
colin-grapl, line 106: Even in tests I'd recommend against unwrap - prefer 'expect', which at least will give you a message, even if it's a trivial one like "serde_json::from_str failed".

src/rust/generators/osquery-subgraph-generator/src/parsers/mod.rs
```diff
@@ -1,108 +1,36 @@
...
+    #[serde(rename = "pack_grapl_files")]
+    File(grapl_pack::files::FileEvent),
 }
 
-#[derive(Serialize, Deserialize)]
-enum OSQueryAction {
-    #[serde(rename = "added")]
+#[derive(Serialize, Deserialize, Debug, Clone, Hash)]
+#[serde(rename_all = "camelCase")]
+pub(self) enum OSQueryAction {
```
colin-grapl, line 22: TIL pub(self)
https://doc.rust-lang.org/reference/visibility-and-privacy.html

And apparently there are even more - I had no idea there were more than pub(crate).

src/rust/grapl-service/src/decoder/ndjson.rs
```diff
@@ -1,31 +1,67 @@
...
+                        }
+                        None
+                    }
+                }
+            })
             .collect();
 
-        Ok(results)
+        match first_deserialization_error {
+            Some(error) if events.is_empty() => Err(error.into()),
```
colin-grapl, line 63: Can we just add a comment here? The interfaces in sqs-executor are so bad it seems worth pointing out that the in the case where events is *not* empty we still want to make forward progress, so we just log the errors but otherwise pass what we can forward.

colin-grapl (approved):

src/rust/generators/osquery-subgraph-generator/src/generator.rs
inickles-grapl: I have https://github.com/grapl-security/grapl/pull/946/files#diff-3aae8671eae60cb82cbeefa6b5f3cbff38689e3db8a22e200d13d7db49352ebbR60-R63. Were you thinking of something different?

src/rust/generators/osquery-subgraph-generator/src/parsers/grapl_pack/files.rs
inickles-grapl: @colin-grapl oh cool. i'll keep in the mind for next time, but i already committed a solution.

src/rust/generators/osquery-subgraph-generator/src/generator.rs
```diff
@@ -20,89 +10,78 @@ use sqs_executor::{
...
     metrics::OSQuerySubgraphGeneratorMetrics,
-    parsers::PartiallyDeserializedOSQueryLog,
+    parsers::OSQueryEvent,
 };
 
 #[derive(Clone)]
-pub struct OSQuerySubgraphGenerator<C: Cache> {
+pub struct OSQuerySubgraphGenerator<C>
+where
+    C: Cache + Clone + Send + Sync + 'static,
```
d0nut-grapl, line 24: Cache is defined as 

```
pub trait Cache: Clone + Send + Sync
```

So you shouldn't need to add `+ Clone + Send + Sync` here as part of the trait bounds. I also assume that `'static` isn't needed here either.

d0nut-grapl (approved):

PR #947, opened 2021-06-24
Change Sysmon generator event handler to only take a Vec of events only

inickles-grapl:

Change Sysmon generator event handler to only take a Vec of events only, instead of Vec<Result<Event, Error>>.

Also, change the Sysmon parser to filter for supported events after parsing. This should help ensure we don't filter events that we shouldn't be.

I'm relying on CI tests.

### DIFF
--- a/src/rust/generators/sysmon-subgraph-generator/src/serialization.rs
+++ b/src/rust/generators/sysmon-subgraph-generator/src/serialization.rs
@@ -33,43 +33,59 @@ impl CheckedError for SysmonDecoderError {
 #[derive(Debug, Clone, Default)]
 pub struct SysmonDecoder;
 
-impl PayloadDecoder<Vec<Result<Event, SysmonDecoderError>>> for SysmonDecoder {
+impl PayloadDecoder<Vec<Event>> for SysmonDecoder {
     type DecoderError = SysmonDecoderError;
 
-    fn decode(
-        &mut self,
-        body: Vec<u8>,
-    ) -> Result<Vec<Result<Event, Self::DecoderError>>, Self::DecoderError> {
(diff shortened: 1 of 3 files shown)


src/rust/generators/sysmon-subgraph-generator/src/serialization.rs
```diff
@@ -33,43 +33,57 @@ impl CheckedError for SysmonDecoderError {
...
-        let events: Vec<Result<Event, Self::DecoderError>> = decompressed
+        let events: Vec<_> = decompressed
             .split(|byte| *byte == b'\n')
             .map(String::from_utf8_lossy)
             .map(|s| s.to_string())
-            .filter(|event| {
-                (!event.is_empty() && event != "\n")
-                    && (event.contains("EventID>1<")
-                        || event.contains("EventID>3<")
-                        || event.contains("EventID>11<"))
```
wimax-grapl, line 65: 😱 good riddance

src/rust/generators/sysmon-subgraph-generator/src/serialization.rs
```diff
@@ -33,43 +33,57 @@ impl CheckedError for SysmonDecoderError {
...
+            .filter_map(|event_str| {
+                let parsed_event = Event::from_str(&event_str);
+                match parsed_event {
+                    Ok(event) => Some(event),
+                    Err(error) => {
+                        tracing::error!(message="Unable to deserialize Sysmon event", error=?error);
```
wimax-grapl, line 60: should we log the event_str?

wimax-grapl (approved):

src/rust/generators/sysmon-subgraph-generator/src/serialization.rs
inickles-grapl: Yeah, good idea. 👍  Done.

PR #954, opened 2021-06-25
Rust integration test fixes

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/594

Unfiled (at the moment):
The Rust integration tests were split into multiple containers, which adds unnecessary complexity and brittleness. An example of how this can cause issues can be seen with the latest addition of the kafka-metrics-exporter, which was not being built in the Dockerfile.

There are a number of small commits in this PR:
1. Consolidate a duplicate apt install line in the Rust Dockerfile
2. Add the kafka-metrics-exporter/integration feature to the cargo test build command.
3. Consolidate Rust integration tests to a single container.
4. ~Add the graph-merger integration tests~ Reverted, this apparently breaks other tests. I'll file a ticket for that.
5. Fix an issue in the graph-merger integration test where it was trying to connect to DGraph on local host.
6. Define CMD in the Dockerfile and use an environment variable to specify the integration test features only once.
7. Make the target directory copying more robust to destination directories existing or not. 

Running the integration tests locally and in CI.

### DIFF
--- a/test/docker-compose.integration-tests.yml
+++ b/test/docker-compose.integration-tests.yml
@@ -12,7 +12,7 @@ services:
       target: build-test
       args:
         - RUST_BUILD=test-integration
-    command: cargo test --features sqs-executor/integration,node-identifier/integration,kafka-metrics-exporter/integration --test '*'
+    command: cargo test --features sqs-executor/integration,node-identifier/integration,kafka-metrics-exporter/integration,graph-merger/integration --test '*'
     environment:
       - AWS_REGION
       - DEPLOYMENT_NAME
--- a/test/docker-compose.integration-tests.yml
+++ b/test/docker-compose.integration-tests.yml
@@ -3,7 +3,8 @@ version: "3.8"
 # environment variable PWD is assumed to be grapl root directory
 
 services:
-  rust-node-identifier-integration-tests:
+
+  rust-integration-tests:
     image: grapl/rust-integration-tests:${TAG:-latest}
     build:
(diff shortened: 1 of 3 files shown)


test/docker-compose.integration-tests.yml
```diff
@@ -3,15 +3,16 @@ version: "3.8"
...
+  rust-integration-tests:
     image: grapl/rust-integration-tests:${TAG:-latest}
     build:
       context: ${PWD}/src
       dockerfile: rust/Dockerfile
       target: build-test
       args:
         - RUST_BUILD=test-integration
-    command: cargo test --features node-identifier/integration --test '*'
+    command: cargo test --features sqs-executor/integration,node-identifier/integration,kafka-metrics-exporter/integration --test '*'
```
wimax-grapl, line 15: - Can you just add a comment in <both places that specify `--features`> to update the other one? like, comment here to update the dockerfile + comment in dockerfile to update it here
- Not a must have, but thoughts on baking this into a CMD in the dockerfile instead? separate command made sense when multiple integration tests were using the same image, but that seems no longer the case

wimax-grapl (approved):

test/docker-compose.integration-tests.yml
inickles-grapl: Good idea. I added a diff to use the default command in the Dockerfile to bring that value to one file, and then I move it to a variable so there's only one spot now. 👍

colin-grapl (approved):

PR #957, opened 2021-06-26
Pass missing TAG argument to provisioner build.

inickles-grapl:
### Which issue does this PR correspond to?

https://github.com/grapl-security/issue-tracker/issues/593

### What changes does this PR make to Grapl? Why?

This passes a missing TAG arg to the provisioner Docker build.

### How were these changes tested?

```
$ make clean
...
$ export TAG=asdf
$ make graplctl
...
$ make build-services
...
$ echo $?
0
```

PR #958, opened 2021-06-26
Set TAG env var to non-default value in CI pipeline 

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/593

This sets the `TAG` environment variable in CI to catch potential issues related to the above issue.

We'll test with CI, but need https://github.com/grapl-security/grapl/pull/957 to land before we can test this PR, which is why I'm keeping in draft mode for now.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -484,6 +484,7 @@ services:
         target: /home/grapl/pulumi-outputs
         read_only: false
     environment:
+      TAG:
       PULUMI_CONFIG_PASSPHRASE: local-grapl-passphrase
       # Other environment variables like MG_ALPHAS are passed in via
       # Pulumi.local-grapl.yaml
--- a/.buildkite/pipeline.verify.yml
+++ b/.buildkite/pipeline.verify.yml
@@ -1,6 +1,7 @@
 ---
 env:
   PANTS_CONFIG_FILES: "['pants.toml', 'pants.ci.toml']"
+  TAG: ci-pipeline
 
 # TODO: Cache for JS, Rust deps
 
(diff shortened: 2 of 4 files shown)


inickles-grapl:
As an alternative to renaming GRAPL_LAMBDA_TAG, we could just set it here (`GRAPL_LAMBDA_TAG: ${TAG:-latest}`):
https://github.com/grapl-security/grapl/pull/958/commits/a6311997ec75537fe2f1e0d6829b76cf808c9636#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R487
instead of passing up TAG, but it seems leaving two different names could leave to more confusion at some other point.

docker-compose.yml
```diff
@@ -484,6 +484,7 @@ services:
         target: /home/grapl/pulumi-outputs
         read_only: false
     environment:
+      TAG:
```
jgrillo-grapl, line 487: Should this have a value? What does it mean to just specify the key in this mapping?

jgrillo-grapl (approved):
👍 one question, lgtm

docker-compose.yml
inickles-grapl: It just passes whatever the value of that env var, in the docker-compose execution env, to the container env. We do the same this in other places but with the alternate yaml format: https://github.com/grapl-security/grapl/blob/main/docker-compose.yml#L383. I didn't wanna change from the format used already by this service. I kind of like it better anyway.

https://docs.docker.com/compose/compose-file/compose-file-v3/#environment

christophermaier (approved):

docker-compose.yml
jgrillo-grapl: 👍

PR #959, opened 2021-06-28
Rename Buildkite pipeline file for the verify pipeline

christophermaier:
This adopts the convention from the pipeline-infrastructure
repository, in preparation of extending the overall deployment
pipeline for this repository.


inickles-grapl:
I don't understand why the Hello World steps are needed, but looks like that's intentional. At least for now?

inickles-grapl (approved):

christophermaier:
I've set up the next pipeline in the sequence, and that'll be triggered when this merges... the "hello world" bit is just to have a stub for that pipeline so we don't have spurious pipeline failures.

PR #965, opened 2021-06-29
Create lambda zip files in dist directory without "tags"

christophermaier:
The idea of a "tag" doesn't make much sense when thinking about lambda
zip files, particularly in the near future when we'll have an artifact
repository for them. As such, we can remove this logic from our zip
file creation.

Additionally, zip files will not be created in
`src/aws-provision/zips` anymore, since they are artifacts and not
source code. Instead, they will be created in `dist`. Though this is
the default directory in which Pants creates its packages, we will
also use this for our Rust and Typescript-based lambdas.

Now, all lambda zip files will simply be named like
`<LAMBDA_NAME>-lambda.zip`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,7 +4,6 @@ build-support/grapl-venv
 .pants.workdir.file_lock*
 .pids
 **/__pycache__
-src/aws-provision/zips
 src/rust/target/**
 .yarn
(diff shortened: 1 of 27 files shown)


jgrillo-grapl (approved):
awesome!

PR #968, opened 2021-06-30
Grapl Web Authn and Authz Framework

d0nut-grapl:

https://github.com/grapl-security/issue-tracker/issues/534

Rewrites auth to use Argon2ID and, temporarily, avoid doing any client-side hashing (to get back to the basics and redo our authn/authz system).

Most importantly, adds an Authn/Authz framework which introduces authentication and authorization primitives to make gating access to sensitive information or operations trivial and secure.

When we get the actix code up and running we can easily integrate this framework in to enforce authorization requirements.


src/rust/grapl-web-sec/src/authn/dynamodb.rs
```diff
@@ -0,0 +1,237 @@
...
+        table_name: grapl_config::user_session_table_name(),
+        ..Default::default()
+    };
+
+    let session_row_hashmap = dynamodb_client
+        .get_item(session_query)
+        .await?
+        .item
+        .ok_or(WebSecError::MissingSession)
+        .tap_err(|_| warn!("No entries for session found in session table."))?;
```
inickles-grapl, line 59: 1. It's been asked that anything we send to a tracing log macro go into a field, so if anything this should be something like `warn!(message="...")`, however:
2. In this particular case the error text should be attached to the error itself with the `#[error("...")]` enum variant macro from `thiserror`, which implements `Display` for it. So at this point, this should look more like:
```
tap_err(|error| warn!(message=%error))?;
```

Maybe use "error" for field name instead of "message", idk.

src/rust/grapl-web-sec/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
```
inickles-grapl, line 1: Is this to live as a separate crate from the Actix web application that will be the frontend, or is this just a placeholder for now?

src/rust/grapl-web-sec/src/authn/user.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    pub fn get_role(&self) -> &GraplRole {
+        &self.role
+    }
+}
+
+impl AuthenticatedUser {
+    #[allow(dead_code)]
+    #[cfg(test)]
+    /// Used only for constructing tests; should NEVER be used in production code
```
inickles-grapl, line 39: It seems like this code should be reachable from production if that's the case, but also I'm a little confused as to this in the first place: it looks more like a `new` function and I don't see why it should be test specific?

src/rust/grapl-web-sec/src/authz/capability.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl<C: CapabilityT> Deref for Capability<C> {
+    type Target = C;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl<C: CapabilityT<Resource = ()>> FromRequest for Capability<C> {
```
inickles-grapl, line 39: What would an example Resource look like?

inickles-grapl (changes requested):
I have some requests and a couple questions.

src/rust/grapl-web-sec/Cargo.toml
d0nut-grapl: separate crate

src/rust/grapl-web-sec/src/error.rs
```diff
@@ -0,0 +1,41 @@
...
+    AuthenticationFailure,
+    ArgonError(argon2::Error),
+    PasswordHashingError(argon2::password_hash::Error),
+    SessionCreationError,
+    Unauthorized,
+}
+
+impl ResponseError for WebSecError {
+    fn status_code(&self) -> StatusCode {
+        StatusCode::UNAUTHORIZED
```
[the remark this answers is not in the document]
d0nut-grapl: ahh good point. I implemented ResponseError pretty early on before i had some other types of errors in place. I'll update it to show some 500s in cases of dynamodb error, etc.

PR #969, opened 2021-06-30
Makefile: Remove `build` target and fix dependency issue

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/596

1. This makes the `lambdas` target a dependency of `build-services`, helping ensure all targets that depend on `build-services` will build in the correct order.
2. This removes the `build` target alias for `build-services`. Note: there is still no help text for the `build-services` target in `make help`. This is intention as we want to discourage users from using it directly, at least unless we exercise it directly in CI.

I verified  `make build-services` works locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,17 +146,17 @@ build-test-typecheck:
 	docker buildx bake --file ./test/docker-compose.typecheck-tests.yml
 
 .PHONY: build-test-integration
-build-test-integration: lambdas build-services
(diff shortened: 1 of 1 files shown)


d0nut-grapl (approved):
lgtm

jgrillo-grapl (approved):
LGTM

christophermaier (approved):

PR #975, opened 2021-07-02
Swap the base image used for Rust service containers.

inickles-grapl:

NA

The Rust service containers were previously built from the `debian:buster-slim` image from Docker Hub. This changes that to use `gcr.io/distroless/cc`[1] instead, which is a much smaller image that contains a minimal environment for running our Rust services. The distroless image is significantly smaller than the buster-slim image:

```
debian                                      buster-slim          7c5871f12659   9 days ago           69.2MB
...
gcr.io/distroless/cc                        latest    2d782957ae08   51 years ago         21.2MB
```

Note: this is pulling from Google's registry, not Docker Hub.

[1] https://github.com/GoogleContainerTools/distroless/blob/main/cc/README.md

I tested locally by running the E2E tests.

### DIFF
--- a/src/rust/Dockerfile
+++ b/src/rust/Dockerfile
@@ -110,14 +110,12 @@ RUN mkdir -p /grapl/zips; \
 
 # images for running services
 ################################################################################
-FROM debian:buster-slim AS rust-dist
+# More information about the base image used here can be found at: 
+# https://github.com/GoogleContainerTools/distroless/blob/main/cc/README.md.
+# For debugging see: https://github.com/GoogleContainerTools/distroless#debug-images
+FROM gcr.io/distroless/cc AS rust-dist
 
-RUN --mount=type=cache,target=/var/lib/apt/lists \
-    apt-get update && apt-get install -y --no-install-recommends \
-        ca-certificates \
-        netbase
-
-USER nobody
+USER nonroot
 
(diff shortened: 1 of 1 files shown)


src/rust/Dockerfile
```diff
@@ -110,14 +110,12 @@ RUN mkdir -p /grapl/zips; \
 
 # images for running services
 ################################################################################
-FROM debian:buster-slim AS rust-dist
+# More information about the base image used here can be found at: 
+# https://github.com/GoogleContainerTools/distroless/blob/main/cc/README.md.
+# For debugging see: https://github.com/GoogleContainerTools/distroless#debug-images
+FROM gcr.io/distroless/cc AS rust-dist
```
christophermaier, line 116: I assume you tested, but this is indeed the minimal container for our current uses, correct?

christophermaier (approved):
🎉

src/rust/Dockerfile
inickles-grapl: Yeah. We could probably get away with `static` if we statically linked to musl libc, and I'd tried with the `base` image, but we need libgcc, which is only found on the `cc` image. The description for this image even says it's suitable for Rust, so I'm confident this is the one. https://github.com/GoogleContainerTools/distroless/blob/main/cc/README.md

PR #979, opened 2021-07-06
Switch integration and e2e tests back to "beefy" queue

christophermaier:
These shouldn't actually run on our "docker" queue as it exists right
now, because more than one agent can run on those machines. These
tests don't seem to be able to tolerate having multiple instances
running at once on a single machine.

For now, we'll just move back to the "beefy" queue, which is one agent
per machine, pending further investigation.


jgrillo-grapl (approved):
🥩

PR #982, opened 2021-07-06
Selectively build artifacts only if we need to

christophermaier:
Rather than creating new versions of every single artifact every time
we merge a PR, we will only build if something has changed that
requires a new artifact.

For our Python artifacts, we can rely on Pants' fine-grained
dependency awareness to only build the artifacts whose inputs have
changed since our last successful run through the merge pipeline.

For Rust- and Javascript-based artifacts, however, we have to use a
different approach (since Pants does not know about these languages
yet). Here, we'll use the `monorepo-diff` Buildkite plugin, along with
our tagging-convention-aware diff script. It is not a perfect
solution, since we currently either build all Rust containers or no
Rust containers, but it's a step forward.

(Absent code-aware "what changed" logic, we would have to maintain
detailed manual lists of all the inter-file dependencies across our
Rust workspace, which is tedious and error prone.)

### DIFF
--- a/.buildkite/pipeline.merge.yml
+++ b/.buildkite/pipeline.merge.yml
@@ -5,68 +5,42 @@ env:
 steps:
 
   - label: ":aws-lambda::rust: Build Rust Lambdas"
-    command:
-      - "make lambdas-rust"
-    key: "rust-lambdas"
-    artifact_paths:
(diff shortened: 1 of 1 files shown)


jgrillo-grapl (approved):
👍

PR #987, opened 2021-07-09
I screwed up some shell stuff in the AMI merge step

wimax-grapl:
fixes the following at the merge ami stage:
https://buildkite.com/grapl/grapl-merge/builds/54#b1d4a25f-f2f6-4d18-8a30-ca103b009e03
```
.buildkite/scripts/lib/packer_constants.sh: line 26: readonly: `=': not a valid identifier
.buildkite/scripts/lib/packer_constants.sh: line 26: readonly: `.packer-manifest.json': not a valid identifier
```

### DIFF
--- a/.buildkite/pipeline.merge.yml
+++ b/.buildkite/pipeline.merge.yml
@@ -87,6 +87,7 @@ steps:
                 - "packer/"
                 - ".buildkite/scripts/build_packer_ci.sh"
                 - ".buildkite/scripts/lib/packer.sh"
+                - ".buildkite/scripts/lib/packer_constants.sh"
               config:
                 label: ":pipeline: Upload AMI Build"
                 command: "buildkite-agent pipeline upload .buildkite/pipeline.merge.ami-build.yml"
--- a/.buildkite/pipeline.verify.yml
+++ b/.buildkite/pipeline.verify.yml
@@ -143,8 +143,11 @@ steps:
           diff: .buildkite/shared/scripts/diff.sh
           watch:
             - path:
-                - "packer/"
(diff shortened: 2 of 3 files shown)


wimax-grapl:
caused by https://github.com/koalaman/shellcheck/issues/2279

christophermaier:
Weird... when I run `shellcheck` locally, I get this:

```
shellcheck .buildkite/scripts/lib/packer_constants.sh 

In .buildkite/scripts/lib/packer_constants.sh line 26:
readonly PACKER_MANIFEST_SUFFIX = ".packer-manifest.json"
                                ^-- SC1068: Don't put spaces around the = in assignments.
```

That's with 0.5.0 on my machine. This failure doesn't show up running through Pants, however, which uses 0.7.2. Installing 0.7.2 locally and running the above does not show this [SC1068](https://github.com/koalaman/shellcheck/wiki/SC1068) violation, either.

Seems like something important is different between 0.5.0 and 0.7.2 🤔

.buildkite/pipeline.merge.yml
```diff
@@ -87,6 +87,7 @@ steps:
                 - "packer/"
                 - ".buildkite/scripts/build_packer_ci.sh"
                 - ".buildkite/scripts/lib/packer.sh"
+                - ".buildkite/scripts/lib/packer_constants.sh"
```
christophermaier, line 90: Maybe `.buildkite/scripts/lib/packer*.sh`?

>A path can also be a glob pattern. For example specify path: "**/*.md" to match all markdown files.
From https://github.com/chronotc/monorepo-diff-buildkite-plugin#path

christophermaier (approved):

PR #989, opened 2021-07-09
Use some globs in the packer-build pipelines. Unrelated: Also, add some ymls to the formatter.

wimax-grapl:
using globs was requested by cm previously in https://github.com/grapl-security/grapl/pull/987
but that PR has since been merged.

unrelated (and should probably be a different PR, but I am lazy):
also, realized hidden files are not automatically picked up by the prettier **/* glob, so manually specified buildkite and github ymls.

### DIFF
--- a/.buildkite/scripts/test_packer.sh
+++ b/.buildkite/scripts/test_packer.sh
@@ -8,4 +8,8 @@ echo -e "--- :packer: Performing test build of AMI"
 
 source .buildkite/scripts/lib/packer.sh
 
+# We don't actually use anything in `constants` here, but make sure we can
+# source it without errors.
+source .buildkite/scripts/lib/packer_constants.sh
+
 PACKER_VARS="-var build_ami=false" build_ami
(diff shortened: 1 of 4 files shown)


.buildkite/scripts/test_packer.sh
```diff
@@ -8,4 +8,8 @@ echo -e "--- :packer: Performing test build of AMI"
 
 source .buildkite/scripts/lib/packer.sh
 
+# We don't actually use anything in `constants` here, but make sure we can
+# source it without errors.
+source .buildkite/scripts/lib/packer_constants.sh
```
christophermaier, line 13: I don't understand the rationale for this.

wimax-grapl:
Rationale:
1. That's where my read-only failure occurred during the merge stage
2. It's be nice to get that in CI  before merge

wimax-grapl:
nvm, going to just do the `.buildkite` yml stuff in another PR

PR #990, opened 2021-07-09
add dist/codegen to pyrightconfig.json

jgrillo-grapl:
NA

update `build-support/editor_setup.py` to emit a pyright config which ideally should make pyright aware of the generated protobuf stubs emitted by pants with `./pants export-codegen ::`

checked that `dist/codegen` is in the `"extraPaths"` for the `"src/python"` execution environment


christophermaier (approved):

wimax-grapl:
Not a must-have, but if you could throw a print into the `editor_setup.py` about "hey, don't forget to ./pants whatever", that'd be cool :)

PR #991, opened 2021-07-09
Fix AMI merge again; parallelize client and server builds; and add some half decent testing

wimax-grapl:
AMI issue was reported on slack
https://grapl-internal.slack.com/archives/C024C7TMJ2J/p1625857360079300

parallelizing is this issue:
https://github.com/grapl-security/grapl/pull/991

- The upload-manifest stuff used to refer to a var PACKER_MANIFESTS; that no longer exists.
- So i fixed that.
- and I also made client and server build in parallel. 

added some unit and integration tests


.buildkite/scripts/build_packer_ci.sh
```diff
@@ -9,24 +9,22 @@
 set -euo pipefail
 
 source .buildkite/scripts/lib/packer.sh
-source .buildkite/scripts/lib/packer_constants.sh
 
-echo -e "--- :packer: Performing build of AMI"
 export GIT_SHA="${BUILDKITE_COMMIT}"
 export GIT_BRANCH="${BUILDKITE_BRANCH}"
+# This syntax marks these two as required
 : "${BUILDKITE_BUILD_NUMBER}"
```
christophermaier, line 16: The `:` is the "command that does nothing" ([reference](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Bourne-Shell-Builtins))

Using it in this construction (and with our `set -u` configuration), will fail the script if that variable is not set, since `:` expands the variable for us.

This is a convenient way to front-load the environment variable checking, particularly since the variable isn't actually used anywhere else in this script. (In fact, it's optional inside the Packer template, too, so unless we did something like this, we could just silently be missing it and not know until we went to go search for our AMIs, expecting them to be tagged with a build number.)

.buildkite/scripts/lib/packer_unit_test.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+# This test runs with `make test-unit-python` (poorly named).
```
christophermaier, line 2: We should probably break that up.

.buildkite/scripts/lib/packer_unit_test.sh
```diff
@@ -0,0 +1,55 @@
...
+    exitcode=$?
+
+    assertEquals 0 $exitcode
+    # `upload_manifest` conveniently `cat`s our file
+    assertContains "im-a-manifest" "${stdout}"
+}
+
+test_upload_manifest_file_doesnt_exist() {
+    local -r image_name="fakeimage2"
+    # no touch
```
christophermaier, line 47: ???

.buildkite/scripts/lib/packer.sh
```diff
@@ -1,15 +1,41 @@
...
+        # shellcheck disable=SC2086
+        packer build ${PACKER_VARS:-} -var is_server=true packer/nomad-server-client
+    elif [ "${packer_image_name}" == "${PACKER_IMAGE_NAME_CLIENT}" ]; then
+        # shellcheck disable=SC2086
+        packer build ${PACKER_VARS:-} -var is_server=false packer/nomad-server-client
+    else
+        echo "Unknown PACKER_IMAGE_NAME ${packer_image_name}"
+        exit 42
+    fi
+}
```
christophermaier, line 25: Since there are only two choices (it's either a server or its not), it might be simpler to pull this naming logic into the Packer template itself, rather than having to contend with making sure we've used the right concrete names everywhere else.

.buildkite/scripts/lib/packer_unit_test.sh
wimax-grapl: removed, useless test, agreed

.buildkite/scripts/lib/packer.sh
wimax-grapl: 1. On one hand, I agree
2. On the other hand, I don't want to code us into a corner where we only support "true" or "false"; we could potentially have a situation in the future where we have >2 AMIs. 

Ultimately, I really wanted some HCL way to either:
- define a string enum of <only 2 valid values right now> 
- or throw an error if a string does not equal "value1" or "value2"

buuuut i couldn't find that

wimax-grapl: found it


  validation {
    condition     = length(var.image_id) > 4 && substr(var.image_id, 0, 4) == "ami-"
    error_message = "The image_id value must be a valid AMI id, starting with \"ami-\"."
  }

.buildkite/scripts/lib/packer_unit_test.sh
wimax-grapl: agreed. have it in another branch

.buildkite/scripts/lib/BUILD
```diff
@@ -1,5 +1,5 @@
 shell_library()
 
 shunit2_tests(
-    name="tests"
+    name="tests",
 )
```
wimax-grapl, line 5: okay, yes, this is not a required change, but i'd prefer to not run CI again if i don't need to lol

.buildkite/pipeline.verify.ami-test.yml
```diff
@@ -11,20 +11,42 @@ env:
...
+      - docker#v3.8.0:
+          image: "hashicorp/packer:1.7.2"
+          entrypoint: bash
+          environment:
+            - "BUILDKITE_BRANCH"
+            - "BUILDKITE_BUILD_NUMBER"
+            - "BUILDKITE_COMMIT"
+            - "BUILDKITE_PIPELINE_NAME"
+            - "BUILDKITE_STEP_KEY"
+            - "PACKER_IMAGE_NAME=grapl-nomad-consul-client"
```
christophermaier, line 49: You can just use "PACKER_IMAGE_NAME" here; you've already defined the value above in the job's own `env` map.

(Likewise for the other jobs.)

christophermaier (approved):

PR #996, opened 2021-07-12
Data tenancy

colin-grapl:

612

1. Consolidates generated protobuf into one crate. This is apparently the typical way to do things, and allows us to share implementations on protobuf types.
2. Creates a new Envelope message, and a Meta message
3. Weaves the Envelope through Grapl. This is primarily handled in the sqs-executor.
4. I've also gotten the Envelope into the analyzer-executor, though we don't use it to communicate to the engagement-creator. I think that the engagement-creator really needs to be re-thought anyways, and should be more of an RPC, so I'd rather handle it separately.

The 'Meta' contains a 'client_id', 'tenant_id', and 'trace_id'.

So far I've just run e2e locally. I intend to run AWS e2e as well before I take this PR out of draft.


src/python/analyzer_executor/src/analyzer_executor_lib/analyzer_executor.py
```diff
@@ -389,9 +393,13 @@ def chunker(seq: List[BaseView], size: int) -> List[List[BaseView]]:
...
 
 
-def emit_event(s3: S3ServiceResource, event: ExecutionHit) -> None:
+def emit_event(s3: S3ServiceResource, event: ExecutionHit, metadata: Meta) -> None:
     LOGGER.info(f"emitting event for: {event.analyzer_name, event.nodes}")
 
+    meta_dict = {
+        "trace_id": str(proto_uuid_to_pyuuid(metadata.trace_id)),
+    }
+
```
colin-grapl, line 402: The Python services don't communicate via protobuf right now - I'd rather handle that in another PR, and ideally post-nomad so that we can move a lot of things to RPC.

src/proto/graplinc/grapl/api/services/v1beta1/types.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+package graplinc.grapl.api.services.v1beta1;
+
+// A protobuf, encoded by splitting the 128bit representation into two 64bit values
+message ProtoUuidV4 {
```
jgrillo-grapl, line 6: Maybe just `Uuid`? `Proto` is redundant because it's a protobuf and `V4` is redundant because it's already included in the serialized data.

src/proto/graplinc/grapl/api/services/v1beta1/types.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+package graplinc.grapl.api.services.v1beta1;
```
jgrillo-grapl, line 3: Could we repackage this as `graplinc.grapl.pipeline.v1beta1` instead?

src/proto/graplinc/grapl/api/services/v1beta1/types.proto
```diff
@@ -0,0 +1,27 @@
...
+// A protobuf, encoded by splitting the 128bit representation into two 64bit values
+message ProtoUuidV4 {
+  // Least significant bytes
+  fixed64 lsb = 1;
+  // Most significant bytes
+  fixed64 msb = 2;
+}
+
+// Metadata about a specific message, or its context
+message Meta {
```
jgrillo-grapl, line 14: Can we call it `Metadata`?

src/proto/graplinc/grapl/api/services/v1beta1/types.proto
```diff
@@ -0,0 +1,27 @@
...
+  // The name of the service/ client that produced this message
+  string client_id = 1;
+  // Customer identity
+  ProtoUuidV4 tenant_id = 2;
+  // Unique identity representing a unified request throughout Grapl
+  ProtoUuidV4 trace_id = 3;
+}
+
+message Envelope {
+  Meta metadata = 1;
```
jgrillo-grapl, line 26: docstrings would be nice here

src/python/grapl-common/grapl_common/envelope.py
```diff
@@ -0,0 +1,71 @@
...
+        return Envelope(envelope)
+
+    @staticmethod
+    def from_proto(s: bytes) -> Envelope:
+        envelope = _Envelope()
+        envelope.ParseFromString(s)
+
+        return Envelope(envelope)
+
+    def SerializeToString(self) -> str:
```
jgrillo-grapl, line 58: this should be snake_case

src/python/grapl-common/grapl_common/envelope.py
```diff
@@ -0,0 +1,71 @@
...
+    metadata = Meta()
+    metadata.client_id = client_id
+    metadata.trace_id.CopyFrom(pyuuid_to_proto_uuid(uuid4()))
+    metadata.tenant_id.CopyFrom(pyuuid_to_proto_uuid(uuid4()))
+    return metadata
+
+
+class Envelope(object):
+    def __init__(self, envelope: _Envelope) -> None:
+        self.envelope = envelope
```
jgrillo-grapl, line 41: This should be `self._envelope` because it's not exported in the `Envelope` class' public API

src/python/grapl-common/grapl_common/envelope.py
```diff
@@ -0,0 +1,71 @@
+from __future__ import annotations
+
+import uuid
+from uuid import uuid4
+
+from graplinc.grapl.api.services.v1beta1.types_pb2 import Envelope as _Envelope
+from graplinc.grapl.api.services.v1beta1.types_pb2 import Meta, ProtoUuidV4
```
jgrillo-grapl, line 7: I like what you did above renaming the generated protobuf class to `_Envelope`, would also be nice to do it for `_Meta` and `_ProtoUuidV4`

src/python/grapl-common/grapl_common/envelope.py
```diff
@@ -0,0 +1,71 @@
...
+        envelope = _Envelope()
+        envelope.ParseFromString(s)
+
+        return Envelope(envelope)
+
+    def SerializeToString(self) -> str:
+        return self.envelope.SerializeToString()
+
+    @property
+    def metadata(self) -> Meta:
```
jgrillo-grapl, line 62: It would be nice if this returned a type we control as opposed to one owned by the generated protobuf stubs

src/python/grapl-common/grapl_common/envelope.py
```diff
@@ -0,0 +1,71 @@
...
+    lsb = int.from_bytes(lsb_bytes, byteorder="little", signed=False)
+    msb = int.from_bytes(msb_bytes, byteorder="little", signed=False)
+
+    proto_uuid = ProtoUuidV4()
+    proto_uuid.lsb = lsb
+    proto_uuid.msb = msb
+    return proto_uuid
+
+
+def init_metadata(client_id: str) -> Meta:
```
jgrillo-grapl, line 31: If we had a type to represent the `Meta` class, this could just be that class' constructor.

src/python/grapl-common/grapl_common/envelope.py
```diff
@@ -0,0 +1,71 @@
+from __future__ import annotations
+
+import uuid
+from uuid import uuid4
+
+from graplinc.grapl.api.services.v1beta1.types_pb2 import Envelope as _Envelope
+from graplinc.grapl.api.services.v1beta1.types_pb2 import Meta, ProtoUuidV4
+
+
+def proto_uuid_to_pyuuid(proto_uuid: ProtoUuidV4) -> uuid.UUID:
```
jgrillo-grapl, line 10: These methods feel like they want to be `Uuid::from(uuid: ProtoUuidV4) -> Uuid` and `Uuid::into() -> uuid.UUID`, which would be nice.

src/rust/Cargo.toml
```diff
@@ -19,6 +19,7 @@ members = [
     "./kafka-metrics-exporter",
     "./metric-forwarder",
     "./node-identifier",
+    "./rust-proto",
```
jgrillo-grapl, line 22: why twice?

src/rust/rust-proto/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "grapl-graph-descriptions"
+name = "rust-proto"
```
jgrillo-grapl, line 2: the word `rust` feels redundant here

src/rust/rust-proto/src/lib.rs
```diff
@@ -0,0 +1,28 @@
+pub(crate) mod graplinc {
+    pub(crate) mod grapl {
+        pub(crate) mod api {
+            pub(crate) mod graph {
+                pub(crate) mod v1beta1 {
+                    include!(concat!(
+                        env!("OUT_DIR"),
+                        "/graplinc.grapl.api.graph.v1beta1.rs"
```
jgrillo-grapl, line 8: This obviously isn't where this term is defined, but since the corresponding `.proto` file wasn't changed in this PR I'll comment here. I'm having a little trouble understanding what `api` means in this package name. These data don't seem to be RPC APIs, instead they seem like pipeline data. It would be a good time to flesh out a package structure that makes sense.

src/rust/sqs-executor/src/lib.rs
```diff
@@ -306,16 +334,37 @@ async fn process_message<
...
+            if events.is_empty() {
+                tracing::debug!(message = "Serialized events produced no output");
+                return;
+            }
+
+            let mut envelopes = vec![];
+            for event in events {
+                let envelope = rust_proto::services::Envelope {
+                    metadata: Some(meta.clone()),
+                    inner_type: "".to_string(),
```
jgrillo-grapl, line 351: Is this intentional? The `inner_type` value should be the fully-qualified package of the protobuf type which was used to serialize `inner_message`.

jgrillo-grapl (changes requested):

src/proto/graplinc/grapl/api/services/v1beta1/types.proto
```diff
@@ -0,0 +1,31 @@
...
+  ProtoUuidV4 tenant_id = 2;
+  // Unique identity representing a unified request throughout Grapl
+  ProtoUuidV4 trace_id = 3;
+}
+
+// A generic wrapper around other messages
+message Envelope {
+  // Metadata about this message
+  Meta metadata = 1;
+  // The string name of this message's type
```
jgrillo-grapl, line 27: More specifically, this needs to be the fully-qualified name of the protobuf type used to serialize this message. This will become important later because upon deserialization we will need to look up the appropriate deserializer.

jgrillo-grapl (changes requested):

src/proto/graplinc/grapl/api/services/v1beta1/types.proto
colin-grapl: True about V4, mostly the 'Proto' is just to avoid name collisions - like I'm going to end up importing it `as ProtoUuid` everywhere where I use `use uuid::Uuid`. I don't care though, can change if you'd like. Removing V4 seems fine.

src/python/grapl-common/grapl_common/envelope.py
colin-grapl: I'm just matching the existing casing from Google's protobuf libraries, but can change.

src/python/grapl-common/grapl_common/envelope.py
colin-grapl: Yes but you can't do that in Python.

src/rust/sqs-executor/src/lib.rs
colin-grapl: Not intentional, thanks for catching - I should have used a todo!() so that the tests failed.

src/proto/graplinc/grapl/api/services/v1beta1/types.proto
```diff
@@ -0,0 +1,31 @@
...
+  // Least significant bytes
+  fixed64 lsb = 1;
+  // Most significant bytes
+  fixed64 msb = 2;
+}
+
+// Metadata about a specific message, or its context
+message Meta {
+  // The name of the service/ client that produced this message
+  string client_id = 1;
```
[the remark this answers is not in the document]
colin-grapl: No, this was a holdover and can be removed.

src/python/grapl-common/grapl_common/envelope.py
jgrillo-grapl: Sorry, to be clear I'm talking about writing python types like this:

```python
class Uuid:

    def __init__(self, uuid: uuid.UUID) -> Uuid:
        self._uuid = uuid

    @staticmethod
    def from(uuid: ProtoUuidV4) -> Uuid:
        ...

    def into() -> uuid.UUID:
        ...

    def into_proto() -> ProtoUuidV4:
        ...
```
and
```python
class Metadata:

    def __init__(self, tenant_id: Uuid, trace_id: Uuid):
        self.tenant_id = tenant_id
        self.trace_id = trace_id

    @staticmethod
    def from(metadata: Meta) -> Metadata:
        ...

   def into_proto() -> Meta:
        ...
```

src/rust/rust-proto/Cargo.toml
colin-grapl: Yeah, I'm not sure. `grapl-proto` doesn't feel much better. Really this is just the generated rust code for our protobuf, so I went with `rust-proto`.

src/python/grapl-common/grapl_common/envelope.py
colin-grapl: Oh for sure. Yeah, it just gets really awkward really fast. You either have to upfront convert everything, and then serialize back, or you need to have conversion code everywhere.

src/proto/graplinc/grapl/api/services/v1beta1/types.proto
jgrillo-grapl: It should be fine to just refer to it as `rust_proto::Uuid`, right? In any case, I think the type _here_ should just be `Uuid` to avoid redundancy, and consumers can rename it however they like.

jgrillo-grapl:
Closing this to remove clutter, will re-open when it's ready for review.

PR #1007, opened 2021-07-14
Flatten the AMI IDs map, so that create_rc can use it to do pulumi config set stuff

wimax-grapl:
```jq stuff - basically turn
{ 
"some-ami-name": {
  "us-east-1": "ami-1234"
}
}
into
{ "some-ami-name.us-east-1": "ami-1234" }
```

see https://grapl-internal.slack.com/archives/C017DKYF55H/p1626220139413700 for more context

i'm tackling this by changing the input, not the receiving side

### DIFF
--- a/.buildkite/scripts/record_artifacts.sh
+++ b/.buildkite/scripts/record_artifacts.sh
@@ -10,11 +10,11 @@ upload_artifacts_file() {
     # a corresponding packer manifest names "image_name.packer-manifest.json".
 
     # e.g. "grapl-service"
-    local -r root_name=$1
+    local -r packer_image_name=$1
     # e.g. "grapl-service.packer-manifest.json"
-    local -r manifest_file="${root_name}${PACKER_MANIFEST_SUFFIX}"
+    local -r manifest_file="${packer_image_name}${PACKER_MANIFEST_SUFFIX}"
     # e.g. "grapl-service.artifacts.json"
-    local -r artifacts_file="${root_name}${ARTIFACTS_FILE_SUFFIX}"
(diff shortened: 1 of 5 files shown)


.buildkite/scripts/record_artifacts.sh
```diff
@@ -28,14 +28,13 @@ upload_artifacts_file() {
...
-    # { "us-east-1": "ami-111", "us-east-2": "ami-222", ...}
-    local -r ami_ids_dict=$(jq --raw-output --from-file "${jq_filter_path}" "${manifest_file}")
+    # { "imagename.us-east-1": "ami-111", "imagename.us-east-2": "ami-222", ...}
+    local -r ami_ids_dict=$(jq --raw-output --arg IMAGE_NAME ${packer_image_name} --from-file "${jq_filter_path}" "${manifest_file}")
     echo "${ami_ids_dict}"
 
     # Creating artifacts file
     echo -c "--- :gear: Creating ${artifacts_file} file"
-    echo "{\"${root_name}-amis\": ${ami_ids_dict}}" > "${artifacts_file}"
-    jq '.' "${artifacts_file}"
```
christophermaier, line 38: FWIW, this `jq` call was to output the contents in the Buildkite log for visibility.

christophermaier (approved):

.buildkite/scripts/record_artifacts.sh
wimax-grapl: yep! the above echo on 33 does it too tho

wimax-grapl:
https://github.com/grapl-security/buildkite-common/pull/3
is where i tried to solve this on the receiving side (instead of the sending side)

PR #1011, opened 2021-07-15
Copy artifacts from origin/rc to your local pulumi config

wimax-grapl:

### DIFF
--- a/pulumi/bin/copy_artifacts_from_rc.sh
+++ b/pulumi/bin/copy_artifacts_from_rc.sh
@@ -2,17 +2,21 @@
 # This will grab a subset of keys from `origin/rc`'s Pulumi.testing.yaml.
 set -euo pipefail
 
-readonly THIS_DIR=$(dirname "${BASH_SOURCE[0]}")
+THIS_DIR=$(dirname "${BASH_SOURCE[0]}")
+readonly THIS_DIR
 
 ensureRightDir() {
-    cd $THIS_DIR/../grapl
+    cd "$THIS_DIR/../grapl"
 }
 ensureRightDir
 
 # Read from $1 or default to what `pulumi stack` says
-readonly CURRENT_STACK="${1:-$(pulumi stack --show-name)}"
-readonly GRAPL_ROOT="$(git rev-parse --show-toplevel)"
-readonly RC_CONFIG_FILE="/tmp/rc_pulumi_testing.yaml"
+CURRENT_STACK="${1:-$(pulumi stack --show-name)}"
+GRAPL_ROOT="$(git rev-parse --show-toplevel)"
+RC_CONFIG_FILE="/tmp/rc_pulumi_testing.yaml"
+readonly CURRENT_STACK
+readonly GRAPL_ROOT
+readonly RC_CONFIG_FILE
 
 confirmModify() {
     read -r -p "This will modify your ${CURRENT_STACK} config. Continue (y/n)?" choice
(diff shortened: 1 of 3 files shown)


.buildkite/shared/lib/json_tools.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
wimax-grapl, line 1: (I'm committing that other PR before this one, so upon commit this will all just disappear)

pulumi/bin/subset_of_artifacts.jq
```diff
@@ -0,0 +1,8 @@
+. 
+| with_entries(
+    select(
+        # Select keys that match this regex.
+        # (note: match also supports N things to match on, makes adding to this subset easy)
+        .key|match("grapl-nomad-consul-*")
```
christophermaier, line 6: I'm not quite sure of the utility of this; why not just pull the whole `artifacts` object over?

pulumi/bin/copy_artifacts_from_rc.sh
```diff
@@ -0,0 +1,69 @@
...
+    add_artifacts "${CURRENT_STACK}" "${artifacts_subset}"
+
+    # ensure it worked
+    local -r get_result=$(pulumi config get --path artifacts.grapl-nomad-consul-client.us-east-1)
+    if [[ $get_result =~ ami-.* ]]; then
+        exit 0
+    else
+        echo "Unexpected result for $(pulumi config get): ${get_result}"
+        exit 42
+    fi
```
christophermaier, line 64: I think you could just show the updated stack and end, rather than focusing on a single value.

pulumi/bin/copy_artifacts_from_rc.sh
```diff
@@ -0,0 +1,68 @@
...
+            pulumi config set \
+                --path "artifacts.${key}" \
+                "${value}"
+        done
+}
+
+main() {
+    confirmModify
+
+    echo "--- Grab artifacts from origin/rc config file"
```
wimax-grapl, line 45: (yes, this doesn't run in buildkite, i just like that section marker)

pulumi/bin/copy_artifacts_from_rc.sh
```diff
@@ -0,0 +1,69 @@
...
+readonly THIS_DIR=$(dirname "${BASH_SOURCE[0]}")
+
+ensureRightDir() {
+    cd $THIS_DIR/../grapl
+}
+ensureRightDir
+
+readonly GRAPL_ROOT="$(git rev-parse --show-toplevel)"
+readonly RC_CONFIG_FILE="/tmp/rc_pulumi_testing.yaml"
+readonly CURRENT_STACK="$(pulumi stack --show-name)"
```
christophermaier, line 14: I think there are cases where there is no currently selected stack; it would probably be better to pass in a name as an argument.

pulumi/bin/copy_artifacts_from_rc.sh
```diff
@@ -0,0 +1,68 @@
...
+                "${value}"
+        done
+}
+
+main() {
+    confirmModify
+
+    echo "--- Grab artifacts from origin/rc config file"
+    git fetch origin
+    git show origin/rc:pulumi/grapl/Pulumi.testing.yaml > ${RC_CONFIG_FILE}
```
christophermaier, line 47: At some point the branch/tag and the stack will need to be parameters, but it's fine for now.

.buildkite/shared/lib/json_tools.sh
christophermaier: I think a rebase would be cleaner 🙏

christophermaier (approved):

PR #1015, opened 2021-07-16
Build UX tarball in the merge pipeline

christophermaier:
A new makefile target, `ux-tarball`, has been added, which creates a
`tar.gz` file containing the UX assets. This is then used in our CI
merge pipeline when we generate artifacts, pushing it to Cloudsmith
and adding the version to our Pulumi stack(s) as we do with our other
artifacts.

The Pulumi code has been modified slightly to take this into
account. One wrinkle is that Pulumi has no native way to expand a
tarball into an S3 bucket, which is how our frontend is currently
deployed. To get around this, a helper script
`pulumi/bin/prepare_grapl_ux_dependency.sh` has been added. When given
the name of a Pulumi stack, it will prepare the UX assets as
appropriate. If a version pin for the UX is present in the stack, that
version will be downloaded from Cloudsmith and expanded into
`dist/grapl-ux-<VERSION>`. If a version pin is not present, we run
`make build-ux` on the source code that is present. In either case,
Pulumi will pull the code from the correct location and deploy.

If we ever change our deployment approach, or Pulumi gets new
abilities to unpack tarballs into S3 buckets, we should *definitely*
revisit this situation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,6 +176,24 @@ graplctl: ## Build graplctl and install it to the project root
 build-ux: ## Build website assets
 	cd src/js/engagement_view && yarn install && yarn build
 
+# Create the dist directory if necessary; don't need to document this
+# formally for `make help`, though.
+dist:
+	mkdir dist
+
+# This is used to create the artifact that will be uploaded to our
+# artifact repository in CI, and will be the artifact that is used by
+# our Pulumi deployments.
+.PHONY: ux-tarball
+ux-tarball: build-ux dist ## Build website asset tarball
+	tar \
+		--create \
+		--gzip \
+		--verbose \
+		--file="dist/grapl-ux.tar.gz" \
+		--directory=src/js/engagement_view/build \
+		.
+
 .PHONY: lambdas
 lambdas: lambdas-rust lambdas-js lambdas-python ## Generate all lambda zip files
 
--- a/pulumi/bin/prepare_grapl_ux_dependency.sh
+++ b/pulumi/bin/prepare_grapl_ux_dependency.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+
+# Since Pulumi does not appear to have a native way to extract a
+# tarball into an S3 bucket, we need to have a bit of external logic
+# to download the artifact for our UX assets to the local machine and
+# extract it into a well-known directory (here,
+# /dist/grapl-ux-<VERSION>) that is accessible to our Pulumi code.
+#
+# If the Pulumi stack being deployed has a pinned version for the
+# grapl-ux artifact, this script should be run prior to `pulumi up`.
+
+set -euo pipefail
+
+# A Pulumi stack name, such as "grapl/testing".
+#
+# This should correspond to a stack from our `pulumi/grapl` project.
+readonly stack="${1}"
+
+# Download a tar.gz artifact from Cloudsmith to `/tmp`, then extract
+# it into the `dist` directory.
+#
+#    retrieve_targz_from_cloudsmith foo 0.0.1
+#    # => Downloads foo.tar.gz, version 0.0.1 from our Cloudsmith
+#    #    "raw" repository
+#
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -176,6 +176,24 @@ graplctl: ## Build graplctl and install it to the project root
 build-ux: ## Build website assets
 	cd src/js/engagement_view && yarn install && yarn build
 
+# Create the dist directory if necessary; don't need to document this
+# formally for `make help`, though.
+dist:
+	mkdir dist
```
wimax-grapl, line 182: I think you need the "oh it's okay if dist already exists" option `--parents` (poorly named) on this

mkdir __pycache__
mkdir: cannot create directory ‘__pycache__’: File exists
echo $?
1

pulumi/bin/prepare_grapl_ux_dependency.sh
```diff
@@ -0,0 +1,87 @@
...
+
+    # Extract the artifact into our dist directory. The directory has
+    # the version appended to it to ensure that we're using the
+    # correct assets in Pulumi.
+    mkdir -p "dist/${base_filename}-${version}"
+    tar --extract \
+        --gunzip \
+        --verbose \
+        --directory="dist/${base_filename}-${version}" \
+        --file="${savename}"
```
wimax-grapl, line 55: thank you for not `-xvjpf`'ing

pulumi/bin/prepare_grapl_ux_dependency.sh
```diff
@@ -0,0 +1,87 @@
...
+    #
+    # However, by running the main logic in a subshell and changing to
+    # the root as our first operation, we can allow the script to be
+    # run from any subdirectory.
+    REPOSITORY_ROOT=$(git rev-parse --show-toplevel)
+    readonly REPOSITORY_ROOT
+    cd "${REPOSITORY_ROOT}" || exit 2
+
+    ux_version=$(
+        cd pulumi/grapl || exit 1
```
wimax-grapl, line 75: don't these cause an `exit` anyway if they fail? or do you want different codes based on where they fail

pulumi/grapl/__main__.py
```diff
@@ -227,13 +227,44 @@ def main() -> None:
...
+        # from.
+        #
+        # We have to do this because there isn't an apparent way in
+        # Pulumi to expand a tarball into an S3 bucket. If that every
+        # becomes possible, we should us it, by all means!
+        #
+        # If there's no configured version, we just pull from the
+        # build directory in the local source tree (it will need to
+        # have been built first, too).
+        grapl_ux_version = configured_version_for("grapl-ux")
```
wimax-grapl, line 242: Can we pull all this ux-specific stuff out to its own function? I think it's large and self-contained enough to make an argument against having it in main()

Makefile
christophermaier: Make takes care of this. `dist` is not a phony target, so Make treats this as a recipe for how to "build" `dist`.

pulumi/bin/prepare_grapl_ux_dependency.sh
christophermaier: I think this was from a prior iteration of this script without `set -e`, which triggered a violation of [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164).

I'll remove them. Thanks for catching that.

pulumi/bin/prepare_grapl_ux_dependency.sh
christophermaier: I will _never_ use short options in scripts if they have corresponding long options. 😂

pulumi/grapl/__main__.py
christophermaier: Yup, that's fair. Will do 👍

wimax-grapl (approved):

wimax-grapl (approved):

PR #1018, opened 2021-07-17
Nomad servers defined in Pulumi

wimax-grapl:

grapl/issue-tracker#600

- Pulumi now requires an artifact be defined for your AMIs.
- Add a NomadServerCluster (and Ec2Cluster) to Pulumi.

`pulumi up` locally. Doesn't work yet due to 
https://github.com/grapl-security/grapl/pull/1017


pulumi/grapl/Pulumi.local-grapl.yaml
```diff
@@ -25,3 +25,13 @@ config:
   kafka:bootstrapServers:
     - "kafka-broker:9092"
   kafka:tlsEnabled: false
+
+  grapl:artifacts:
+    # You probably don't want any service versions in here.
+
+    # You DO need to specify AMIs (they're require_artifact()'d)
+    # Since this is local-grapl, we don't need real AMIs for Localstack EC2
+    grapl-nomad-consul-client:
```
jgrillo-grapl, line 34: can these just be `nomad-consul-client` and `nomad-consul-server` since they're already namespaced to `grapl` on L29?

pulumi/infra/nomad_server_cluster.py
```diff
@@ -0,0 +1,171 @@
+import json
```
jgrillo-grapl, line 1: would `nomad_cluster` work? `server` seems redundant

pulumi/infra/nomad_server_cluster.py
```diff
@@ -0,0 +1,171 @@
...
+from infra.config import DEPLOYMENT_NAME, DGRAPH_LOG_RETENTION_DAYS
+from infra.ec2 import Ec2Port
+from infra.ec2_cluster import Ec2Cluster
+from infra.network import Network
+from infra.policies import EC2_DESCRIBE_INSTANCES_POLICY
+
+import pulumi
+
+
+class NomadServer(pulumi.ComponentResource):
```
jgrillo-grapl, line 16: Isn't this actually a `NomadCluster` because it's spread across 3 machines?

pulumi/grapl/__main__.py
```diff
@@ -211,6 +212,12 @@ def main() -> None:
         ),
     )
 
+    nomad_server_cluster = NomadServer(
```
jgrillo-grapl, line 215: given my comments on `infra.nomad_server_cluster` this would become

```python
nomad_cluster = NomadCluster(
    "nomad-cluster",
    network=network,
    internal_service_ports=(),
)
```

jgrillo-grapl (changes requested):
Some naming nits

pulumi/grapl/Pulumi.local-grapl.yaml
wimax-grapl: These two strings are used all over the pipeline - always as an enum of <one of these two names> - and there are places where this is not namespaced (most notably, the `ami_name=` in `image.pkr.hcl`.)

pulumi/infra/nomad_server_cluster.py
wimax-grapl: can do

pulumi/infra/nomad_server_cluster.py
wimax-grapl: yep, done

pulumi/infra/ec2_cluster.py
```diff
@@ -0,0 +1,109 @@
...
+                ],
+                credit_specification=aws.ec2.InstanceCreditSpecificationArgs(
+                    cpu_credits="unlimited",
+                ),
+                tags=instance_tags,
+                iam_instance_profile=iam_instance_profile.name,
+                # vpc_security_group_ids=vpc_security_group_ids,
+                metadata_options=aws.ec2.InstanceMetadataOptionsArgs(  # Consul relies on EC2 metadata
+                    http_endpoint="enabled",
+                    http_tokens="optional",  # Can we at least use v2? Should verify.
```
christophermaier, line 103: I think so.

pulumi/infra/ec2_cluster.py
```diff
@@ -0,0 +1,109 @@
...
+QuorumSize = Union[Literal[1], Literal[3], Literal[5]]
+
+
+class Ec2Cluster(pulumi.ComponentResource):
+    def __init__(
+        self,
+        name: str,
+        vpc: Network,
+        quorum_size: QuorumSize,
+        quorums: int,  # Number of quorums to actually build
```
christophermaier, line 22: These names are confusing. A quorum refers to the size of the smallest majority in a cluster. With a cluster of 5, the quorum is 3.

I think `quorum_size` is actually `cluster_size`; I guess `quorums` is actually how many clusters you want to build, with one cluster per private network, relying on the fact that our private networks are in different availability zones?

Do these clusters form a "super-cluster"?

jgrillo-grapl (approved):

pulumi/infra/nomad_server_fleet.py
```diff
@@ -0,0 +1,181 @@
...
+            Ec2Port("udp", 8600),
+            Ec2Port("tcp", 8600),
+            *internal_service_ports,
+        ):
+            port.allow_internally(self.security_group)
+
+        # allow hosts in the nomad-server security group to make outbound
+        # connections to the Internet for these services:
+        #   TCP 443 -- AWS SSM Agent (for handshake)
+        #   TCP 80 -- yum package manager and wget
```
christophermaier, line 135: Why does it need to run `yum` or `wget`?

pulumi/grapl/Pulumi.local-grapl.yaml
```diff
@@ -25,3 +25,13 @@ config:
...
+      us-east-1: ami-000fake
```
christophermaier, line 35: What does Localstack actually do with these, then?

pulumi/infra/nomad_server_fleet.py
wimax-grapl: it doesn't, looks like cobrien copy pasted this

pulumi/grapl/Pulumi.local-grapl.yaml
wimax-grapl: Not a clue, probably swallows them up. I can't imagine Localstack spins up VMs or anything for EC2. 
Ultimately, this is here just so that we can get past the `require_artifact()` check.

christophermaier: I guess my point is that if Localstack isn't going to do anything meaningful with this, should we put this behind an `if not LOCAL_GRAPL` check?

pulumi/infra/ec2_cluster.py
wimax-grapl: Yeah. I renamed the verbiage to "num_subclusters", "subcluster_size"

jgrillo-grapl: would it be less confusing to just have one number and leave the responsibility of incrementing it properly to the operator?

pulumi/grapl/Pulumi.local-grapl.yaml
wimax-grapl: Personally, I think that if localstack has nominal support for ec2 create-instance (even if it ignores the ami field and creates a docker instance or something), then the less stuff we can gate behind `if local_grapl` the better. 

But also - this is a battle I don't really care about, so - can do

christophermaier: How would it know what docker container to use?

wimax-grapl: update, I *dont* see anything about creating instances in ec2 localstack
https://github.com/localstack/localstack/blob/master/localstack/services/ec2/ec2_listener.py
so it might just be silently slurping up a request it doesn't recognize - so - feel better gating it

pulumi/infra/ec2_cluster.py
```diff
@@ -0,0 +1,92 @@
...
+        vpc_security_group_ids: Sequence[Output[str]],
+        instance_tags: Dict[str, str],
+        child_opts: pulumi.ResourceOptions,
+    ) -> List[aws.ec2.Instance]:
+        instances = []
+        for i in range(0, num_instances):
+            network_interface = aws.ec2.NetworkInterface(
+                f"ec2-eni-{name}-{i}",
+                subnet_id=vpc_private_subnet.id,
+                security_groups=vpc_security_group_ids,
```
wimax-grapl, line 60: Okay, so in one case we just pass in the `id` and in the other case we pass in the real object.

I actually like the real-object more (just form a typing perspective)

PR #1019, opened 2021-07-19
Trigger a UX artifact build, with debug logging

christophermaier:
The UX artifact build was not triggered when #1015 merged. This commit
should trigger a build, and with increased logging on the
`monorepo-diff` plugin to help sort out why.


wimax-grapl (approved):

wimax-grapl:
also, fwiw, you could consider doing something like
`packer/hax_change_this_to_trigger_ami_build.txt`

PR #1024, opened 2021-07-19
Preinstall Vector in our AMIs

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/603

Install Vector in our AMIs.
We will use Vector to forward Nomad logs to Kafka.
I've supplied a skeleton Vector config as well (just so people don't need to go change the pkr.hcl in the future once we need to upload a config)

The Vector provisioner file has a `vector --version` at the end of it that should, theoretically, throw a nonzero exit code if the install failed.


packer/nomad-server-client/image.pkr.hcl
```diff
@@ -167,7 +173,7 @@ build {
 
   provisioner "shell" {
     inline = [
-      "sudo yum install -y git",
+      "sudo yum install --assumeyes git",
```
christophermaier, line 176: 🎉

packer/nomad-server-client/image.pkr.hcl
```diff
@@ -201,6 +207,20 @@ build {
...
+    ]
+    script = "${path.root}/setup_vector.sh"
+  }
+
+  # `vector` invokes will need to specify this config, like
+  # `vector -c /tmp/vector-config/vector.toml`
+  provisioner "file" {
+    source      = "${path.root}/vector-config"
+    destination = "/tmp/"
+  }
```
christophermaier, line 222: Isn't there some place more permanent to put this than `/tmp`?

christophermaier (approved):
Seems legit 👍

PR #1025, opened 2021-07-20
Fix the UX asset creation

christophermaier:
Rather than building the tarball and uploading it all at once, we need
to actually split the jobs up (similar to how we do it with the lambda
zip files).

We'll build the tarball in the `node` container, upload the resulting
tarball to our artifact bucket, then upload to Cloudsmith in a
separate step.


jgrillo-grapl (approved):
⚫

PR #1028, opened 2021-07-20
Install OSQuery in our AMIs

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/604

Add `osquery` and a dummy-config-ready-to-be-edited to our AMIs. 

wishful thinking

### DIFF
--- a/packer/nomad-server-client/image.pkr.hcl
+++ b/packer/nomad-server-client/image.pkr.hcl
@@ -57,7 +57,7 @@ variable "aws_build_region" {
 }
 
 ####################
-# Program Versions
+# Software Versions
 ####################
 
 variable "terraform_aws_nomad_tag" {
@@ -188,7 +188,7 @@ build {
   }
 
   #################### 
-  # Install programs
+  # Install software
   #################### 
 
   provisioner "shell" {
--- a/packer/nomad-server-client/image.pkr.hcl
+++ b/packer/nomad-server-client/image.pkr.hcl
@@ -20,30 +20,6 @@ variable "terraform_aws_consul_tag" {
   default     = "v0.10.1"
 }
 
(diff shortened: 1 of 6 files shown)


packer/nomad-server-client/image.pkr.hcl
```diff
@@ -80,6 +56,40 @@ variable "aws_build_region" {
   default     = "us-east-1"
 }
 
+####################
+# Program Versions
```
wimax-grapl, line 60: i couldn't figure out what to call this. Dependency versions?

christophermaier: "Software Versions"

packer/nomad-server-client/setup_osquery.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Add repo
+curl -L https://pkg.osquery.io/rpm/GPG | sudo tee /etc/pki/rpm-gpg/RPM-GPG-KEY-osquery
+sudo yum-config-manager --add-repo https://pkg.osquery.io/rpm/osquery-s3-rpm.repo
+sudo yum-config-manager --enable osquery-s3-rpm-repo
+
+sudo yum install --assumeyes osquery
+
+# Ensure it's all working correctly
+osqueryi --version
```
christophermaier, line 16: Is this the actual name of the binary, or a typo?

christophermaier (approved):
Seems fine for what it is, but would be more useful with a real configuration in place.

packer/nomad-server-client/setup_osquery.sh
wimax-grapl: Good question, but yes, it is correct!

PR #1029, opened 2021-07-20
Fix up some environment variables for a CI deploy

christophermaier:
This is to get the provisioning pipeline flowing.

### DIFF
--- a/pulumi/infra/config.py
+++ b/pulumi/infra/config.py
@@ -91,8 +91,7 @@ def _require_env_var(key: str) -> str:
     return value
 
 
-# Boy, this env name was not forward-thinking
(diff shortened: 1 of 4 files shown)


jgrillo-grapl (approved):

pulumi/infra/config.py
```diff
@@ -92,7 +101,7 @@ def _require_env_var(key: str) -> str:
 
 
 # Boy, this env name was not forward-thinking
-OPERATIONAL_ALARMS_EMAIL = _require_env_var("GRAPL_CDK_OPERATIONAL_ALARMS_EMAIL")
+OPERATIONAL_ALARMS_EMAIL = _require_env_var("GRAPL_OPERATIONAL_ALARMS_EMAIL")
```
wimax-grapl, line 104: technically, this comment can be removed 🤣  but i won't force you to resubmit just for that

christophermaier: 🤦

PR #1030, opened 2021-07-20
Ensure we're using a clean git checkout before diffing against tags

christophermaier:
I recently noticed some erroneous builds based on the diff
plugin. Since our tags are changing, it's possible that we're not
getting the most up-to-date tags, particularly on machines that have
been servicing other jobs.

We'll explicitly declare that we want a clean checkout on these jobs
to help remedy this.


christophermaier:
Closed in favor of another approach in https://github.com/grapl-security/pipeline-infrastructure/pull/107

PR #1031, opened 2021-07-21
Rename query generators to remove "subgraph" component

christophermaier:
This was initially motivated because our Pulumi code assumed artifacts named "sysmon-generator" and "osquery-generator", but our container names (and all the code underlying them) were named "sysmon-subgraph-generator" and "osquery-subgraph-generator".

To ensure a consistent mapping between artifact and service in our Pulumi code, and an overall consistency in the codebase in general, we have removed the "subgraph" part of the names.


wimax-grapl (approved):
Not sure why Cargo Audit decided to run (and fail) here.

inickles-grapl (approved):
Thanks for cleaning this up!

christophermaier:
cargo-audit is configured to run on PRs that change the Cargo.XXX files: https://github.com/grapl-security/grapl/blob/0940109b5f37aff82f95698b0297f615a2b1e084/.github/workflows/cargo-audit.yml#L9-L16

PR #1032, opened 2021-07-21
Run cargo audit in Buildkite

christophermaier:
Introduces a step that will run `cargo audit` in our verify pipeline
when `Cargo.lock` or `Cargo.toml` files change. Failures here will not
fail the pipeline, but will merely be advisory.

A separate pipeline (`grapl/cargo-audit`) has also been set up and
uses the same pipeline definition. In that pipeline, however, failures
will actually be failures. This pipeline will be run on a schedule,
currently once a week. (This separate pipeline has been set up outside
of this PR).

The old Github Action for running `cargo audit` has also been removed.


wimax-grapl (approved):

.buildkite/pipeline.verify.yml
```diff
@@ -158,3 +158,17 @@ steps:
...
+      - chronotc/monorepo-diff#v2.0.4:
+          diff: .buildkite/shared/scripts/diff.sh
+          log_level: "debug"
+          watch:
+            - path:
+                - "**/Cargo.toml"
+                - "**/Cargo.lock"
+              config:
+                label: ":pipeline: Upload Cargo Audit"
+                command: ".buildkite/pipeline.cargo-audit.sh | buildkite-agent pipeline upload"
```
wimax-grapl, line 174: this is so wild but i kind of love it

PR #1039, opened 2021-07-26
Use ENTRYPOINT instead of CMD for Rust containers

christophermaier:
This means that our containers will behave as though they were the
individual Rust binary, rather than being passed as an argument to
whatever ENTRYPOINT may be set by our base container.

(This doesn't change any behavior right now, but makes things a bit
safer and explicit moving forward.)

See the following for details:
- https://docs.docker.com/engine/reference/builder/#cmd
- https://docs.docker.com/engine/reference/builder/#entrypoint


jgrillo-grapl (approved):
👍

PR #1044, opened 2021-07-27
Pass bucket names to analyzer-executor directly

christophermaier:
Rather than composing the names of our buckets using logic internal to the `analyzer-executor`, we will now pass the full name directly to the service using environment variables.

This will lay the groundwork for ultimately getting rid of concretely-managed bucket names in Pulumi.


jgrillo-grapl (approved):
🙏

PR #1045, opened 2021-07-27
Pass bucket to model-plugin-deployer directly

christophermaier:
Same principle as https://github.com/grapl-security/grapl/pull/1044

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -375,7 +375,6 @@ services:
       <<: *dgraph-env
       IS_LOCAL: "True"
       <<: *log-level
-      UX_BUCKET_URL: "localhost"
     tty: true
     depends_on:
       dgraph:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -359,15 +359,17 @@ services:
 
   model-plugin-deployer:
     image: grapl/model-plugin-deployer:${TAG:-latest}
-    command: |
-      /bin/sh -c '
-        . venv/bin/activate &&
-        cd /home/grapl/app &&
+    entrypoint: ["/bin/bash", "-o", "errexit", "-o", "nounset", "-c"]
+    command:
+      - |
+        export GRAPL_MODEL_PLUGINS_BUCKET=$$(cat /pulumi-outputs/model-plugins-bucket)
+        . venv/bin/activate
+        cd /home/grapl/app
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -382,6 +383,14 @@ services:
         condition: service_healthy
       provisioner:
         condition: service_completed_successfully
+      pulumi:
+        # We must wait until Pulumi has created the buckets to know what their names are
+        condition: service_completed_successfully
+    volumes:
+      - type: volume
+        source: pulumi_outputs
+        target: /pulumi-outputs
```
wimax-grapl, line 392: (Not a must-do, but I sort of would prefer if target were `/mnt/pulumi-outputs` so it's more obvious this is a volume). However, I recognize I'm asking to reinvent the existing wheel - more something to consider in the future.

docker-compose.yml
```diff
@@ -359,21 +359,22 @@ services:
 
   model-plugin-deployer:
     image: grapl/model-plugin-deployer:${TAG:-latest}
-    command: |
-      /bin/sh -c '
-        . venv/bin/activate &&
-        cd /home/grapl/app &&
+    entrypoint: ["/bin/bash", "-o", "errexit", "-o", "nounset", "-c"]
```
wimax-grapl, line 362: Would it be better to specify the entrypoint in the image?  (This should probably be the entrypoint for everything so we can stop doing && stuff)

wimax-grapl (approved):

jgrillo-grapl (approved):

docker-compose.yml
christophermaier: As for baking this into the images: that's not actually a blanket assumption that we can (or even should) make; it all depends on the nature of the container. The fact that we're having to do all this "extra" work in docker-compose is a bit unique, and not something I'd want to enshrine in our containers.

As for the `&&` stuff, yes, this should ultimately be the entrypoint we use in our docker-compose services that are currently doing `/bin/bash -c 'blah && blah && blah`. I'm updating that as I'm going along with the various services I'm updating.

docker-compose.yml
christophermaier: I could take another pass through everything and do that, on a separate PR.

PR #1052, opened 2021-07-27
Fix names of most lambdas; remove some cdk stuff in graplctl

wimax-grapl:
None, but there was notably a regression on how long we retain cloudwatch logs fixed by migrating them back to the correct name.

- Remove some CDK-specific logic from graplctl.
- Existing lambdas were mostly named without DEPLOYMENT_NAME - e.g. named `graphql_endpoint` 
- Now all lambdas have DEPLOYMENT_NAME suffix.
- 

before and after
image

image

### DIFF
--- a/pulumi/infra/lambda_.py
+++ b/pulumi/infra/lambda_.py
@@ -158,7 +158,6 @@ class Lambda(pulumi.ComponentResource):
         name: str,
         args: LambdaArgs,
         network: Network,
-        override_name: Optional[str] = None,
         opts: Optional[pulumi.ResourceOptions] = None,
     ) -> None:
         super().__init__("grapl:Lambda", name, None, opts)
@@ -174,9 +173,10 @@ class Lambda(pulumi.ComponentResource):
         )
 
         lambda_name = f"{DEPLOYMENT_NAME}-{name}"
+        # remove these from scope - everyone should use `lambda_name`
         self.function = aws.lambda_.Function(
             f"{name}-lambda",
-            name=override_name or name,
+            name=lambda_name,
             description=args.description,
             runtime=args.runtime,
             package_type=args.package_type,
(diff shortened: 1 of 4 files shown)


pulumi/infra/lambda_.py
```diff
@@ -173,10 +172,10 @@ def __init__(
             opts=pulumi.ResourceOptions(parent=self),
         )
 
-        lambda_name = f"{DEPLOYMENT_NAME}-{name}"
+        name_with_deployment_prefix = f"{DEPLOYMENT_NAME}-{name}"
         self.function = aws.lambda_.Function(
             f"{name}-lambda",
-            name=override_name or name,
+            name=name_with_deployment_prefix,
```
christophermaier, line 178: If we've only got the one lambda that has to have an explicit name, why don't we keep the "override name" logic (modified to fix this bug, of course), but continue to let the other lambdas get auto-generated names? That way, when we fix things so `graplctl` can just take a lambda name from Pulumi directly, the change to rip out the override name logic will only affect that one lambda, rather than all of them.

Pretty sure you could still have the log group names depend on the generated name (using a `pulumi.Output.apply`).

wimax-grapl: Yeah, that's doable

pulumi/infra/lambda_.py
```diff
@@ -173,10 +173,10 @@ def __init__(
             opts=pulumi.ResourceOptions(parent=self),
         )
 
-        lambda_name = f"{DEPLOYMENT_NAME}-{name}"
+        lambda_name = override_name or f"{DEPLOYMENT_NAME}-{name}"
         self.function = aws.lambda_.Function(
             f"{name}-lambda",
-            name=override_name or name,
+            name=lambda_name,
```
christophermaier, line 179: This is still explicitly setting a name for everything... it would need to be `name = override_name or None`

christophermaier (approved):

PR #1054, opened 2021-07-28
Revise Pulumi mounts in docker-compose files

christophermaier:
Now, our Pulumi outputs mount is attached to `/mnt/pulumi-outputs`, to
emphasize its "volume-ness". Additionally, we extract the mount
definition itself to a reusable YAML chunk to reduce repetition.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,11 @@ volumes:
     # need to be accessible in test containers.
 
 x-common-variables:
+  read-only-pulumi-mnt: &read-only-pulumi-mnt
+    type: volume
+    source: pulumi_outputs
+    target: /mnt/pulumi-outputs
+    read_only: true
   aws-env: &aws-env
     GRAPL_AWS_ENDPOINT: ${GRAPL_AWS_ENDPOINT}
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -11,6 +11,11 @@ volumes:
     # need to be accessible in test containers.
 
 x-common-variables:
+  read-only-pulumi-mnt: &pulumi-mnt
+    type: volume
+    source: pulumi_outputs
+    target: /mnt/pulumi-outputs
+    read_only: true
```
wimax-grapl, line 18: thank you - I know some people dislike the yaml <<: stuff - but this is nice reuse :) 

My only thing is that I usually keep the key (`read-only-pulumi-mnt`) and the anchor (`&pulumi-mnt`) the same - but not the end of the world.

test/docker-compose.integration-tests.yml
```diff
@@ -7,6 +7,13 @@ volumes:
   # here so we can have access to outputs from the stack in our tests.
   pulumi_outputs:
 
+x-common-variables:
+  read-only-pulumi-mnt: &pulumi-mnt
```
wimax-grapl, line 11: same here

docker-compose.yml
christophermaier: Yup, good call.

PR #1059, opened 2021-07-29
Basic Nomad Support

d0nut-grapl:

Internal #602

Adds nomad job definitions for grapl core, grapl local infra, and grapl generators.
Also adds some local development scripts to make it easier to do testing.

Here's how you use these locally.

1. Start a local docker registry
```bash
./nomad/local/start_local_registry.sh
```

2. Build Grapl for local deployment
```bash
REGISTRY=localhost:5000 make build-services
```

3. Push Grapl locally
```bash
make push-local
```

4. Start local development environment
```bash
sudo ./nomad/local/start_development_environment.sh
```

5. Deploy Grapl Core
```bash
# Where <IP> is the ip of linux vm
# e.g. 100.127.1.123

nomad job run \
    -var "container_registry=localhost:5000" \
    -var "aws_sqs_url=http://<IP>:9324" \
    -var "dgraph_replicas=1" \
    -var "dgraph_shards=1" \
    -var "redis_endpoint=redis://<IP>:6379" \
    nomad/grapl-core.nomad
```

6. Deploy Grapl Generator
```bash
./nomad/local/deploy_generator.sh
```
Note: You will need to use a web url to fetch the generator artifact. To do this locally, you can use `python3 -m http.server -p <port>` in the directory that contains the generator artifact you wish to deploy.


nomad/grapl-generator.nomad
```diff
@@ -0,0 +1,72 @@
...
+            port "generator_receiver" { }
+        }
+
+        count = var.generator_count
+
+        task "tenant-generator" {
+            driver = "exec"
+
+            artifact {
+                source = var.generator_artifact_url
```
christophermaier, line 50: What kind of artifact are we expecting here? Just a bare file (i.e., not zipped or compressed at all)? Where is it supposed to come from?

nomad/grapl-generator.nomad
```diff
@@ -0,0 +1,72 @@
...
+
+            config {
+                command = "generator"
+            }
+
+            service {
+                name = "generator-${var.tenant_id}-${var.generator_ingest}"
+                port = "generator_receiver"
+                tags = ["generator-${var.tenant_id}"]
+                canary_tags = ["canary-generator-${var.tenant_id}"]
```
christophermaier, line 68: Should this not incorporate the `generator_ingest` as well (same for `tags`, too)?

nomad/grapl-generator.nomad
```diff
@@ -0,0 +1,72 @@
...
+                mode = "file"
+            }
+
+            env {
+                TENANT_ID = "${var.tenant_id}"
+                INGEST_TYPE = "${var.generator_ingest}"
+            }
+
+            config {
+                command = "generator"
```
christophermaier, line 61: Is this implicitly relative to `local/` (per line 51 above)?

nomad/grapl-generator.nomad
```diff
@@ -0,0 +1,72 @@
...
+        # The min amount of reported "healthy" time before a instance is considered healthy and an allocation is opened up for further updates
+        min_healthy_time = "15s"
+    }
+
+    group "generator" {
+        network {
+            port "generator_receiver" { }
+        }
+
+        count = var.generator_count
```
christophermaier, line 44: Is this actually used?

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
+# For rust logging output
```
christophermaier, line 1: Rather than use comments to describe the variables, it would be good to use the `description` field instead.

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+variable "RUST_LOG" {
```
christophermaier, line 2: The fact that this is the only variable in SCREAMING_SNAKE_CASE is odd... we should name them all consistently.

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+variable "node_identifiers_retry" {
+    type = number
+    default = 1
+}
+
+locals {
+    # DGraph Alphas (shards * replicas)
+    dgraph_alphas = [for alpha_id in range(0, var.dgraph_replicas * var.dgraph_shards): {
+        id: alpha_id,
+        grpc_private_port: alpha_id + 7080,
```
christophermaier, line 81: This is admittedly _super_ nitpicky, but having the "base number" come first, and the variable second reads more naturally (to me, at any rate 😅)

e.g.,
```
7080 + 1
7080 + 2
7080 + 3
```
vs
```
1 + 7080
2 + 7080
3 + 7080
```

christophermaier: Also, why is it necessary for each shard to use a different port? They should be independent, no?

If this is to support local usecases, aren't there other ways of mapping concrete ports from the host to the individual service?

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+    dgraph_zeros = [for zero_id in range(1, var.dgraph_replicas): {
+        id: zero_id,
+        grpc_private_port: zero_id + 5080,
+    }]
+
+    # String that contains all of the Zeros for the Alphas to talk to and ensure they don't go down when one dies
+    zero_alpha_connect_str = join(",", [for zero_id in range(0, var.dgraph_replicas): "localhost:${zero_id + 5080}"])
+
+    # String that contains all of the running Alphas for clients connecting to Dgraph (so they can do loadbalancing)
+    alpha_grpc_connect_str = join(",", [for alpha in local.dgraph_alphas: "localhost:${alpha.grpc_public_port}"])
```
christophermaier, line 96: Is this `localhost` here specifically limiting this file to being used just on our workstations, or is this somehow mapped via Consul / DNS / Hashimagic to be used as-is in real multi-host deployments?

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+    group "dgraph-zero-leader" {
+        network {
+            mode = "bridge"
+        }
+
+        task "dgraph-zero-leader" {
+            driver = "docker"
+
+            config {
+                image = "dgraph/dgraph:latest"
```
christophermaier, line 128: We should make `latest` a variable so our deploys aren't just floating on whatever Dgraph just happened to release.

nomad/grapl-core.nomad
christophermaier: Also, should these base numbers be extracted to variables / locals? I see they're used in multiple places, and magic numbers make me nervous.

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+                args = [
+                    "dgraph",
+                    "zero",
+                    "--my=localhost:5080",
+                    "--replicas", "${var.dgraph_replicas}",
+                    "--raft", "idx=1",
```
christophermaier, line 134: Any significance to having the flag and value in a single value in one case (`--my=localhost:5080`), but split up in others (`--replicas`, `${var.dgraph_replicas}`)?

It would be nice to adopt a single convention.

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+
+                config {
+                    image = "dgraph/dgraph:latest"
+                    args = [
+                        "dgraph",
+                        "zero",
+                        "--my=localhost:${zero.value.grpc_private_port}",
+                        "--replicas", "${var.dgraph_replicas}",
+                        "--raft", "idx=${zero.value.id + 1}",
+                        "-o", "${zero.value.id}",
```
christophermaier, line 195: Is there a long option form for `-o` that we can use?

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+        network {
+            mode = "bridge"
+        }
+
+        task "graph-merger" {
+            driver = "docker"
+
+            config {
+                image = "${var.container_registry}/graph-merger:${var.tag}"
+                args = ["/graph-merger"]
```
christophermaier, line 338: Is this supposed to be a command?

If so, this should not be required anymore, because the Rust containers all have their respective binaries set as the container entrypoints now. Simply running the container will be the equivalent of running that binary.

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+            }
+
+            env {
+                REDIS_ENDPOINT = "${var.redis_endpoint}"
+                MG_ALPHAS = "${local.alpha_grpc_connect_str}"
+                GRAPL_SCHEMA_TABLE = "${var.schema_table_name}"
+                AWS_REGION = "${var.aws_region}"
+                # https://github.com/grapl-security/grapl/blob/18b229e824fae99fa2d600750dd3b17387611ef4/pulumi/grapl/__main__.py#L165
+                DEST_BUCKET_NAME = "subgraphs-merged-bucket"
```
christophermaier, line 347: This isn't going to be the right bucket name. Currently, we still prepend the stack name, but soon, it will have a dynamically-generated suffix from Pulumi.

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+                SOURCE_QUEUE_URL = "${var.aws_sqs_url}/${var.aws_account_id}/graph-merger-queue"
+                DEAD_LETTER_QUEUE_URL = "${var.aws_sqs_url}/${var.aws_account_id}/graph-merger-dead-letter-queue"
```
christophermaier, line 349: Same thing applies to these queue names.

(This naming issue applies to all the other containers, as well, so I won't repeat myself.)

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+        network {
+            mode = "bridge"
+        }
+
+        task "node-identifier" {
+            driver = "docker"
+
+            config {
+                image = "${var.container_registry}/node-identifier:${var.tag}"
+                args = ["/node-identifier"]
```
christophermaier, line 384: This is the container's entrypoint now.

This applies to all the other Rust service containers, as well.

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+                    "zero",
+                    "--my=localhost:5080",
+                    "--replicas", "${var.dgraph_replicas}",
+                    "--raft", "idx=1",
+                ]
+            }
+        }
+
+        service {
+            name = "dgraph-zero-leader-grpc-private"
```
jgrillo-grapl, line 140: What happens if dgraph does a leader election and determines this node is no longer the leader? Won't that cause this service to be "mislabeled" in nomad? If that's the case then I'd suggest just using a numeric naming scheme e.g. `dgraph-zero-0`, `dgraph-zero-1`, ... with the understanding that e.g. `dgraph-zero-0` starts off as leader.

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+
+        network {
+            mode = "bridge"
+        }
+
+        task "node-identifier-retry" {
+            driver = "docker"
+
+            config {
+                image = "${var.container_registry}/node-identifier-retry:${var.tag}"
```
christophermaier, line 412: FYI, with the way we're managing our containers and pinning their versions in the release pipeline, there is no guarantee that all the services' tags will be the same (I think for the time being they are by default, but this will not always be the case).

nomad/grapl-core.nomad
jgrillo-grapl, line 128: We should pick a version and pin it

nomad/local/deploy_generator.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Creates a namespace in nomad and deploys a generator into it
+
+read -p "Tenant ID: " tenant_id
+read -p "Generator Ingest: " generator_ingest
+read -p "Artifact URL: " artifact_url
+read -p "Instances: " generator_count
```
christophermaier, line 8: How is this script meant to be invoked? Why are you using `read` rather than just referencing arguments?

nomad/grapl-generator.nomad
```diff
@@ -0,0 +1,72 @@
+variable "tenant_id" {
+    type = string
+    description = "The opaque tenant id."
+}
+
+variable "generator_ingest" {
+    type = string
+    description = "The expected data ingest type. e.g. sysmon."
```
jgrillo-grapl, line 8: I think this should really be a UUID, not a string. See https://github.com/grapl-security/grapl-rfcs/pull/1#discussion_r678683586 for rationale.

nomad/grapl-generator.nomad
```diff
@@ -0,0 +1,72 @@
+variable "tenant_id" {
+    type = string
+    description = "The opaque tenant id."
+}
+
+variable "generator_ingest" {
```
jgrillo-grapl, line 6: Can we rename this to something more descriptive? Perhaps `data_type_id`?

nomad/grapl-generator.nomad
```diff
@@ -0,0 +1,72 @@
+variable "tenant_id" {
+    type = string
+    description = "The opaque tenant id."
```
jgrillo-grapl, line 3: What does "opaque" mean here? This is just the tenant UUID right?

nomad/grapl-generator.nomad
```diff
@@ -0,0 +1,72 @@
...
+}
+
+variable "generator_ingest" {
+    type = string
+    description = "The expected data ingest type. e.g. sysmon."
+}
+
+variable "generator_artifact_url" {
```
jgrillo-grapl, line 11: We should have a conversation about this... I'd recommend that we flesh this all out in the RFC first.

nomad/local/deploy_generator.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
christophermaier, line 1: We should add a `BUILD` file to `nomad/local` with the following contents:

```
shell_library()
```

This will ensure that Pants will keep these scripts formatted and linted like all our other scripts.

christophermaier: Related, I assume that `nomad` has some facility for formatting it's HCL files, similar to how `packer` does.

It would be good to put that into the `Makefile`, and get it wired up into our verify pipeline.

jgrillo-grapl (commented):
A few minor things, overall LGTM

nomad/grapl-generator.nomad
d0nut-grapl: > What kind of artifact are we expecting here? Just a bare file?

`artifact` automatically (if you don't disable it) decompresses archives (zip, tgz, bz2, and xz) so it doesn't necessarily have to be a bare file.

> Where is it supposed to come from?

We can host these in an S3 bucket as artifact supports http(s) and s3 (as well as git and mercurial).

nomad/grapl-generator.nomad
d0nut-grapl: The intent here was that if we wanted to easily identify all of the generators for a particular tenant, we can open the consul interface and just select the `generator-<tenant_id>` tag. We could also add a `ingest-<generator_ingest>` tag if we want to find all of the generators that support a particular ingest type too!

Do you think that'd be useful?

nomad/grapl-generator.nomad
d0nut-grapl: Yup! By default it'll look in the `local` folder :)

nomad/grapl-generator.nomad
d0nut-grapl: I *suspect* it is if generators are unable to keep up with the load. Otherwise, we're asking that a single machine handle the entire load of processing all logs of a particular ingest type (which for many smaller companies could be a reality, but for larger organizations may not be true).

nomad/grapl-core.nomad
d0nut-grapl: Yup! My bad :P

nomad/grapl-core.nomad
d0nut-grapl: > This is admittedly super nitpicky, but...

Good point :) I'll change that!

> why is it necessary for each shard to use a different port? They should be independent, no?

Ideally, yes! However, we need to use different ports due to the way that consul connect works. In consul connect land, different port == different upstream. It's *probably* possible to make it so that the local port is the same in every instance, but seeing as the `service`s would have to operate on different ports it seemed fine fine to use different ports.

> Also, should these base numbers be extracted to variables / locals?

Good idea on the locals! I can move them over :)

nomad/grapl-core.nomad
d0nut-grapl: > Is this localhost here specifically limiting this file to being used just on our workstations, or is this somehow mapped via Consul / DNS / Hashimagic to be used as-is in real multi-host deployments?

This is for consul connect. In consul connect land, all of the remote systems appear as "localhost" (see: my grapl hashistack presentation)

nomad/grapl-core.nomad
d0nut-grapl: Yup! Good point :)

nomad/grapl-core.nomad
d0nut-grapl: No particular point! I think I was doing a combination of copy+paste from dgraph docs and manually writing/messing about with the config to get it working originally.

nomad/grapl-core.nomad
d0nut-grapl: I don't know offhand, but I suspect that it'll probably be `--offset`. I'll look and see if I can update it :)

nomad/grapl-core.nomad
d0nut-grapl: > Is this supposed to be a command?

Yup! I needed it to actually get the container running (in my testing).

> ..this should not be required anymore..

I'll give it a go! Thanks for the heads up!

nomad/grapl-core.nomad
d0nut-grapl: > This isn't going to be the right bucket name.

This was a section that I was struggling with, honestly. I was trying to understand, for our multi-tenant future, what role the deployment name prefix would play (which I suspected it wouldn't exist anymore). If we're going to make a dynamically generated bucketname, we can always pass it in as a variable.

nomad/grapl-core.nomad
d0nut-grapl: Gotcha; so we'll need a tag-per-service, then?

nomad/grapl-generator.nomad
christophermaier: The way that the Pulumi code works now is that it knows the base URL for our Cloudsmith repository, and then it adds to that the artifact name and the pinned version we're using in order to get a final URL. As long as whatever is invoking this can get that URL here, we should be good.

Actually, what artifacts are these? Are they things we're currently generating?

nomad/local/deploy_generator.sh
d0nut-grapl: I made this script as Ian had brought up concerns about the difficulty of local development. I'm far from a bash expert so, unfortunately, I don't really know of the alternative way to do this. 

If you have an alternative recommendation, would you mind showing me what you mean by referencing arguments?

nomad/grapl-generator.nomad
christophermaier: I haven't messed around with Nomad's tagging yet; can you search on multiple tags at once? If so, then it might just be better to tag by each individual bit of data and let the person doing the search compose them as they need, rather than us having to guess beforehand what kinds of searches we're likely to want to do.

Are tags just a single value, or can they be a key/value pair, like in AWS?

nomad/grapl-generator.nomad
christophermaier: 🆒

nomad/grapl-core.nomad
d0nut-grapl: > .. Won't that cause this service to be "mislabeled" in nomad?

Yup! I can't remember why, but I had a reason why I didn't do the `dgraph-zero-N` thing before.. Ehh, if it comes up again I'll point it out, but yea we should probably make them all homogeneous since, after deployment, they'll have a real leader.

nomad/grapl-generator.nomad
d0nut-grapl: We can update the comment specifically if we want, but the `type` field is a type in HCL2 (https://www.nomadproject.io/docs/job-specification/hcl2/expressions#types-and-values).

nomad/grapl-core.nomad
christophermaier: FYI, this changed in https://github.com/grapl-security/grapl/pull/1039

nomad/grapl-generator.nomad
d0nut-grapl: Yup tenant UUID

nomad/grapl-core.nomad
christophermaier: Is every tennant going to get their own bucket?

If not, then the we'll just need to accept whatever the right bucket name is from Pulumi when this is deployed.

On the other hand, if everyone is getting their own bucket, we have some more planning to do 😬

nomad/grapl-core.nomad
christophermaier: Nice

nomad/grapl-core.nomad
christophermaier: > Ideally, yes! However, we need to use different ports due to the way that consul connect works. In consul connect land, different port == different upstream. It's probably possible to make it so that the local port is the same in every instance, but seeing as the services would have to operate on different ports it seemed fine fine to use different ports.

I need to get more familiar with Nomad to fully grok this, but for now it's probably fine as it is. I'm sure all this stuff will be evolving anyway 😄

nomad/grapl-generator.nomad
christophermaier: Oh, sorry... not quite sure where my brain went. I read this and thought it was declaring a variable that wasn't used elsewhere in this block. And then I remembered "Wait, that's not how HCL works" 😅 

Neeeeeever miiiiiiind...

nomad/grapl-core.nomad
christophermaier: Eventually, yes.

nomad/grapl-core.nomad
wimax-grapl: We use a fixed version elsewhere in our code - :v21.03.1 i believe

nomad/local/deploy_generator.sh
christophermaier:
``` bash
#!/usr/bin/env bash

set -euo pipefail

readonly tenant_id="${1}" # the first argument
readonly generator_ingest="${2}" # the second argument, etc.
readonly artifact_url="${3}"
readonly generator_count="${4}"

# rest of script follows
```
You can get more fancy with flags and whatnot, but this would be the minimum.

This would let you call it like so:

``` bash
nomad/local/deploy_generator.sh my-tenant-id sysmon https://blah.com/artifact.tar.gz 1
```
As it exists now, you can only use this script interactively, which limits it's usefulness.

Incidentally, there is also nothing in the current script that prevents you from just mashing Enter at all those prompts and having a lot of empty variables, which will end in tears. With the `set -u` flag in my example above, it will at least require that all those arguments are present.

nomad/local/start_local_registry.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Starts up a local docker registry.
+# This is required for local nomad deployments
+
+docker run -d -p 5000:5000 --name registry registry:2.7
```
wimax-grapl, line 6: Would there be a way to tie this into our makefile? 

So that it'd look something like
```
.PHONY: run-registry
run-registry:
 	# start local registry *if it hasn't already been started* 

.PHONY: push-local
push-local: run-registry ## Push Grapl container to local registry
	REGISTRY=localhost:5000 docker-compose --file=docker-compose.build.yml push
```

it feels odd that there's now a manual step before we can `make` anything

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,428 @@
...
+                port = "${alpha.value.http_port}"
+
+                connect {
+                    sidecar_service { }
+                }
+            }
+        }
+    }
+
+    group "grapl-graph-merger" {
```
wimax-grapl, line 326: (Doesn't have to be done during this PR)
Are we able to split up the different `group`s at all or do they all have to be in 1 file?

nomad/local/deploy_generator.sh
wimax-grapl: consider a `set -euo pipefail` which is, in general, a decent base bash setup (no unset variables, bail out if something exits nonzero, etc)

nomad/local/deploy_generator.sh
wimax-grapl: i mean the easiest thing (which I personally don't like) is
```
readonly tenant_id="$1"
readonly generator_ingest="$2"
```

or, more of a pain but more explicit:
```
TENANT_ID=12345 GENRATOR_INGEST=12345 deploy_generator.sh
```

nomad/grapl-core.nomad
d0nut-grapl: True! Fixed :)

I also realized I never used the variable! I think it was the very first thing I wrote in the job and forgot to use it lol

nomad/grapl-core.nomad
d0nut-grapl: It turned out to be `--port_offset`; Updated :)

nomad/grapl-core.nomad
d0nut-grapl: Seemed to work for me! Thanks for the heads up :)

nomad/grapl-core.nomad
d0nut-grapl: I actually think we're not going to be using buckets anymore when we move to kafka (if I understand kafka correctly). In this future world, we'll have topics that we're pushing messages through (where messages are a combination of the data and tenant id information).

In the meantime, it's probably fine to leave this as-is as we'll be deleting it (unless I misunderstood. cc: @jgrillo-grapl )

nomad/grapl-core.nomad
d0nut-grapl: I'll add tags for the graph merger and node identifier. I feel like it's safe to assume that node-identifier and node-identifier-retry will share a tag. If not, let me know!

nomad/local/deploy_generator.sh
d0nut-grapl: Ah! I see what you mean now. I choose interactively because I hate having a billion command line arguments if I can help it, but this is totally fine too. I'll update it to use command line arguments instead. Thanks for the help :)

nomad/grapl-core.nomad
christophermaier: Yes, I believe with how things are set up now, node-identifer and node-identifier-retry will always have the same release tag.

nomad/local/deploy_generator.sh
christophermaier: If you _really_ want interactive, you can always add an `--interactive` option, that (if present) would use this `read`-based logic. If it's not there, then it assumes you're passing everything in as arguments.

christophermaier:
> EDIT: One other question--is this PR meant to replace compose? If so, let's also delete all the compose stuff and update the documentation. IMO it doesn't make sense to have two different ways of running grapl in `main` at the same time, as only the one which is exercised in CI will remain functioning. The other will go stale.

@jgrillo-grapl All else being equal, I don't think this can replace our docker-compose setup right now, unless we were to pull over all the "base" containers that we're currently using. In particular, until we're using Kafka everywhere, we rely on the notifications from bucket inserts to populate our pipeline. Unless what's here now can do that, we won't be able to make that switch.

nomad/grapl-core.nomad
d0nut-grapl: Updated this to use a tag provided at deployment time. It'll default to v21.03.1 now.

nomad/local/start_local_registry.sh
d0nut-grapl: Oh yea that's a great idea. I don't know why that didn't cross my mind. I'll migrate that over right away.

nomad/grapl-core.nomad
d0nut-grapl: As far as I'm aware, it must all be defined together :/

nomad/grapl-core.nomad
christophermaier: My point is that, regardless of when Kafka lands, this is _not_ fine to leave as-is... there is nowhere in our infrastructure, local or otherwise, where we have a bucket named simply "subgraphs-merged-bucket", therefore this will not work with our existing code.

nomad/grapl-generator.nomad
d0nut-grapl: > I haven't messed around with Nomad's tagging yet; can you search on multiple tags at once? 

Just to clear up confusion, the tags on a `service` stanza are for Consul, not Nomad. When you use the Consul interface, it'll let you filter down the services more easily by searching using the search bar. You also don't need the full name, either. You could search for `canary` and all of the `canary-generator-<tenant_id>` tagged instances would should up.

> If so, then it might just be better to tag by each individual bit of data and let the person doing the search compose

I assume you mean tag with `tenant_id` and `generator_ingest` individually? I could do that!

> Are tags just a single value...?

Single value, yea. There is a metadata field you can add on though, if you want that. I don't know if metadata can be searched but I suspect no.

nomad/grapl-core.nomad
d0nut-grapl: I think we can discuss this on the call tomorrow so it's more clear.

nomad/grapl-generator.nomad
christophermaier: Gotcha; makes sense re: Consul vs. Nomad.

Yeah, I think going with the more atomic tags makes more sense at this point. We can `AND` things together when we need to. Ultimately, I think this will end up being more flexible.

christophermaier: As for the single vs K/V question, I was thinking it'd be nice if we could say something like `tenant_id: BLAH`, but using `tenant_id_${BLAH}` is fine (particularly if that's our only option 😅)

nomad/grapl-core.nomad
christophermaier: Fair enough.

nomad/local/deploy_generator.sh
d0nut-grapl: It's fine; I'm not married to the idea. I can just use command line arguments hahaa

Makefile
```diff
@@ -413,6 +413,10 @@ clean-artifacts: ## Remove all dumped artifacts from test runs (see dump_artifac
 push: ## Push Grapl containers to Docker Hub
 	docker-compose --file=docker-compose.build.yml push
 
+.PHONY: push-local
+push-local: ## Push Grapl container to local registry
+	REGISTRY=localhost:5000 docker-compose --file=docker-compose.build.yml push
```
wimax-grapl, line 418: actually, we might be able to include this REGISTRY= as part of the WITH_LOCAL_GRAPL_ENV stuff - thoughts? 

That'd be more in line with the rest of our makefile

(also if we could rename it to like DOCKER_REGISTRY or something a bit more namespaced thatd be cool)

nomad/grapl-core.nomad
```diff
@@ -40,60 +40,78 @@ variable "dgraph_shards" {
...
 
 # https://github.com/grapl-security/grapl/blob/af6f2c197d52e9941047aab813c30d2cbfd54523/pulumi/infra/dynamodb.py#L118
-# Without deployment name
 variable "session_table" {
     type = string
     default = "dynamic_session_table"
 }
 
-# How many graph mergers should be running in tandem
 variable "graph_mergers" {
```
wimax-grapl, line 59: I'd personally prefer `num_graph_merger` et al but, certainly not a blocking suggestion

nomad/grapl-core.nomad
```diff
@@ -116,29 +134,30 @@ job "grapl-core" {
...
             driver = "docker"
 
             config {
-                image = "dgraph/dgraph:latest"
+                image = "dgraph/dgraph:${var.dgraph_tag}"
                 args = [
                     "dgraph",
                     "zero",
-                    "--my=localhost:5080",
+                    "--my", "localhost:5080",
```
wimax-grapl, line 150: Not a must have, but worth noting we could totally do 
`localhost:${local.dgraph_zero_grpc_private_port_base}` 

there's a number of other places with an explicit 5080 in this PR that we could do similar string interpolation with; I won't highlight them all, just ctrl-f

nomad/grapl-core.nomad
wimax-grapl: in fact: this is just busywork for you, i don't care, please ignore

nomad/grapl-generator.nomad
wimax-grapl: let's update the name to tenant_uuid or update the description?

nomad/grapl-core.nomad
christophermaier: Counterpoint: clarity in naming is important, because we're going to be reading this code a lot more than we're writing it.

Makefile
d0nut-grapl: Ahh! My bad I haven't had to update the makefile before. That's a cool pattern I'll use instead here.

docker-compose.build.yml
```diff
@@ -12,7 +12,7 @@ services:
   ########################################################################
 
   sysmon-generator:
-    image: grapl/sysmon-generator:${TAG:-latest}
+    image: ${REGISTRY:-grapl}/sysmon-generator:${TAG:-latest}
```
inickles-grapl, line 15: I think we'll want this to be: `${REGISTRY:-registry-1.docker.io}/grapl/sysmon-generator:${TAG:-latest}`, where registry-1.docker.io is Docker's default registry. This way the image name will still be prefixed with "grapl/".

Makefile
```diff
@@ -156,6 +156,12 @@ build-test-e2e: build-services
 build-python-wheels:  ## Build all Python wheels
 	./pants filter --target-type=python_distribution :: | xargs ./pants package
 
+.PHONY: build-services-local
+build-services-local: 
+	$(WITH_LOCAL_GRAPL_ENV) \
+	$(MAKE) build-services
+	$(MAKE) push-local
```
christophermaier, line 163: Since `build-services` already uses `WITH_LOCAL_GRAPL_ENV`, we don't gain anything by doing it here, too.

At that point, it's probably better to write this target as:

```make
build-services-local: build-services
    $(MAKE) push-local
```
But then we have the question: is `build-services-local` really just an alias for `push-local`? Should we just get rid of `build-services-local` and have `push-local` depend on `build-services`? Is there a scenario where you would legitimately need to call `push-local` without building the services?

Makefile
```diff
@@ -412,13 +418,19 @@ clean-mount-cache: ## Prune all docker mount cache (used by sccache)
 clean-artifacts: ## Remove all dumped artifacts from test runs (see dump_artifacts.py)
 	rm -Rf test_artifacts
 
+.PHONY: run-registry
+run-registry:
```
christophermaier, line 422: Could use a descriptive comment here.

Makefile
```diff
@@ -412,13 +418,19 @@ clean-mount-cache: ## Prune all docker mount cache (used by sccache)
...
 .PHONY: push
 push: ## Push Grapl containers to Docker Hub
 	docker-compose --file=docker-compose.build.yml push
 
 .PHONY: push-local
 push-local: ## Push Grapl container to local registry
-	REGISTRY=localhost:5000 docker-compose --file=docker-compose.build.yml push
+	$(MAKE) run-registry
+	$(WITH_LOCAL_GRAPL_ENV) \
+	$(MAKE) push
```
christophermaier, line 433: This would better be expressed as:

```make
push-local: run-registry
    $(WITH_LOCAL_GRAPL_ENV) $(MAKE) push
```

Makefile
```diff
@@ -412,13 +418,19 @@ clean-mount-cache: ## Prune all docker mount cache (used by sccache)
 clean-artifacts: ## Remove all dumped artifacts from test runs (see dump_artifacts.py)
 	rm -Rf test_artifacts
 
+.PHONY: run-registry
+run-registry: 
+	nomad/local/local_grapl_registry.sh
+
 .PHONY: push
 push: ## Push Grapl containers to Docker Hub
```
christophermaier, line 426: Worth updating this comment; we're not pushing anything to Docker Hub.

nomad/local/local_grapl_registry.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -euo pipefail
+
+GRAPL_LOCAL_REGISTRY_PORT=5000
+GRAPL_LOCAL_REGISTRY_NAME="grapl_local_registry"
+
+if [[ ! -z "$(docker ps --quiet --filter name="${GRAPL_LOCAL_REGISTRY_NAME}")" ]]; then
```
christophermaier: It would be simpler to use `-n` ("string is not null") than `! -z` (NOT "string is null")

nomad/local/local_grapl_registry.sh
```diff
@@ -0,0 +1,19 @@
...
+    echo "already running."
+elif [[ ! -z "$(docker container ls --all --quiet --filter name="${GRAPL_LOCAL_REGISTRY_NAME}")" ]]; then
+    echo "Container exists but stopped. Restarting.."
+    docker start "${GRAPL_LOCAL_REGISTRY_NAME}"
+else
+    echo "Creating container registry."
+    docker run -d \
```
christophermaier, line 15: Instead of `-d`, use the long name `--detach`.

nomad/local/local_grapl_registry.sh
```diff
@@ -0,0 +1,19 @@
...
+set -euo pipefail
+
+GRAPL_LOCAL_REGISTRY_PORT=5000
+GRAPL_LOCAL_REGISTRY_NAME="grapl_local_registry"
+
+if [[ ! -z "$(docker ps --quiet --filter name="${GRAPL_LOCAL_REGISTRY_NAME}")" ]]; then
+    echo "already running."
+elif [[ ! -z "$(docker container ls --all --quiet --filter name="${GRAPL_LOCAL_REGISTRY_NAME}")" ]]; then
+    echo "Container exists but stopped. Restarting.."
+    docker start "${GRAPL_LOCAL_REGISTRY_NAME}"
```
christophermaier, line 12: How does the container get stopped?

Would it be useful to add the `--rm` flag in the `docker run` command below?

docker-compose.build.yml
```diff
@@ -120,13 +120,13 @@ services:
...
   pulumi:
-    image: ${REGISTRY:-grapl}/local-pulumi:${TAG:-latest}
+    image: ${DOCKER_REGISTRY:-grapl}/local-pulumi:${TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.pulumi
 
   provisioner:
-    image: ${REGISTRY:-grapl}/provisioner:${TAG:-latest}
+    image: ${DOCKER_REGISTRY:-grapl}/provisioner:${TAG:-latest}
```
christophermaier, line 129: What is the benefit of including a default value in all these (the "grapl" bit)? It seems that since the `localhost:5000` value is now in `local-grapl.env` that covers the Local Grapl case. Since we use Cloudsmith for real registry operations, we will always have to set `DOCKER_REGISTRY` accordingly when we're doing stuff in CI. What would the usecase be for defaulting to "grapl" (and thus Docker Hub, which we aren't using)?

Makefile
d0nut-grapl: > Since build-services already uses WITH_LOCAL_GRAPL_ENV, we don't gain anything by doing it here, too.

I don't think that's true. If I run `make build-services` the docker build does not see the `DOCKER_REGISTRY` environment variable.

Makefile
d0nut-grapl: Would this not push to Docker Hub by default?

nomad/local/local_grapl_registry.sh
d0nut-grapl: This should be corrected now, actually. I made this change as a result of the formatter and pushed the change up before standup.

docker-compose.build.yml
d0nut-grapl: I didn't want to break anything, basically. it used to say `grapl` and as I was changing it, I wanted to avoid breaking anything while we're in a transition phase.

nomad/local/local_grapl_registry.sh
d0nut-grapl: Gotcha; we could do that.

nomad/local/local_grapl_registry.sh
christophermaier: cool 👍

Makefile
christophermaier: If another registry isn't part of the container names, yes. However, we aren't doing anything with Docker Hub, so it's misleading.

docker-compose.build.yml
christophermaier: I think it should be safe, but we can also loop back around later if you like 👍

Makefile
d0nut-grapl: I'm not sure what change I should make here then. If this isn't going to push to docker hub, it sounds like it should either:

* be removed
* updated to say "pushed to supplied DOCKER_REGISTRY"

Which makes the most sense?

christophermaier: I think Option 2 sounds good for now... whether or not Option 1 should be eventually adopted depends on some stuff that @inickles-grapl and I are tossing around.

docker-compose.build.yml
```diff
@@ -12,7 +12,7 @@ services:
   ########################################################################
 
   sysmon-generator:
-    image: ${DOCKER_REGISTRY:-registry-1.docker.io}/grapl/sysmon-generator:${TAG:-latest}
+    image: ${DOCKER_REGISTRY:-docker.io}/grapl/sysmon-generator:${TAG:-latest}
```
christophermaier, line 15: Not to flog a dead horse, but this could be simplified if we require that `$DOCKER_REGISTRY` ends in a `/`.

Then you could just use `${DOCKER_REGISTRY:-}grapl/sysmon-generator:${TAG:-latest}`... if you supply a registry, it gets added, but if not, you still have a valid identifier.

Not required or anything, though; just something that occurred to me just now.

nomad/grapl-core.nomad
```diff
@@ -0,0 +1,451 @@
...
+      env {
+        RUST_LOG           = "${var.rust_log}"
+        REDIS_ENDPOINT     = "${var.redis_endpoint}"
+        MG_ALPHAS          = "${local.alpha_grpc_connect_str}"
+        GRAPL_SCHEMA_TABLE = "${var.schema_table_name}"
+        AWS_REGION         = "${var.aws_region}"
+        # https://github.com/grapl-security/grapl/blob/18b229e824fae99fa2d600750dd3b17387611ef4/pulumi/grapl/__main__.py#L165
+        DEST_BUCKET_NAME      = "subgraphs-merged-bucket"
+        SOURCE_QUEUE_URL      = "${var.aws_sqs_url}/${var.aws_account_id}/graph-merger-queue"
+        DEAD_LETTER_QUEUE_URL = "${var.aws_sqs_url}/${var.aws_account_id}/graph-merger-dead-letter-queue"
```
christophermaier, line 372: All the S3 bucket names, SQS queue URLs, and DynamoDB table names should all be passed in as variables, because of:

- https://github.com/grapl-security/grapl/pull/1072
- https://github.com/grapl-security/grapl/pull/1075
- https://github.com/grapl-security/grapl/pull/1076

The values present here right now wouldn't have worked anyway, because we used to prefix each one with the "deployment name". By just passing the full values, all the naming logic remains centralized in our infrastructure code, and all our services and deployment code (e.g., Docker Compose and Nomad files) can be a little bit simpler.

You won't need an `aws_sqs_url` or `aws_account_id` variable, though.

This applies to all other services in this file as well.

(I'd also recommend double-checking that these environments line up with what's in our Pulumi code for these services.)

nomad/local/start_development_environment.sh
```diff
@@ -0,0 +1,24 @@
...
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root"
+    exit 1
+fi
+
+trap 'kill $(jobs -p)' EXIT
+
+echo "Starting nomad and consul locally."
+nomad agent -dev-connect > /dev/null 2>&1 &
+consul agent -dev > /dev/null 2>&1 &
```
christophermaier, line 11: I don't think swallowing all Nomad and Consul output is the way to go 😅

nomad/local/start_development_environment.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root"
+    exit 1
+fi
```
christophermaier, line 5: It may be useful to put a comment here that this guard is strictly informative; `nomad agent -dev-connect` _can't_ run as a non-root user.

(That is to say, some future engineer working on this script isn't going to horribly break things if they happen to remove this check. Also, Nomad outputs an error message saying it has to run as root anyway.)

christophermaier (approved):

PR #1066, opened 2021-07-30
Rearrange and clarify Make targets, and use docker run to extract from images

inickles-grapl:

NA

This does a number a few things for clarifying and simplifying the build process:
- Consolidate the Docker image build steps that were in docker-compose.lambda-zips.js.yml and docker-compose.lambda-zips.rust.yml to docker-compose.build.yml, where the other image build specs are.
- Use `docker run` instead of docker-compose to create the containers used to pull AWS Lambda zips out of previously built Docker images.
- Delete docker-compose.lambda-zips.js.yml and docker-compose.lambda-zips.rust.yml as they are no longer needed.
- Rename the `lambdas` Make targets we had to names that speak more to what they're doing. `lambdas-python` -> `build-python-lambda-zips`, `lambdas-rust` -> `extract-lambda-zips-rust`, ...
- Clean up Make target dependencies accordingly

Using the test suite we have.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ TAG ?= latest
 RUST_BUILD ?= debug
 UID = $(shell id -u)
 GID = $(shell id -g)
+PWD = $(shell pwd)
 DOCKER_BUILDX_BAKE_OPTS ?=
 ifneq ($(GRAPL_RUST_ENV_FILE),)
 DOCKER_BUILDX_BAKE_OPTS += --set *.secrets=id=rust_env,src="$(GRAPL_RUST_ENV_FILE)"
@@ -17,16 +18,12 @@ COMPOSE_IGNORE_ORPHANS=1
 COMPOSE_PROJECT_NAME ?= grapl
 export
 
-export EVERY_LAMBDA_COMPOSE_FILE=--file docker-compose.lambda-zips.js.yml \
-	--file docker-compose.lambda-zips.rust.yml
-
 export EVERY_COMPOSE_FILE=--file docker-compose.yml \
 	--file ./test/docker-compose.unit-tests-rust.yml \
 	--file ./test/docker-compose.unit-tests-js.yml \
 	--file ./test/docker-compose.integration-tests.yml \
 	--file ./test/docker-compose.e2e-tests.yml \
-	--file ./test/docker-compose.typecheck-tests.yml \
-	${EVERY_LAMBDA_COMPOSE_FILE}
+	--file ./test/docker-compose.typecheck-tests.yml
 
 DOCKER_BUILDX_BAKE := docker buildx bake $(DOCKER_BUILDX_BAKE_OPTS)
 
@@ -143,22 +140,56 @@ build-test-typecheck: build-python-wheels
 		--file ./test/docker-compose.typecheck-tests.yml
 
 .PHONY: build-test-integration
-build-test-integration: build-services
+build-test-integration: build
 	$(WITH_LOCAL_GRAPL_ENV) \
 	$(DOCKER_BUILDX_BAKE) --file ./test/docker-compose.integration-tests.yml
 
 .PHONY: build-test-e2e
-build-test-e2e: build-services
+build-test-e2e: build
 	$(WITH_LOCAL_GRAPL_ENV) \
 	$(DOCKER_BUILDX_BAKE) --file ./test/docker-compose.e2e-tests.yml
+
+.PHONY: build-lambda-zips
+build-lambda-zips: build-lambda-zips-rust build-lambda-zips-js build-lambda-zips-python ## Generate all lambda zip files
+
+.PHONY: build-lambda-zips-rust
+build-lambda-zips-rust: export COMPOSE_FILE := docker-compose.lambda-zips.rust.yml
+build-lambda-zips-rust: ## Build Rust lambda zips
+	$(DOCKER_BUILDX_BAKE) --file "${COMPOSE_FILE}"
+	# Extract the zip from the Docker image.
+	# Rely on the default CMD for copying artifact to /dist mount point.
+	docker-compose run \
+		--rm \
+		--user "${UID}:${GID}" \
+		--volume="${PWD}/dist":/dist \
+		metric-forwarder-zip
+
+.PHONY: build-lambda-zips-js
+build-lambda-zips-js: export COMPOSE_FILE := docker-compose.lambda-zips.js.yml
+build-lambda-zips-js: ## Build JS lambda zips
+	$(DOCKER_BUILDX_BAKE) --file "${COMPOSE_FILE}"
+	# Extract the zip from the Docker image.
+	# Rely on the default CMD for copying artifact to /dist mount point.
+	docker-compose run \
+		--rm \
+		--user "${UID}:${GID}" \
+		--volume="${PWD}/dist":/dist \
+		graphql-endpoint-zip
+
+.PHONY: build-lambda-zips-python
+build-lambda-zips-python: build-python-wheels ## Build Python lambda zips
+	./pants filter --target-type=python_awslambda :: | xargs ./pants package
 
 .PHONY: build-python-wheels
 build-python-wheels:  ## Build all Python wheels
 	./pants filter --target-type=python_distribution :: | xargs ./pants package
 
-.PHONY: build-services
-build-services: graplctl lambdas build-python-wheels ## Build Grapl services
+.PHONY: build-docker-images
+build-docker-images: graplctl
 	$(DOCKER_BUILDX_BAKE) --file docker-compose.build.yml
+
+.PHONY: build
+build: build-lambda-zips build-docker-images ## Build Grapl services
 
 .PHONY: build-formatter
 build-formatter:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -156,10 +153,23 @@ build-test-e2e: build-services
...
 	./pants filter --target-type=python_distribution :: | xargs ./pants package
 
-.PHONY: build-services
-build-services: graplctl lambdas build-python-wheels ## Build Grapl services
+.PHONY: build-python-lambda-zips
+build-python-lambda-zips:
+	./pants filter --target-type=python_awslambda :: | xargs ./pants package
+
+.PHONY: build-python-pants
+build-python-pants: build-python-wheels build-python-lambda-zips
```
inickles-grapl, line 161: Maybe `build-python-with-pants` is better?

inickles-grapl: I'm wondering if we should just include graplctl build into this target.

christophermaier: 🤔 Does it need to include "pants" at all anymore?

It's somewhat like having a `build-rust-with-cargo`, `compile-c-with-gcc`, or `build-jar-with-maven` targets.

Makefile
```diff
@@ -194,22 +204,26 @@ ux-tarball: build-ux dist ## Build website asset tarball
...
-lambdas-rust: ## Build Rust lambda zips
-	$(DOCKER_BUILDX_BAKE) -f docker-compose.lambda-zips.rust.yml
-	docker-compose -f docker-compose.lambda-zips.rust.yml up
-
-.PHONY: lambdas-js
-lambdas-js: ## Build JS lambda zips
-	$(DOCKER_BUILDX_BAKE) -f docker-compose.lambda-zips.js.yml
-	docker-compose -f docker-compose.lambda-zips.js.yml up
-
-.PHONY: lambdas-python
```
christophermaier, line 210: The merge pipeline uses these targets and files; you'll have to update it accordingly.

Makefile
```diff
@@ -194,22 +204,26 @@ ux-tarball: build-ux dist ## Build website asset tarball
...
+		sh -c 'cp /grapl/zips/metric-forwarder.zip /tmp/zips/metric-forwarder-lambda.zip'
+
+.PHONY: extract-lambda-zips-js
+extract-lambda-zips-js: build-docker-services
+	docker run --rm \
+		--user "${UID}:${GID}" \
+		--volume "${PWD}/dist:/grapl" \
+		--workdir /grapl \
+		grapl/graphql-endpoint-zip:${TAG} \
+		cp /home/grapl/lambda.zip graphql-endpoint-lambda.zip
```
christophermaier, line 226: It would be more clear to have the mountpoint map more obviously to its location on the workstation. That is, don't call it `zips` in the container, when it's `dist` outside the container. This just makes people reading the code have to do unnecessary mental gymnastics. If that calls for structuring the filesystem in the container in a different way (i.e., create a `dist` container, rather than `zips` in `/grapl`), then so be it.

Also, is there a particular reason that the `cp` commands, their invocations, and the workdirs are so different between these two `docker run` commands? To the extent possible, we should make them as identical as we can (it makes it easier to reason about, and opens the door to future abstraction and automation around this). If that requires overriding entrypoints because these have different base containers, or anything else, we should do that.

docker-compose.lambda-zips.js.yml
```diff
@@ -1,14 +0,0 @@
-version: "3.8"
```
christophermaier, line 1: Changes to this file are used to determine if we need to build the JS lambda in our merge pipeline. I would hate to lose this ability; I don't want to be building new versions of services if we don't need to. This takes up time in the pipeline, space in artifact repositories, and leads to unnecessary churn when deploying services to production.

As it is, we do this with our Rust services because we don't yet have a good way to determine which of them may (or may not) have changed in a given PR. I have some notes on how to do this, though; once that's in place we would only build what we needed to for any given run through the pipeline.

Makefile
```diff
@@ -194,22 +204,26 @@ ux-tarball: build-ux dist ## Build website asset tarball
...
+extract-lambda-zips-rust: build-docker-services
+	docker run --rm \
+		--user "${UID}:${GID}" \
+		--volume "${PWD}/dist:/tmp/zips" \
+		--workdir /grapl/zips \
+		grapl/metric-forwarder-zip:${TAG} \
+		sh -c 'cp /grapl/zips/metric-forwarder.zip /tmp/zips/metric-forwarder-lambda.zip'
+
+.PHONY: extract-lambda-zips-js
+extract-lambda-zips-js: build-docker-services
```
christophermaier, line 220: An unfortunate side effect of this dependency structure is that if (for whatever reason) you only wanted to build this single Javascript lambda, you will necessarily end up building every other container, along with all our Python artifacts.

Similar logic holds for the Rust lambdas.

Makefile
inickles-grapl: > It would be more clear to have the mountpoint map more obviously to its location on the workstation. That is, don't call it `zips` in the container, when it's `dist` outside the container. This just makes people reading the code have to do unnecessary mental gymnastics. If that calls for structuring the filesystem in the container in a different way (i.e., create a `dist` container, rather than `zips` in `/grapl`), then so be it.

Yeah, that seems fair. I'd only ported what was defined in the compose files to docker args, so I wouldn't need to also change their respective Dockerfiles, but I agree it's better to use a common mount point.

> Also, is there a particular reason that the `cp` commands, their invocations, and the workdirs are so different between these two `docker run` commands? To the extent possible, we should make them as identical as we can (it makes it easier to reason about, and opens the door to future abstraction and automation around this). If that requires overriding entrypoints because these have different base containers, or anything else, we should do that.

The paths to copy _from_ are defined in the respective Dockerfiles, which I wasn't planning on changing in this PR. I was looking to only relieve our use of Docker Compose. I could include Dockerfile changes as well so source paths match better as well, if that's preferred.

Makefile
inickles-grapl: Yeah, this was the idea for having these separate targets in the first place, but I know there have been complaints about complexity with this system and this seemed like a decent opportunity to simplify all the things. I wasn't aware of this being used, but sounds like it's potentially useful for, as you said in another comment, only building when the sources have changed.

Makefile
christophermaier: Yeah, I think this is a valid opportunity to make this change.

Makefile
inickles-grapl: I think including "pants" is helpful because it looks like we're building sources with both Pants and Docker, and I think this helps clarify that. The new name I made for Docker could better match this Pants one, how about we keep `build-with-pants` and `build-docker-services` becomes `build-with-docker`?

christophermaier: Sounds good.

I look forward to the day when we can hopefully just have a `build` target 😂

docker-compose.lambda-zips.js.yml
inickles-grapl: In that case it seems the Make build targets should be broken up and rearranged different to include thing like a `build-rust` and `build-graphql-endpoint` (as well as others).

It seems the merge pipeline has different upload steps based on type of artifact, but it seems it'd be more appropriate for that to match the build targets, so we'd have one for Rust and one for graplql-endpoint.

Makefile
inickles-grapl: Maybe today is that day.

Makefile
```diff
@@ -17,16 +18,12 @@ COMPOSE_IGNORE_ORPHANS=1
 COMPOSE_PROJECT_NAME ?= grapl
 export
 
-export EVERY_LAMBDA_COMPOSE_FILE=--file docker-compose.lambda-zips.js.yml \
-	--file docker-compose.lambda-zips.rust.yml
-
```
inickles-grapl, line 22: This was used for the heavy-handed make down target, but it's no longer needed because the containers now clean up after themselves. Example below in line 161: https://github.com/grapl-security/grapl/pull/1066/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R161

Makefile
```diff
@@ -143,23 +140,55 @@ build-test-typecheck: build-python-wheels
...
+.PHONY: build-lambda-zips-js
+build-lambda-zips-js: ## Build JS lambda zips
+	$(DOCKER_BUILDX_BAKE) --file docker-compose.lambda-zips.js.yml
+	# Extract the zip from the Docker image.
+	# Rely on the default CMD for copying artifact to /dist mount point.
+	docker run \
+		--rm \
+		--user "${UID}:${GID}" \
+		--mount type=bind,source="${PWD}/dist",target=/dist \
+		grapl/graphql-endpoint-zip:${TAG}
```
inickles-grapl, line 175: We could use `docker-compose run` here instead, which IMO has a benefit of anchoring on the image name defined in the compose file, so if the image name changes in that file you wouldn't have to update it here as well. I chose `docker run` just to avoid compose altogether, but I think we could try using `docker-compose run` and see if it gives us issues. It's still better than using compose up, which was previously the case.

docker-compose.lambda-zips.js.yml
```diff
@@ -4,11 +4,6 @@ services:
   graphql-endpoint-zip:
     image: grapl/graphql-endpoint-zip:${TAG:-latest}
     build:
-      context: src
-      dockerfile: js/graphql_endpoint/Dockerfile
+      context: src/js/graphql_endpoint
+      dockerfile: Dockerfile
```
christophermaier, line 8: Wow, good catch.

Makefile
christophermaier: I guess I'm agnostic on this point. It's definitely better than `docker-compose up`. Having the image name "take care of" in the compose file _is_ nice.

On the other hand, moving away from Docker Compose for things that aren't actually services is _also_ nice. This could potentially open the door to moving our build-only compose files to HCL for `buildx bake`, which solidifies that distinction. However, that's just speculation at this point, and wouldn't be hard to implement either way.

I'm fine with `docker run` 😄

christophermaier (approved):

Makefile
inickles-grapl: Yeah.. I'm tempted to switch back to 'docker-compose run' for now, not just for handing the image name, but also because 'docker-compose run' is used elsewhere in the Makefile, so it'd also make Makefile more consistent.

I think I'll make the switch, since you seem fine either way, though I'll ping you for confirmation before merging to main. I see this has been accepted already (ty ty).

inickles-grapl: (such as `lint-prettier`)

PR #1067, opened 2021-07-30
Use Docker multi-stage builds for building our Python Lambdas

inickles-grapl:

We'd previously switched to using Docker multi-stage builds[1] but the builder pattern crept back in for building Python Lambdas images. This PR converts that builder pattern to Docker multi-stage builds and reduces the number of moving parts in our build system.

[1] https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#use-multi-stage-builds

The builder pattern can be problematic in that it's easy for issues like https://github.com/grapl-security/issue-tracker/issues/593 to creep in. The worst case scenario for this kind of issue is when a build with a non-default TAG occurs _after_ a build with the default TAG. The non-default build will silently COPY files from the wrong Docker image. Multi-stage builds mitigate this kind of issue.

Using our tests.


src/python/lambdas.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    --home /home/grapl \
+    --shell /bin/bash \
+    grapl
+
+USER grapl
+ENV USER=grapl
+WORKDIR /home/grapl
+RUN mkdir -p /home/grapl/bin
+RUN mkdir -p /home/grapl/.aws
+RUN echo "[default]\naws_access_key_id=fake\naws_secret_access_key=fake" > /home/grapl/.aws/credentials
```
christophermaier, line 28: 🤔 I know this wasn't introduced in this PR, but why are these being baked into the image, rather than being passed in as environment variables like we do with every other container in our test environments?

(cc: @jgrillo-grapl)

christophermaier (approved):
I like that this is consolidating things, so 👍 

Have we confirmed that these lambdas cannot run _as_ lambdas under Localstack? If so, that would be good information to add as a documentation comment in the Dockerfile or compose YAML (probably both?). If not, it would be a good idea to try running these in Localstack, just to remove as much variance as we can.

(I suspect @jgrillo-grapl knows the answer to this.)

src/python/lambdas.Dockerfile
jgrillo-grapl: I believe in my testing I found that if the file wasn't present on disk then the boto3 library threw an error, but it was a long time ago so I don't really remember.

jgrillo-grapl:
> I like that this is consolidating things, so 👍
> 
> Have we confirmed that these lambdas cannot run _as_ lambdas under Localstack? If so, that would be good information to add as a documentation comment in the Dockerfile or compose YAML (probably both?). If not, it would be a good idea to try running these in Localstack, just to remove as much variance as we can.
> 
> (I suspect @jgrillo-grapl knows the answer to this.)

The only Lambda I know of that _won't_ run in Localstack is the end-to-end test runner. However, since we don't know _why_ it won't run, I think it's probably best to avoid running any lambdas in localstack. Seems like a ticking time bomb.

PR #1070, opened 2021-07-30
Various Engagement View Dockerfile improvements

inickles-grapl:

NA

This makes a number of changes to the Engagement View Dockerfile, most notably:
- Use cache mount for package installation instead of relying on image fs layer.
- Change the base image from `node:alpine3.10` to `node:16-buster-slim`, which will put the base image more inline with other Dockerfiles we have and will now explicitly set the node version to use. The `node:alpine3.10` image is no longer in the list of supported images: https://hub.docker.com/_/node.

Using CI tests.

### DIFF
--- a/src/js/engagement_view/Dockerfile
+++ b/src/js/engagement_view/Dockerfile
@@ -22,8 +22,6 @@ FROM engagement-view-deps AS engagement-view-build
 RUN yarn add react-scripts
 RUN yarn build
 
-# no-op the base image cmd, so it doesn't launch a Node repl
-CMD :
 
 # create an image that we can deploy from locally
 ################################################################################
--- a/src/js/engagement_view/Dockerfile
+++ b/src/js/engagement_view/Dockerfile
@@ -19,7 +19,6 @@ COPY js/engagement_view .
 FROM engagement-view-deps AS engagement-view-build
 
 # build sources
-RUN yarn add react-scripts
 RUN yarn build
 
 
--- a/src/js/engagement_view/Dockerfile
+++ b/src/js/engagement_view/Dockerfile
@@ -1,6 +1,5 @@
 FROM node:alpine3.10 AS engagement-view-deps
 
-RUN apk add bash yarn
 WORKDIR /grapl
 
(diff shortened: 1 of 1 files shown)


christophermaier (approved):
I would love to remove this altogether, as it's only in support of uploading the view assets in Local Grapl.

However, these cleanups are good by themselves, so 👍

inickles-grapl:
> I would love to remove this altogether, as it's only in support of uploading the view assets in Local Grapl.
> 
> However, these cleanups are good by themselves, so 👍

I notice we have the `build-ux` Make target, which does similar outside of Docker. It does seem like we should converge on either of the two for building. That is if we're to do `yarn build` directly via Make, then the deploy image should just COPY from that. Thoughts?

christophermaier:
> I notice we have the `build-ux` Make target, which does similar outside of Docker. It does seem like we should converge on either of the two for building. That is if we're to do `yarn build` directly via Make, then the deploy image should just COPY from that. Thoughts?

Probably so, yeah. Centralizing on one operational flow makes sense.

inickles-grapl:
> Probably so, yeah. Centralizing on one operational flow makes sense.

The only concern I have with building without Docker is that we don't seem to have controls to ensure a consistent version of Node is installed/used to build the sources. That's one of the nice things about doing this in Docker (not to say there aren't reasons to do it outside of Docker). So I'm not sure what the best way centralize is at this point, otherwise I'd address it in this PR. Lemme know if you think it's worth filing a ticket for to address later.

christophermaier:
@inickles-grapl FWIW, we _do_ lock down the node version when we build our artifacts in CI:
https://github.com/grapl-security/grapl/blob/7f7be585b5b83c2c6fbebef728188feeff9e55df/.buildkite/pipeline.merge.ux-build.yml#L6-L12

My only concern with whatever we end up doing is that we don't end up doing something that complicates the local developer experience (including how developers interact with code through their editors and IDEs). I think this is worth an issue.

inickles-grapl:
I'm going to drop a51e237 because it breaks tests.

PR #1071, opened 2021-07-31
Removed broken link for Setup

ebcarty:
Added links to run grapl on your computer and to deploy on AWS under Setup.

### Which issue does this PR correspond to?

Updated Setup section under ReadMe to link to running locally and deploying to AWS.

### What changes does this PR make to Grapl? Why?

Good to have working links, especially to learn about setting up Grapl.

### How were these changes tested?

I clicked on the links and verified that they work.

PR #1074, opened 2021-08-02
Make clearer in our documentation that Local Grapl is not supported

wimax-grapl:
None, but it became clear due to https://github.com/grapl-security/grapl/pull/1071 that we need to make this more obvious!

- Remove a big chunk of CONTRIBUTING.md that's kind of useless
- Direct all of our setup documentation to docs/setup on RtD
- Write a short blurb in docs/setup/index.rst that we don't support local right now.

make build-docs

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,18 +74,11 @@ To run your images locally, execute the following command in the
 project root (after building):
 
 ``` bash
-TAG=latest docker-compose up
+make up
 ```
 
-Alerternatively, you can use `make up` to rebuild the sources and launch `docker-compoes up`.
-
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -74,18 +74,11 @@ To run your images locally, execute the following command in the
 project root (after building):
 
 ``` bash
-TAG=latest docker-compose up
+make up
 ```
```
wimax-grapl, line 78: yes - this is a very minimal reference to local grapl - it's in `contributing.md` and i think it's _okayish_ to let contributors know they can bring up an environment. Like, consider our new employee Thomas - how would he know about `make up` if not from tribal knowledge?

colin-grapl (approved):

PR #1075, opened 2021-08-03
No more explicit table naming

christophermaier:
This is the same idea behind https://github.com/grapl-security/grapl/pull/1072, but applied to DynamoDB tables.

Of particular interest is the deletion of nearly all our "history" tables, which were unused.


src/python/graplctl/graplctl/idempotency_checks.py
```diff
@@ -1,19 +1,18 @@
...
 
 
 def is_grapl_provisioned(
     dynamodb: DynamoDBServiceResource,
-    deployment_name: str,
+    schema_table: str,
 ) -> bool:
     """
-    We are doing a very simple check - "is there >0 things in schema_properties_table?"
+    We are doing a very simple check - "is there >0 things in schema_table?"
```
wimax-grapl, line 12: good catch :)

wimax-grapl (approved):

PR #1079, opened 2021-08-04
Format TOML with prettier. Split up `make format-python` into also format-shell.

wimax-grapl:
No issue. Tested locally with a `make format` and `make lint`. 

I don't *love* the new formatting of the tomls if I'm gonna be honest with you, but it's fine (and semi useful for another PR I have in-flight, which potentially mutates pants.toml)


Makefile
```diff
@@ -335,7 +335,11 @@ format-rust: ## Reformat all Rust code
...
 format-python: ## Reformat all Python code
-	./pants fmt ::
+	./pants filter --target-type=python_library,python_tests :: | xargs ./pants fmt
+
+.PHONY: format-shell
+format-shell: ## Reformat all shell_libraries
+	./pants filter --target-type=shell_library,shunit2_tests :: | xargs ./pants fmt
 
 .PHONY: format-prettier
 format-prettier: build-formatter ## Reformat js/ts/yaml
```
christophermaier, line 345: Don't forget to add "toml" to the comment now (markdown, too, I suppose!)

Makefile
```diff
@@ -349,7 +353,7 @@ format-packer: ## Reformat all Packer HCLs
 	packer fmt -recursive packer/
 
 .PHONY: format
-format: format-python format-prettier format-rust format-packer ## Reformat all code
+format: format-python format-shell format-prettier format-rust format-packer ## Reformat all code
```
christophermaier, line 356: This is not necessarily something to fix in this PR, but more of an observation. We have all these different formatting targets for the various kinds of files in our repository (python, shell, rust, packer), and then we have `format-prettier`, which isn't a file type, but a tool, and it does the formatting for all the JS, YAML, Markdown, and TOML in one go, with no ability to target one specifically. This is a bit of an weird incongruity. It's also a bit of a downer that if I just want to make sure my TOML file is formatted properly, I have to sit through formatting all the JS, YAML, and Markdown files, just because the TOML formatting currently happens to be at the end of that script.

Makefile
christophermaier: It also occurs to me that there's actually no benefit of having a `build-formatter` target, since it's not a dependency of anything else.

We could just pull that build command into this target directly and simplify the Makefile just a tiny bit. `build-formatter` is a `PHONY` target, so it's always going to run anyway.

Actually, even better than that would just be to enable Buildkit builds in Docker Compose and get rid of the explicit build step altogether, and just have this be a `docker-compose run` invocation.

Per https://www.docker.com/blog/faster-builds-in-compose-thanks-to-buildkit-support/, we'll need to set the following environment variables: `COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1`. (I tested this locally; works great.)

(Thoughts on this, @inickles-grapl?)

christophermaier (approved):

Makefile
wimax-grapl: So on one hand I agree, but timing it - it's so quick that I'm not sure it's worth the added complexity of chopping it up.

real    0m13.621s
user    0m0.793s
sys     0m0.360s

Makefile
wimax-grapl: (currently, lint-prettier also depends on that build step. however, I whole-heartedly agree with your docker-compose suggestion, but I think it's best suited for a separate PR that applies it universally?)

christophermaier: `lint-prettier` would end up looking the same as I suggest here; `build-formatter` would just disappear completely.

Doing this in a separate PR is a good way to go, though.

inickles-grapl: > It also occurs to me that there's actually no benefit of having a build-formatter target, since it's not a dependency of anything else.
> 
> We could just pull that build command into this target directly and simplify the Makefile just a tiny bit. build-formatter is a PHONY target, so it's always going to run anyway.

Agreed, and I had a similar thought recently about some of the the `build-test-` targets we have as well.

> Actually, even better than that would just be to enable Buildkit builds in Docker Compose and get rid of the explicit build step altogether, and just have this be a docker-compose run invocation.
> 
> Per https://www.docker.com/blog/faster-builds-in-compose-thanks-to-buildkit-support/, we'll need to set the following environment variables: COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1. (I tested this locally; works great.)

Yeah, I think that's fine as long as the builds depend on a single Dockerfile of ours, and we're not using the 'builder pattern' as described from: https://docs.docker.com/develop/develop-images/multistage-build/.

PR #1083, opened 2021-08-04
Makefile: remove the 'dist' target as it shouldn't be needed because we

inickles-grapl:

This just removes the `dist` Makefile target. It looks like it's no longer useful because the directory should always exist with `dist/.gitkeep` in git.

It looks like this target isn't used, I'll rely on CI tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,16 +175,11 @@ graplctl: ## Build graplctl and install it to the project root
 build-ux: ## Build website assets
 	cd src/js/engagement_view && yarn install && yarn build
 
-# Create the dist directory if necessary; don't need to document this
-# formally for `make help`, though.
-dist:
-	mkdir dist
-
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -175,11 +175,6 @@ graplctl: ## Build graplctl and install it to the project root
 build-ux: ## Build website assets
 	cd src/js/engagement_view && yarn install && yarn build
 
-# Create the dist directory if necessary; don't need to document this
-# formally for `make help`, though.
-dist:
-	mkdir dist
-
```
christophermaier, line 182: If you're going to delete this, you should also delete the dependency of the `ux-tarball` target, too.

christophermaier (approved):

Makefile
inickles-grapl: Ah, yes. Good catch.

PR #1086, opened 2021-08-04
Simplify makefile

wimax-grapl:
Simplify makefile by removing some of the 'build-test-*' targets and using build-during-compose-up

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,9 @@ export EVERY_COMPOSE_FILE=--file docker-compose.yml \
 
 DOCKER_BUILDX_BAKE := docker buildx bake $(DOCKER_BUILDX_BAKE_OPTS)
 
+# https://www.docker.com/blog/faster-builds-in-compose-thanks-to-buildkit-support/
+DOCKER_COMPOSE_WITH_BUILDX := COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker-compose
+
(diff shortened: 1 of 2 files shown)


jgrillo-grapl (approved):
👏

christophermaier (approved):
Niiiiice

inickles-grapl (changes requested):
I think this will slow down our builds by a lot. Using docker-compose to build will build each target serially, one-by-one, where the bake command basically converges them to one build command and lets the buildkit engine handle the rest, like building in parallel.

It's not an issue for build files that have only one target, but it'll slow down builds with more than one, like `test/docker-compose.integration-tests.yml`.

wimax-grapl:
That's enough of a reason for me to abandon this.

inickles-grapl:
> That's enough of a reason for me to abandon this.

There is a `--parallel` option we could consider instead.

inickles-grapl:
> That's enough of a reason for me to abandon this.

Also, even without changing how we do parallel builds (bake vs compose build --parallel), it might be worth this kind of change for build files with a single target.

inickles-grapl:
> > That's enough of a reason for me to abandon this.
> 
> Also, even without changing how we do parallel builds (bake vs compose build --parallel), it might be worth this kind of change for build files with a single target.

I take that back a bit, even to use the `--parallel` flag, I think you'd still need to the explicit build command, which this PR removes.

inickles-grapl (approved):
👍

PR #1092, opened 2021-08-05
Fix up and better formalize our SQS queue usage

christophermaier:
Our "service queue" abstraction has the following structure:

    main queue --> retry queue --> dead-letter queue

Our services are split into two paired variants: the main service and the retry service.

The main service should pull from the main queue and write failed messages to the retry queue. On the other hand, the retry service should read from the _retry_ queue, and write failed messages to the _dead-letter_ queue.

In reality, our services in AWS were actually configured to be passed all three queues, and they all had read and write access to all three queues. In Local Grapl, the identities of many queues were incorrect (retry queues were passed in with an unused `RETRY_QUEUE_URL` environment variable, while dead-letter queues were _always_ passed in with `DEAD_LETTER_QUEUE_URL` variables).

Now, the proper queues are passed to the appropriate services. Additionally, IAM permissions have been restricted to the minimum needed; services can only read from their input queue and can only write to their output queue.

Additional cleanups and refactorings were also made, as appropriate.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -348,7 +348,7 @@ services:
       - |
         export GRAPL_ANALYZERS_BUCKET=$$(cat /mnt/pulumi-outputs/analyzers-bucket)
         export DEST_BUCKET_NAME=$$(cat /mnt/pulumi-outputs/dispatched-analyzer-bucket)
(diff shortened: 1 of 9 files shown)


wimax-grapl (approved):
great simplification

jgrillo-grapl (approved):

PR #1094, opened 2021-08-06
Fix port binding collision by changing GRAPL_GRAPHQL_PORT to 4000.

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/628

This changes GRAPL_GRAPHQL_PORT to 4000 to avoid port binding collision with the local container registry added by https://github.com/grapl-security/grapl/pull/1059.

In trying to decide which of these two values to change, and to what value, I tried looking for a default port, or at least one that's typically used by GraphQL. I found this, which uses 4000: https://graphql.org/graphql-js/running-an-express-graphql-server/. Since that's different than what we had, I took it.

I verified I am no longer able to reproduce the error. I ran:
1. `make build-docker-images-local`
2. `make test-e2e`


christophermaier (approved):
Why not 9000? 🙃

PR #1095, opened 2021-08-09
Generalize artifact handling code

christophermaier:
Rather than concern ourselves with the specific names of each JSON artifacts file we create in our merge pipeline, we generalize our handling of this information to be more decoupled and open to future expansion.

Now, rather than having a concrete name for each such file we create, and requiring the identity of this file to be known in the final artifact merge step, we create the names programmatically. When given a descriptive "slug" (say, "containers"), we generate a filename that includes the unique Buildkite job ID, along with a special file extension, `grapl-artifacts.json`. This yields artifact files with names like `containers-e44f9784-e20e-4b93-a21d-f41fd5869db9.grapl-artifacts.json`. The nice thing about this is that the incorporation of the job ID eliminates the possibility of artifact file naming conflicts across the entire pipeline; all you have to do is make sure the "slug" provided is unique within a given job (a much simpler task).

We've also tightened up the naming; now _all_ artifacts files have the `grapl-artifacts.json` (previously, some ended in `_artifacts.json`, while others were `.artifacts.json`. By unifying behind a very company-specific extension, we can take advantage of globbing when uploading and downloading files without fear of accidentally pulling in other files. We also don't need to bother isolating the files in a particular directory, either, which simplifies the code further. 

Another simplification is the removal of the "manual" file upload steps _within_ a job's code, in favor of using the `artifact_paths` configuration for the job itself. Not only does this remove a lot of "glue" code, it also makes the job code more general by reducing its coupling with Buildkite itself. This in turn opens the way to more reuse and easier testing. It also helps to make the overall information flow of the pipeline more obvious when looking at the corresponding pipeline YAML file.

Overall, this PR reduces coupling between jobs, reduces coupling to Buildkite itself, reduces custom code, makes the flow of logic more obvious, and makes it easier to introduce or modify additional artifact-creating steps.

### DIFF
--- a/.buildkite/pipeline.merge.ami-build.yml
+++ b/.buildkite/pipeline.merge.ami-build.yml
@@ -61,3 +61,5 @@ steps:
   - label: ":writing_hand: Record AMI IDs"
     command:
       - .buildkite/scripts/record_artifacts.sh
+    artifact_paths:
+      - "artifact_manifests/*.json"
--- a/.buildkite/pipeline.merge.ami-build.yml
+++ b/.buildkite/pipeline.merge.ami-build.yml
@@ -62,4 +62,4 @@ steps:
     command:
       - .buildkite/scripts/record_artifacts.sh
     artifact_paths:
-      - "artifact_manifests/*.grapl-artifacts.json"
+      - "*.grapl-artifacts.json"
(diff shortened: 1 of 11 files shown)


jgrillo-grapl (approved):
👏

PR #1103, opened 2021-08-10
Integration test for the Grapl Template Generator

wimax-grapl:
None.

Add a test that:
- generates a grpc service
- uses `make built-test-unit-rust` to make sure that both the library and the tests of it will *compile* - though not that it'll necessarily *run* (it doesn't yet)

Ran successfully locally.


Makefile
```diff
@@ -240,7 +240,7 @@ ux-tarball: build-ux ## Build website asset tarball
 ##@ Test 🧪
 
 .PHONY: test
-test: test-unit test-integration test-e2e test-typecheck ## Run all tests
+test: test-unit test-integration test-e2e test-typecheck test-grapl-template-generator ## Run all tests
```
wimax-grapl, line 243: tbh I think we should just remove this target; I don't think *anybody* uses it.

jgrillo-grapl: I used it by accident today, it was a mistake. I'd be in favor of removing it.

jgrillo-grapl (approved):

PR #1105, opened 2021-08-10
Add first policies to pulumi

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/629

This adds:
1. Awsguard, pulumi's base set of policies for AWS. These are set in advisory/warning mode
2. Our first set of internal policies:
    a. Prevent public s3 bucket acls (default pulumi policy)
    b. Prevent IAM policies containing s3:ListBuckets
3. A README giving an overview of policies

These are our first set of pulumi policies as code. Policies will allow us to prevent some known bad configurations such as unencrypted resources or wildcard policies, etc. This basically adds a bit of test coverage for known bad cases, similar to setting semgrep rules in place (but executed post create/on update). These won't matter as much now in the early stages but will be more important as we more people contribute. Importantly, this provides a manner to add security guiderails for infrastructure

Functionally this should give warnings for awsguard default policies, and block s3 buckets with public acls and IAM policies with s3:ListBuckets

Not tested yet (waiting on corp laptop + AWS access)

To test locally:
cd pulumi/grapl
pulumi preview --policy-pack ../grapl
pulumi preview --policy-pack ../3rdparty/awsguard

Once these are tested and confirmed they can be published to pulumi and enabled server-side there:
pulumi policy publish <org-name>
pulumi policy enable <org-name>/<policy-pack-name> <latest|version>


pulumi/policies/README.md
```diff
@@ -0,0 +1,32 @@
...
+- for running a policy against a single resource type
+- Checked **before** resources are created/modified  
+- ex: checking that s3 buckets have a specific acl set
+
+### Stack validation 
+- for running a policy against 2+ types of resources at the same time OR to validate against a resource that must already be created to validate (like ACM certificates)
+- Checked **after** resources are created (it's essentially a post-creation audit) 
+- ex: Enforcing a tagging policy
+- ex: checking that all fargate instances have an autoscaling policy set. 
+- ex:  They can also be
```
jgrillo-grapl, line 18: fragment

pulumi/policies/grapl/PulumiPolicy.yaml
```diff
@@ -0,0 +1,6 @@
+runtime:
+  name: python
+  options:
+    virtualenv: venv
```
jgrillo-grapl, line 4: this should be a pants project, and therefore will not use a pulumi-supplied virtualenv

pulumi/policies/grapl/__main__.py
```diff
@@ -0,0 +1,47 @@
+from pulumi_policy import (
+    EnforcementLevel,
+    PolicyPack,
+    ReportViolation,
+    ResourceValidationArgs,
+    ResourceValidationPolicy,
+)
+
+
+def s3_no_public_read_validator(args: ResourceValidationArgs, report_violation: ReportViolation):
```
jgrillo-grapl, line 10: This looks like it probably doesn't typecheck (no return type specified). I assume this is due to a missing `BUILD` file (i.e. `pants` is unaware that this code exists, so it doesn't attempt to build or typecheck it).

pulumi/policies/grapl/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pulumi-policy>=1.3.0,<2.0.0
```
jgrillo-grapl, line 1: This file can go away once we get this project building with `pants`

jgrillo-grapl (changes requested):
A few things, mostly having to do with integrating this code with the rest of our python build system.

PR #1109, opened 2021-08-11
Provide a TIMEOUT_SECS variable to adjust timeouts for e2e tests

christophermaier:
Rather than providing a fixed timeout in our AWS e2e tests (to account
for the different times it takes to run there), we'll instead provide
a way to change the timeout values within the code based on an
environment variable.

This will be a better approach, because it takes some of the guesswork
out of the problem, and also allows the tests to take only as long as
they need.

It does allow our LensView query to run longer (300 vs 120 seconds) in
our Docker Compose environment, but I felt the simplicity of having
one timeout value outweighed the complexity of multiple timeouts, or
computing relative timeouts.

Apart from that, this should behave identically in our Docker Compose
setup.

### DIFF
--- a/pulumi/infra/e2e_test_runner.py
+++ b/pulumi/infra/e2e_test_runner.py
@@ -49,6 +49,7 @@ class E2eTestRunner(pulumi.ComponentResource):
                     "GRAPL_API_HOST": api.invoke_url.apply(
                         lambda url: urlparse(url).netloc
                     ),
+                    "TIMEOUT_SECS": "500",
                 },
                 timeout=60 * 15,  # 15 minutes
                 memory_size=256,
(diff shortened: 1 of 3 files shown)


pulumi/infra/e2e_test_runner.py
```diff
@@ -49,8 +49,9 @@ def __init__(
                     "GRAPL_API_HOST": api.invoke_url.apply(
                         lambda url: urlparse(url).netloc
                     ),
+                    "TIMEOUT_SECS": "500",
                 },
-                timeout=60 * 15,  # 15 minutes
+                timeout=60 * 20,  # 20 minutes
```
jgrillo-grapl, line 54: I believe the longest a lambda can run is 15min (900sec) https://docs.aws.amazon.com/whitepapers/latest/serverless-architectures-lambda/timeout.html

jgrillo-grapl (approved):

pulumi/infra/e2e_test_runner.py
christophermaier: _Sigh..._ OK.

PR #1110, opened 2021-08-11
Add some sleeps between various provisioning steps in e2e test

christophermaier:
We suspect there may be some nondeterministic shenaniganery between
various analyzer upload, data upload, and test kickoff steps that is
accounting for the variation in how long the e2e tests can take in
AWS.

Rather than simply cranking up the timeouts again, we'll add a little
time up front to try and get a better indication of where the issue
may be.

This is obviously not a Real Solution.


jgrillo-grapl (approved):
😴

Issue #1101, opened 2021-08-10
Set up unit tests for pulumi


twunderlich-grapl:
Set up a testing structure and at least one unit test for pulumi.

The candidate for the unit test is the network.py file

PR #1115, opened 2021-08-12
Pulumi unit tests

twunderlich-grapl:

https://github.com/grapl-security/grapl/issues/1101

This adds a pulumi/tests folder with a pants BUILD file.
Adds unit tests for network.py
Adds setuptools as a requirement for pulumi_aws

This sets up the first group of pulumi unit tests

Manual tests
cd pulumi
python -m pytest --disable-pytest-warnings 

Running from pants:
./pants test pulumi/tests/:tests


3rdparty/python/requirements.txt
```diff
@@ -36,5 +36,6 @@ pyjwt
 pytest
 requests
 redis
+setuptools
```
twunderlich-grapl, line 39: This is required since pulumi_aws apparently needs pkg_resources

pulumi/infra/test_network.py
```diff
@@ -0,0 +1,67 @@
...
+        # This is necessary since infra.config calls this
+        if args.token == "aws:index/getCallerIdentity:getCallerIdentity":
+            return {
+                "account_id": "000000000000"
+            }
+        return {}
+
+pulumi.runtime.set_mocks(MyMocks())
+
+sys.modules['infra.network.infra.config'] = MagicMock()
```
christophermaier: Is "infra.network.infra.config" correct? That's not an existing module 🤔

pulumi/tests/test_network.py
```diff
@@ -0,0 +1,70 @@
...
+            return {
+                "account_id": "000000000000"
+            }
+        return {}
+
+pulumi.runtime.set_mocks(MyMocks())
+
+# Environment variable is required on import so we can't monkeypatch. Set it here for now
+os.environ["GRAPL_OPERATIONAL_ALARMS_EMAIL"] = "fake-ops-email@fake.domain"
+sys.modules['infra.network.infra.config'] = MagicMock()
```
twunderlich-grapl, line 30: Log-term we'll need to either make infra.config side-effect free on import OR create a test runner that sets the environment variables properly.

pulumi/tests/test_network.py
```diff
@@ -38,9 +38,9 @@ def test_number_of_subnets(self):
 
         def check_number_of_subnets(args):
             public_subnets, private_subnets = args
-            self.assertEqual(len(public_subnets), 2, f'There should be 2 public subnets since desired_az_spread is 2')
-            self.assertEqual(len(public_subnets), len(private_subnets),
-                             f'There should be the same number of public subnets and private subnets')
+            assert len(public_subnets) == 2, f'There should be 2 public subnets since desired_az_spread is 2'
```
christophermaier, line 42: No reason to have these messages be formatting strings (`f"..."`), since there are no parameters.

3rdparty/python/constraints.txt
```diff
@@ -1,62 +1,54 @@
-# Generated by ./build-support/manage_virtualenv.sh on Wed 04 Aug 2021 04:13:46 PM PDT
+# Generated by ./build-support/manage_virtualenv.sh on Fri Aug 13 12:13:23 EDT 2021
 # DO NOT EDIT BY HAND
-ansicolors==1.1.8
```
twunderlich-grapl, line 3: Can someone check this carefully? There seems to be a lot of deleted libraries, including pantsbuild and pex, which makes me think that there's an issue with the regenerate-constraints command

3rdparty/python/requirements.txt
christophermaier: I wonder if this is somehow related to https://github.com/grapl-security/grapl/blob/dd8ba3bfd442f95b88d64d4fdac5b3e8a4959d7c/build-support/manage_virtualenv.sh#L54-L61

🤔

3rdparty/python/constraints.txt
christophermaier: 🤔 Yeah, this looks very odd. Was this on a Mac, by chance? That's about the only thing I can think of at the moment that _might_ explain this.

christophermaier: For reference, here's what it looks like on my machine (a Chromebook):

```diff
--- a/3rdparty/python/constraints.txt
+++ b/3rdparty/python/constraints.txt
@@ -1,4 +1,4 @@
-# Generated by ./build-support/manage_virtualenv.sh on Wed 04 Aug 2021 04:13:46 PM PDT
+# Generated by build-support/manage_virtualenv.sh on Fri 13 Aug 2021 12:23:27 PM EDT
 # DO NOT EDIT BY HAND
 ansicolors==1.1.8
 Arpeggio==1.10.2
@@ -6,9 +6,9 @@ arrow==1.1.1
 attrs==20.3.0
 binaryornot==0.4.4
 blessed==1.17.6
-boto3==1.18.14
-botocore==1.21.14
-bugout==0.1.15
+boto3==1.18.20
+botocore==1.21.20
+bugout==0.1.16
 certifi==2021.5.30
 chalice==1.22.4
 chardet==4.0.0
@@ -19,7 +19,7 @@ dill==0.3.4
 fasteners==0.16
 grpcio==1.39.0
 humbug==0.1.9
-hypothesis==6.14.5
+hypothesis==6.14.6
 idna==3.2
 importlib-metadata==4.6.3
 iniconfig==1.1.1
@@ -28,17 +28,17 @@ Jinja2==3.0.1
 jinja2-time==0.2.0
 jmespath==0.10.0
 MarkupSafe==2.0.1
-mypy-boto3-cloudwatch==1.18.14
-mypy-boto3-dynamodb==1.18.14
-mypy-boto3-ec2==1.18.14
-mypy-boto3-lambda==1.18.14
-mypy-boto3-route53==1.18.14
-mypy-boto3-s3==1.18.14
-mypy-boto3-sagemaker==1.18.14
-mypy-boto3-secretsmanager==1.18.14
-mypy-boto3-sns==1.18.14
-mypy-boto3-sqs==1.18.14
-mypy-boto3-ssm==1.18.14
+mypy-boto3-cloudwatch==1.18.20
+mypy-boto3-dynamodb==1.18.20
+mypy-boto3-ec2==1.18.20
+mypy-boto3-lambda==1.18.20
+mypy-boto3-route53==1.18.20
+mypy-boto3-s3==1.18.20
+mypy-boto3-sagemaker==1.18.20
+mypy-boto3-secretsmanager==1.18.20
+mypy-boto3-sns==1.18.20
+mypy-boto3-sqs==1.18.20
+mypy-boto3-ssm==1.18.20
 mypy-extensions==0.4.3
 packaging==20.9
 pantsbuild.pants==2.6.0rc1
@@ -49,14 +49,14 @@ pluggy==0.13.1
 poyo==0.5.0
 protobuf==3.17.3
 psutil==5.8.0
-pulumi==3.9.1
+pulumi==3.10.1
 pulumi-aws==4.6.0
 pulumi-docker==3.0.0
 pulumi-kafka==3.0.1
 pulumi-random==4.2.0
 py==1.10.0
 pydantic==1.8.2
-pydgraph==21.3.1
+pydgraph==21.3.2
 PyJWT==2.1.0
 pyparsing==2.4.7
 pytest==6.2.4
@@ -83,6 +83,6 @@ types-toml==0.1.3
 typing-extensions==3.7.4.3
 urllib3==1.26.6
 wcwidth==0.2.5
-wheel==0.36.2
+wheel==0.37.0
 zipp==3.5.0
 zstd==1.4.9.1
```

3rdparty/python/requirements.txt
twunderlich-grapl: So after thinking about this for a while, I don't think so. If this was the problem then I'd expect to see setuptools in the pants cache directory even if pkg_resources was not in there. My guess is with a properly updated constraints file I might be able to remove this

pulumi/infra/test_network.py
twunderlich-grapl: I've removed this since we don't really need to patch the DEPLOYMENT_NAME. 
That said, I'll note that when I tried to use infra.config by itself I ran into errors. It seems like you need to use infra.network in front so that pytest knows that infra.config is being called by infra.network. Seems weird though, so there's probably (hopefully?) a better way to do this. Or even better would be to get rid of the side effects so things could be imported without needing mocks

christophermaier (approved):
Looks good! 👍

PR #1125, opened 2021-08-16
Use our hcl-formatter image to lint and format all HCL files

christophermaier:
This has the added benefit of removing a few scripts. It also makes it
more clear that we're doing the exact same thing in CI that we do
locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ RUST_BUILD ?= debug
 UID = $(shell id -u)
 GID = $(shell id -g)
 PWD = $(shell pwd)
+COMPOSE_USER=${UID}:${GID}
 DOCKER_BUILDX_BAKE_OPTS ?=
 ifneq ($(GRAPL_RUST_ENV_FILE),)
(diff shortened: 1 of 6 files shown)


jgrillo-grapl (approved):

Makefile
```diff
@@ -350,8 +350,7 @@ lint-prettier: build-formatter ## Run ts/js/yaml lint checks
 
 .PHONY: lint-hcl
 lint-hcl: ## Check to see if Packer templates are formatted properly
-	.buildkite/scripts/lint_packer.sh
-	.buildkite/scripts/lint_nomad.sh
+	docker-compose --file=docker-compose.check.yml run --rm hcl-lint
```
inickles-grapl, line 353: My preference is that we always run with `--user` and never run containers as root, regardless of whether we're writing files back to the host or not. But I'm fine with it for now.

inickles-grapl (approved):

Makefile
inickles-grapl: I'd previously tried enforcing this by putting `user: ${UID}:${GID}` in compose files, both of which are set in the Makefile.

inickles-grapl: BTW, is this PR missing a def for `COMPOSE_USER` in the Makefile?

PR #1121, opened 2021-08-13
Data tenancy

jgrillo-grapl:
### Which issue does this PR correspond to?
https://github.com/grapl-security/issue-tracker/issues/612

### What changes does this PR make to Grapl? Why?
This PR refactors our protobufs, adding an `Envelope` message with a `tenant_id` field. We'll use this to group system data by tenant.

Consequentially, this change set impacts many disparate parts of the codebase. In the course of this work I also found cause to refactor the way we do python integration tests because we were building two different copies of the protobuf artifacts (one in docker and one in pants) which was just crazy. Now we just use pants.

### How were these changes tested?
automated tests

PR #1126, opened 2021-08-16
Remove src/python/graplinc; protobuf codegen creates correct `graplinc.` structure now.

wimax-grapl:
None. Helping jesse with protobuf stuff.

- `src/python/graplinc` was a snapshot of codegen, wildly out of date; I've removed it.
- pants is smart enough to automatically import `src/protos:graplinc` when it encounters a `from graplinc.` in Python code - _yes, really!_ - so we don't need explicit dependencies anymore (with reasonable exceptions like the repl)

CI and `make repl`
but the fact that previous users of `src/python/graplinc` can now use the new one, no issue - implies good things.


src/python/grapl_analyzerlib/grapl_analyzerlib/BUILD
```diff
@@ -1,10 +1,5 @@
 python_library(
-    dependencies=[
-        # TODO: Ideally, this would be picked up by Pants' automatic
-        # dependency resolution, but something appears to be off with
-        # it.
-        "src/python/graplinc/graplinc/grapl/api"
-    ],
+    dependencies=[],
```
jgrillo-grapl: I think this kwarg can go away completely

jgrillo-grapl (approved):
🥇

src/BUILD
```diff
@@ -5,12 +5,11 @@
 grapl_python_distribution(
     name = "graplinc_grapl_api",
     dependencies = [
-        "src/proto/graplinc/grapl/api",
-        "src/python/graplinc/graplinc/grapl/api"
+        "src/proto:graplinc",
```
wimax-grapl, line 8: yes, this technically includes more than we need to, and I could do a proto library that's just the api. Not convinced it's worth it.

src/python/grapl_analyzerlib/grapl_analyzerlib/BUILD
wimax-grapl: yep i noticed that :)

src/BUILD
jgrillo-grapl: Yes, I agree. I think all our protobuf stubs should be one "block", and ultimately they should be wrapped in a library. This library (`python-proto`, and its counterpart `rust-proto`) should be the sole point-of-contact for code which needs to serialize or deserialize protocol buffers. "Internal" code should not depend directly on the protobuf stubs.

This is one of the changes I'm attempting to introduce in https://github.com/grapl-security/grapl/pull/1121

PR #1131, opened 2021-08-20
Add first policies to pulumi

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/629
Note: the original PR was https://github.com/grapl-security/grapl/pull/1105 but I apparently can't reopen after force pushing (thank you rebase)

This adds:
1. Awsguard, pulumi's base set of policies for AWS. These are set in advisory/warning mode
2. Our first set of internal policies:
    a. Prevent public s3 bucket acls (default pulumi policy)
    b. Prevent IAM policies containing s3:ListBuckets
3. A README giving an overview of policies

These are our first set of pulumi policies as code. Policies will allow us to prevent some known bad configurations such as unencrypted resources or wildcard policies, etc. This basically adds a bit of test coverage for known bad cases, similar to setting semgrep rules in place (but executed post create/on update). These won't matter as much now in the early stages but will be more important as we more people contribute. Importantly, this provides a manner to add security guiderails for infrastructure

Functionally this should give warnings for awsguard default policies, and block s3 buckets with public acls and IAM policies with s3:ListBuckets

Tested using the grapl-local stack

To test locally:
Set up local grapl (make up)
cd pulumi/grapl
pulumi preview --policy-pack ../policies/grapl/ --policy-pack ../policies/3rdparty/awsguard/

I didn't get any warnings for the custom policy pack but did get warnings for awsguard, primarily around s3 logging and api gateway types.

Once these are tested and confirmed they can be published to pulumi and enabled server-side there:
pulumi policy publish <org-name>
pulumi policy enable <org-name>/<policy-pack-name> <latest|version>


Makefile
```diff
@@ -493,7 +493,7 @@ repl: ## Run an interactive ipython repl that can import from grapl-common etc
 	./pants --no-pantsd repl --shell=ipython src/python/repl
 
 .PHONY: pulumi-prep
-pulumi-prep: graplctl lambdas build-ux ## Prepare some artifacts in advance of running a Pulumi update (does not run Pulumi!)
+pulumi-prep: graplctl build-lambda-zips build-ux ## Prepare some artifacts in advance of running a Pulumi update (does not run Pulumi!)
```
twunderlich-grapl, line 496: lambdas no longer exists, we need to use build-lambda-zips instead

twunderlich-grapl:
@jgrillo-grapl I think I've addressed all the changes you requested in the original PR

twunderlich-grapl:
Open question:
For the awsguard policies that show advisories should I fix them in this PR? Are these even things we care about ie should I suppress the s3 logging warning?

PR #1134, opened 2021-08-23
Set Name tag automatically for ec2 servers

twunderlich-grapl:

https://github.com/grapl-security/issue-tracker/issues/649
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/649

Automatically sets a Name tag for ec2 servers. This will be the name passed to the server and the instance number. For example, when we spin up 3 Nomad servers we'll see nomad-servers-cluster-0, nomad-servers-cluster-1, nomad-servers-cluster-2

Tested using pulumi up and seeing the diff, and then manually checked upon creation


jgrillo-grapl (approved):
LGTM

PR #1139, opened 2021-08-24
Localstack fork: Preload ElasticMQ to reduce load times and enable Nomad

wimax-grapl:
- Whenever we bring up a Localstack container, it goes and downloads `elasticmq.jar` - a 32M download every time we `up` localstack!
- Downloading ElasticMQ to the container causes an unspecified error when running under Nomad. 
- I'm trying to just avoid that issue entirely.

This mysterious URL is the exact one that Localstack uses: 
https://github.com/localstack/localstack/blob/c2fa7bb5dba64d9318c7a5bbe295ac96ad3f2a02/localstack/constants.py

We can stop doing this all the moment we have Kafka.

### DIFF
--- a/docker-compose.build.yml
+++ b/docker-compose.build.yml
@@ -125,6 +125,12 @@ services:
       context: .
       dockerfile: Dockerfile.pulumi
 
+  localstack:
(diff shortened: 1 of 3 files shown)


jgrillo-grapl (approved):
:100:

PR #1145, opened 2021-08-30
Kafka executor

colin-grapl:
I plan to add integration tests.

### TODO
* Clean up unused crates
* Re-enable warnings
* Add integration tests
* Add documentation comments
* Add metrics


src/rust/kafka-executor/src/event_consumer.rs
```diff
@@ -0,0 +1,82 @@
...
+const KAFKA_CONSUMER_TOPIC: &str = "KAFKA_CONSUMER_TOPIC";
+const KAFKA_CONSUMER_BROKERS: &str = "KAFKA_CONSUMER_BROKERS";
+const KAFKA_CONSUMER_GROUP_ID: &str = "KAFKA_CONSUMER_GROUP_ID";
+const KAFKA_CONSUMER_CLIENT_ID: &str = "KAFKA_CONSUMER_CLIENT_ID";
+const KAFKA_CONSUMER_SESSION_TIMEOUT_MS: &str = "KAFKA_CONSUMER_SESSION_TIMEOUT_MS";
+const KAFKA_CONSUMER_ENABLE_AUTO_COMMIT: &str = "KAFKA_CONSUMER_ENABLE_AUTO_COMMIT";
+const KAFKA_CONSUMER_MAX_POLL_INTERVAL_MS: &str = "KAFKA_CONSUMER_MAX_POLL_INTERVAL_MS";
+const KAFKA_CONSUMER_REQUEST_TIMEOUT_MS: &str = "KAFKA_CONSUMER_REQUEST_TIMEOUT_MS";
+
+pub struct EventConsumer<C=DefaultConsumerContext, R=DefaultRuntime>
```
jgrillo-grapl: Would "KafkaConsumer" work here? Similarly, this file can be renamed to "kafka_consumer.rs". The reason I suggest this is that this isn't so much about events (we don't really have a coherent "Event" type here).

src/rust/kafka-executor/src/event_decoder.rs
```diff
@@ -0,0 +1,5 @@
+pub trait EventDeserializer {
```
jgrillo-grapl, line 1: Can we also lose the "Event" word here?

src/rust/kafka-executor/src/event_encoder.rs
```diff
@@ -0,0 +1,5 @@
+pub trait EventSerializer {
```
jgrillo-grapl, line 1: and here

src/rust/kafka-executor/src/event_handler.rs
```diff
@@ -0,0 +1,13 @@
+use crate::errors::CheckedError;
+
+#[async_trait::async_trait]
+pub trait EventHandler {
```
jgrillo-grapl, line 2: and here

src/rust/kafka-executor/src/event_handler.rs
```diff
@@ -0,0 +1,13 @@
...
+    type InputEvent;
+    type OutputEvent;
+    type Error: CheckedError;
+
+    async fn handle_event(
```
jgrillo-grapl, line 7: This isn't really a generic event handler. In Kafka we have three types of clients:
1. Producer
2. Consumer
3. Stream processor

This appears to be a client of type (3).

src/rust/kafka-executor/src/event_producer.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+#[derive(Clone)]
+pub struct EventProducer {
+    producer: FutureProducer,
+    pub(crate) topic_name: String,
+}
+
+impl EventProducer {
+    pub async fn emit_event(&self, payload: &[u8]) -> Result<(), EventProducerError> {
```
jgrillo-grapl, line 32: I'd suggest calling this `produce`

src/rust/kafka-executor/src/event_producer.rs
```diff
@@ -0,0 +1,89 @@
...
+    KafkaError(#[from] KafkaError),
+}
+
+#[derive(Clone)]
+pub struct EventProducer {
+    producer: FutureProducer,
+    pub(crate) topic_name: String,
+}
+
+impl EventProducer {
```
jgrillo-grapl, line 31: just "Producer"? or maybe "KafkaProducer"?

src/rust/kafka-executor/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+    EventSerializerT: event_encoder::EventSerializer<OutputEvent = OutputEventT> + Unpin + Send + 'static,
+>(
+    message: BorrowedMessage<'_>,
+    mut event_handler: Pin<&mut EventHandlerT>,
+    consumer: &EventConsumer<DefaultConsumerContext, DefaultRuntime>,
+    producer: Pin<&EventProducer>,
+    mut deserializer: Pin<&mut EventDeserializerT>,
+    mut serializer: Pin<&mut EventSerializerT>,
+) {
+    let event = match deserializer.deserialize(message.payload().unwrap()) {
```
jgrillo-grapl, line 124: This assumption feels a little less naunced than it needs to be. We're going to have an Envelope event which _should_ be deserialized but the _inner_ event _shouldn't_ be deserialized (unless the caller requests it).

EDIT: maybe we can express this in the `Deserializer`?

jgrillo-grapl (changes requested):
A few call-outs, seems like a reasonable approach overall

src/rust/Cargo.toml
```diff
@@ -11,6 +11,7 @@ members = [
   "./graph-merger",
   "./grapl",
   "./grapl-config",
+  "./kafka-executor",
```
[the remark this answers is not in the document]
colin-grapl: Executor to me is "drives execution". So, given some sort of a service construct, this orchestrates its execution - like tokio does for futures. Client feels lower level to me.

I don't have strong opinions on this though. `kafka-event-handler` would be another option as EventHandler is the primary interface exported.

src/rust/kafka-executor/src/lib.rs
colin-grapl: I would imagine that that's expressed in the Deserializer, yeah. If you want to just pass it along, your Deserializer would only deserialize the Envelope.

colin-grapl: We should probably unconditionally deserialize the Envelope in the process_message function, and then allow the Deserializer to optionally deserialize the internal bytes (it can just hand bytes back if it isn't interested).

PR #1142, opened 2021-08-25
Twunderlich/execute nomad jobs via pulumi

twunderlich-grapl:
### Which issue does this PR correspond to?
https://github.com/grapl-security/issue-tracker/issues/653

### What changes does this PR make to Grapl? Why?
1. Updates pulumi so it can run the grapl core nomad job
2. Updates grapl-core so that it runs via a docker network. This means that consul connect is disabled for the moment
3. Removes local infra jobs from docker-compose since they've been ported to nomad
4. Some minor formatting changes

### How were these changes tested?
make start-nomad-dev
make up-pulumi
In your browser go to http://localhost:4646/
Confirm that all tasks are green

PR #1147, opened 2021-08-31
Port kafka and zookeeper to nomad

ghost:
https://github.com/grapl-security/issue-tracker/issues/602

This ports kafka and zookeeper to Nomad and gets them running via consul connect

### DIFF
--- a/local-grapl.env
+++ b/local-grapl.env
@@ -60,10 +60,10 @@ AWS_REGION=us-east-1
 LOCALSTACK_HOST=aws.grapl.test
 DGRAPH_HOST=dgraph.grapl.test
 GRAPL_API_HOST=api.grapl.test
-KAFKA_BROKER_HOST=kafka.grapl.test
+KAFKA_BROKER_HOST=kafka-broker
 REDIS_HOST=redis.grapl.test
 S3_HOST=s3.${AWS_REGION}.amazonaws.com
-ZOOKEEPER_HOST=zookeeper.grapl.test
+ZOOKEEPER_HOST=zookeeper
 
 # Lambda "Hosts"
 ########################################################################
--- a/local-grapl.env
+++ b/local-grapl.env
@@ -60,10 +60,10 @@ AWS_REGION=us-east-1
 LOCALSTACK_HOST=aws.grapl.test
 DGRAPH_HOST=dgraph.grapl.test
 GRAPL_API_HOST=api.grapl.test
+KAFKA_BROKER_HOST=kafka.grapl.test
 REDIS_HOST=redis.grapl.test
 S3_HOST=s3.${AWS_REGION}.amazonaws.com
-ZOOKEEPER_HOST=zookeeper
-KAFKA_BROKER_HOST=kafka-broker
+ZOOKEEPER_HOST=zookeeper.grapl.test
 
 # Lambda "Hosts"
 ########################################################################
--- a/nomad/local/grapl-local-infra.nomad
+++ b/nomad/local/grapl-local-infra.nomad
@@ -208,13 +208,7 @@ job "grapl-local-infra" {
 
       connect {
         sidecar_service {
-          proxy {
-            # connect to zookeeper
-            upstreams {
-              destination_name = "zookeeper"
-              local_bind_port  = var.ZOOKEEPER_PORT
-            }
-          }
+
         }
       }
 
@@ -248,12 +242,7 @@ job "grapl-local-infra" {
 
       connect {
         sidecar_service {
-          proxy {
-            upstreams {
-              destination_name = "kafka-broker"
-              local_bind_port  = var.KAFKA_BROKER_PORT
-            }
-          }
+
         }
       }
 
--- a/nomad/local/grapl-local-infra.nomad
+++ b/nomad/local/grapl-local-infra.nomad
@@ -178,8 +178,11 @@ job "grapl-local-infra" {
 
       config {
         image = "confluentinc/cp-kafka:6.2.0"
-        # Hitting OOM errors
-        memory_hard_limit = 2048
+      }
+
+      resources {
+        cpu    = 500
+        memory = 1024
       }
 
       env {
--- a/nomad/local/grapl-local-infra.nomad
+++ b/nomad/local/grapl-local-infra.nomad
@@ -182,7 +182,7 @@ job "grapl-local-infra" {
 
       config {
         image = "confluentinc/cp-kafka:6.2.0"
-        ports = "kafka"
+        ports = ["kafka"]
         network_aliases = [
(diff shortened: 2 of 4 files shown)


wimax-grapl:
Seems superseded by https://github.com/grapl-security/grapl/pull/1142/files#diff-82c0615c711b7e15090cd77892a10ac647a8b99b262fcb613a5bd878615c8bd2R182-R184

nomad/local/grapl-local-infra.nomad
```diff
@@ -142,4 +167,95 @@ job "grapl-local-infra" {
...
+      mode = "bridge"
+    }
+
+    task "kafka" {
+      driver = "docker"
+
+      config {
+        image = "confluentinc/cp-kafka:6.2.0"
+        # Hitting OOM errors
+        memory_hard_limit = 2048
```
d0nut-grapl, line 182: I'd recommend using the `resources` stanza. By default tasks are allotted 100Mhz compute and 300 MB memory usage. This 300 is probably being passed into the container. By setting the `memory_hard_limit` in this way, I believe (and I could be wrong) that nomad will still treat this task as if it only has 300MB of memory usage which means it could place it on a machine without much headroom left for memory. This could result in an OOM at some point.

Perhaps try 

```
resource {
  cpu = 1000 # or more; we'll need to set resources for all tasks when we get around to deploying in production
  memory = 2048 # this will act as a hard limit and reserve 2GB of memory for this task specifically
}
```

nomad/local/grapl-local-infra.nomad
```diff
@@ -6,6 +6,21 @@ variable "container_registry" {
 
 # The following variables are all-caps to clue in users that they're
```
d0nut-grapl, line 7: btw if you make changes to a nomad or packer file, you'll want to run `make format-hcl` otherwise you might fail our linter on CI (which it appears this change has). 

It's annoying, I know haha.

nomad/local/grapl-local-infra.nomad
```diff
@@ -26,6 +41,16 @@ variable "LOCALSTACK_HOST" {
   description = "External hostname for Localstack"
 }
 
+variable "ZOOKEEPER_HOST" {
```
d0nut-grapl, line 44: It doesn't seem this is in use anywhere

nomad/local/grapl-local-infra.nomad
```diff
@@ -6,6 +6,21 @@ variable "container_registry" {
 
 # The following variables are all-caps to clue in users that they're
 # imported from `local-grapl.env`.
+variable "KAFKA_BROKER_HOST" {
```
d0nut-grapl, line 9: This also seems like this isn't used anywhere

nomad/local/grapl-local-infra.nomad
```diff
@@ -142,4 +167,95 @@ job "grapl-local-infra" {
...
+        KAFKA_TRANSACTION_STATE_LOG_MIN_ISR = 1
+        KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR = 1
+        KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS = 0
+        KAFKA_JMX_PORT = var.KAFKA_JMX_PORT
+        KAFKA_JMX_HOSTNAME = "localhost"
+        KAFKA_LOG4J_ROOT_LOGLEVEL = "INFO"
+      }
+    }
+
+      service {
```
d0nut-grapl: this all looks good to me!

nomad/local/grapl-local-infra.nomad
```diff
@@ -142,4 +167,95 @@ job "grapl-local-infra" {
...
+        name = "zookeeper"
+        port = var.ZOOKEEPER_PORT
+        tags = ["zookeeper"]
+
+        connect {
+          sidecar_service {
```
d0nut-grapl: Just for my clarification, did this not work if you didn't have the bidirectional consul connect setup? I don't see anything passed into zookeeper regarding kafka's location. Does zookeeper establish additional connections to kafka or does it just keep open the connection that kafka establishes?

I wonder if you could just reduce this down to:

```
service {
  name = "zookeeper"
  port = var.ZOOKEEPER_PORT
  tags = ["zookeeper"]
  
  connect {
    sidecar_service { }
  }
}
```

d0nut-grapl (changes requested):

local-grapl.env
```diff
@@ -60,10 +60,10 @@ AWS_REGION=us-east-1
 LOCALSTACK_HOST=aws.grapl.test
 DGRAPH_HOST=dgraph.grapl.test
 GRAPL_API_HOST=api.grapl.test
+KAFKA_BROKER_HOST=kafka.grapl.test
```
twunderlich-grapl, line 63: The host variables may no longer be necessary with consul connect. I may remove them, after I take a quick look at Max's work to expose dgraph

nomad/local/grapl-local-infra.nomad
twunderlich-grapl: Yep, I realized that after I pushed it up 🤦 . I've fixed that for the moment

nomad/local/grapl-local-infra.nomad
twunderlich-grapl: That seems to work 👍

nomad/local/grapl-local-infra.nomad
```diff
@@ -142,4 +157,130 @@ job "grapl-local-infra" {
...
+      driver = "docker"
+
+      config {
+        image = "confluentinc/cp-zookeeper:6.2.0"
+      }
+
+      env {
+        ZOOKEEPER_CLIENT_PORT = var.ZOOKEEPER_PORT
+        ZOOKEEPER_TICK_TIME   = 2000
+        KAFKA_OPTS            = "-Dzookeeper.4lw.commands.whitelist=ruok"
```
twunderlich-grapl, line 246: This is necessary for the healthcheck so ruok actually works

nomad/local/grapl-local-infra.nomad
```diff
@@ -142,4 +157,130 @@ job "grapl-local-infra" {
...
+      }
+
+      service {
+        check {
+          type    = "script"
+          name    = "check_zookeeper"
+          command = "/bin/bash"
+          args = [
+            "-c",
+            "echo ruok | nc -w 2  localhost 2181 | grep imok || exit 2",
```
twunderlich-grapl, line 256: The || exit section is necessary for the health check to actually fail if imok is not returned

d0nut-grapl (approved):

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/670

PR #1152, opened 2021-09-02
Run integration tests inside Nomad with `make test-integration`

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/659

This diff:
- adds a NONFUNCTIONAL [doesn't do any service discovery or anything yet] basic port of `rust-integration-tests` from `docker-compose.yml` into a new Nomad file `integration-tests.nomad`
- Adds a new way to run this Nomad file against Nomad, wait for job completion, and exit-code based on status. Think of this as `docker-compose-with-error` but for Nomad.
- Integrates that into `make test-integration`.

image

Locally with `make test-integration`


Makefile
```diff
@@ -308,10 +308,18 @@ test-grapl-template-generator:  # Test that the Grapl Template Generator spits o
...
 # This target is not intended to be used directly from the command line, it's
 # intended for tests in docker-compose files that need the Grapl environment.
+.PHONY: test-with-env-docker
+test-with-env-docker: # (Do not include help text - not to be used directly)
+	$(MAKE) test-with-env EXEC_TEST_COMMAND=test/docker-compose-with-error.sh
+
+# This target is not intended to be used directly from the command line.
+# Think of it as a Context Manager that:
+# - Before test-time, brings up a `make up-detach`
+# - After test-time, tears it all down and dumps artifacts.
```
wimax-grapl, line 322: If you're curious what happened to `test-with-env`: this comment explains

nomad/local/nomad_cli_tools.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
```
wimax-grapl, line 1: Before reviewing `nomad_cli_tools.sh` i suggest reviewing `nomad/local/run_integration_tests.sh` instead, which shows you _how_ each of these gets used

nomad/local/run_integration_tests.sh
```diff
@@ -0,0 +1,42 @@
...
+THIS_DIR=$(dirname "${BASH_SOURCE[0]}")
+
+# shellcheck source-path=SCRIPTDIR
+source "${THIS_DIR}/nomad_cli_tools.sh"
+
+echo "--- Deploying integration tests"
+
+# Wait a short period of time before attempting to deploy infrastructure
+# shellcheck disable=SC2016
+timeout 30 bash -c -- 'while [[ -z $(nomad status 2>&1 | grep running) ]]; do printf "Waiting for nomad-agent\n";sleep 1;done'
```
wimax-grapl, line 14: Not really needed - seems highly unlikely nomad would not be up already by the time we did this (like, a Pulumi deploy would have happened first)

nomad/local/run_integration_tests.sh
```diff
@@ -0,0 +1,42 @@
...
+# the jobspec, since it's a Parameterized Batch Job.
+
+echo "--- Dispatching integration tests"
+
+job_id=$(nomad_dispatch integration-tests)
+echo "${job_id}"
+
+# Wait for them to finish
+await_nomad_dispatch_finish "${job_id}" 10
+
```
wimax-grapl, line 37: 10 attempts @ every 5 seconds - this obviously needs to be bumped

Makefile
```diff
@@ -296,9 +296,12 @@ test-typecheck: test-typecheck-docker test-typecheck-pants ## Typecheck all Pyth
 
 .PHONY: test-integration
 test-integration: export COMPOSE_PROJECT_NAME := $(COMPOSE_PROJECT_INTEGRATION_TESTS)
-test-integration: export COMPOSE_FILE := ./test/docker-compose.integration-tests.yml
+#test-integration: export COMPOSE_FILE := ./test/docker-compose.integration-tests.yml
 test-integration: build-test-integration ## Build and run integration tests
-	$(MAKE) test-with-env
+	$(WITH_LOCAL_GRAPL_ENV)
+	docker-compose --file=test/docker-compose.integration-tests.yml push
```
twunderlich-grapl, line 302: Ah yeah, this is definitely needed. Took me a bit to realize this was missing

nomad/local/nomad_cli_tools.sh
twunderlich-grapl: I suspect that we'll need to refactor some of this in the future given that it makes a number of assumptions that may not hold true forever. That said its definitely good enough for now

twunderlich-grapl (approved):

PR #1156, opened 2021-09-07
Nomad integration tests are now deployed by Pulumi (this is going into origin/nomad-integ-tests)

wimax-grapl:
Instead of deploying via `run_integration_tests.sh`, we do it in Pulumi.
Seems nice and functional and not awful.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -318,6 +318,8 @@ services:
     environment:
       TAG:
       PULUMI_CONFIG_PASSPHRASE: local-grapl-passphrase
+      # This gets exported to True by `make test-integration`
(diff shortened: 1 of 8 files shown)


docker-compose.yml
```diff
@@ -318,6 +318,8 @@ services:
     environment:
       TAG:
       PULUMI_CONFIG_PASSPHRASE: local-grapl-passphrase
+      # This gets exported to True by `make test-integration`
+      SHOULD_DEPLOY_INTEGRATION_TESTS: "${SHOULD_DEPLOY_INTEGRATION_TESTS:-False}"
```
wimax-grapl, line 322: This here is probably the most debatable aspect of this PR - we are _optionally deploying resources_ based on an environment variable.

Happy to change it if people have any better ideas.

wimax-grapl:
I have approval from cm, so I'll merge this into the nomad-integ-tests branch.

PR #1159, opened 2021-09-08
Update Pants to 2.7.0rc2

christophermaier:
We're updating to this RC on the recommendation of Toolchain Labs;
apparently there are a lot of cache performance improvements.

Some additional changes here are:
- the introduction of tool-specific dependency lockfiles
- `[pytest].pytest_plugins` has been replaced with `[pytest].extra_requirements`
- The platform-specification for `known_versions` has been tightened up

The `pulumi/policies/3rdparty/awsguard/BUILD` file was also removed, since
there are no shell files for that target.


3rdparty/python/black.lockfile
```diff
@@ -0,0 +1,119 @@
+# This lockfile was autogenerated by Pants. To regenerate, run:
```
wimax-grapl, line 1: Absolutely not a must-do, but you may want to consider adding these lockfiles to .gitattributes (so that they can be skipped at PR

wimax-grapl (approved):

3rdparty/python/black.lockfile
christophermaier: That's a great idea

PR #1160, opened 2021-09-08
Buildkite should run integration tests via Nomad

ghost:
grapl-security/issue-tracker#659

This gets integration tests via nomad working in buildkite

Check the integration tests in this PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -438,7 +438,7 @@ up-detach: build ## Bring up local Grapl and detach to return control to tty
 	unset COMPOSE_FILE
 	docker-compose \
 		--file docker-compose.yml \
-		up --detach --force-recreate --always-recreate-deps --renew-anon-volumes
+		up --detach --force-recreate --always-recreate-deps --renew-anon-volumes pulumi
 
 .PHONY: down
 down: ## docker-compose down - both stops and removes the containers
--- a/Makefile
+++ b/Makefile
@@ -494,11 +494,11 @@ local-pulumi:  ## launch pulumi via docker-compose up
 .PHONY: start-nomad-detach
 start-nomad-detach: push-local  ## Start the Nomad environment, detached
 	$(WITH_LOCAL_GRAPL_ENV)
-	sudo --preserve-env nomad/local/start_detach.sh
+	nomad/local/start_detach.sh
 
 .PHONY: stop-nomad-detach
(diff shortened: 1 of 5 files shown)


wimax-grapl (approved):

Makefile
```diff
@@ -438,7 +438,7 @@ up-detach: build ## Bring up local Grapl and detach to return control to tty
 	unset COMPOSE_FILE
 	docker-compose \
 		--file docker-compose.yml \
-		up --detach --force-recreate --always-recreate-deps --renew-anon-volumes
+		up --detach --force-recreate --always-recreate-deps --renew-anon-volumes pulumi
```
twunderlich-grapl, line 441: Only run the one container we care about. This also cuts down on the time it takes the tests to run

nomad/local/nomad_run_local_infra.sh
```diff
@@ -7,7 +7,7 @@ echo "Deploying local infrastructure."
 
 # Wait a short period of time before attempting to deploy infrastructure
 # shellcheck disable=SC2016
-timeout 30 bash -c -- 'while [[ -z $(nomad status 2>&1 | grep running) ]]; do printf "Waiting for nomad-agent\n";sleep 1;done'
+timeout 120 bash -c -- 'while [[ -z $(nomad status 2>&1 | grep running) ]]; do printf "Waiting for nomad-agent\n";sleep 1;done'
```
twunderlich-grapl, line 10: We were timing out before nomad had started

nomad/local/stop_detach.sh
```diff
@@ -2,5 +2,5 @@
 
 set -euo pipefail
 
-killall nomad || true
+sudo killall nomad || true
```
twunderlich-grapl, line 5: I suspect that this might be broken in regards to not working with the sudoers approved commands, but I'm seeing the dump_artifacts throw an exception so it's not getting to this

nomad/local/nomad_run_local_infra.sh
wimax-grapl: 120 feels long to me, but, eh

nomad/local/stop_detach.sh
wimax-grapl: I have a fix for dump artifacts btw, but good call

PR #1163, opened 2021-09-09
Make the Nomad rust integ tests actually talk kafka

wimax-grapl:
grapl-security/issue-tracker#659

### DIFF
--- a/etc/ci_scripts/dump_artifacts/nomad_artifacts.py
+++ b/etc/ci_scripts/dump_artifacts/nomad_artifacts.py
@@ -7,7 +7,7 @@ import os
 import shutil
 import sys
 from pathlib import Path
-from typing import Any, Dict, List, Union, cast
+from typing import Any, Dict, List, Optional, Union, cast
 
 from nomad import Nomad
 from nomad.api.exceptions import URLNotFoundNomadException
@@ -60,7 +60,7 @@ class NomadTask:
     name: str
     parent: NomadAllocation = dataclasses.field(repr=False)
 
-    def get_logs(self, nomad_client: Nomad, type: OutOrErr) -> str:
+    def get_logs(self, nomad_client: Nomad, type: OutOrErr) -> Optional[str]:
         try:
             return cast(
                 str,
--- a/etc/ci_scripts/dump_artifacts/nomad_artifacts.py
+++ b/etc/ci_scripts/dump_artifacts/nomad_artifacts.py
@@ -72,18 +72,28 @@ class NomadTask:
             LOGGER.info(f"Couldn't get logs for {self.name}")
             return None
 
+def _get_nomad_job_names(
+    nomad_client: Nomad
+) -> List[str]:
+    jobs: List[str] = [
+        j['ID'] 
+        for j in nomad_client.jobs
+        if not j['ParameterizedJob']
+    ]
+    return jobs
 
 def _get_nomad_logs_for_each_service(
     artifacts_dir: Path,
 ) -> Dict[str, List[NomadAllocation]]:
-    nomad = Nomad(host="localhost", timeout=5)
+    nomad_client = Nomad(host="localhost", timeout=5)
+    job_names = _get_nomad_job_names(nomad_client)
     job_to_allocs: Dict[str, List[NomadAllocation]] = {
-        job_name: [NomadAllocation(a) for a in nomad.job.get_allocations(job_name)]
-        for job_name in ("grapl-core", "grapl-local-infra", "integration-tests")
+        job_name: [NomadAllocation(a) for a in nomad_client.job.get_allocations(job_name)]
+        for job_name in job_names
     }
 
     for job, allocs in job_to_allocs.items():
-        _write_nomad_logs(nomad, artifacts_dir, job_name=job, allocs=allocs)
+        _write_nomad_logs(nomad_client, artifacts_dir, job_name=job, allocs=allocs)
 
     return job_to_allocs
 
--- a/etc/ci_scripts/dump_artifacts/nomad_artifacts.py
+++ b/etc/ci_scripts/dump_artifacts/nomad_artifacts.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Union, cast
 
 from nomad import Nomad
+from nomad.api.exceptions import URLNotFoundNomadException
 from typing_extensions import Literal
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


pulumi/infra/config.py
```diff
@@ -120,6 +120,8 @@ def _require_env_var(key: str) -> str:
 
 
 OPERATIONAL_ALARMS_EMAIL = _require_env_var("GRAPL_OPERATIONAL_ALARMS_EMAIL")
+REDIS_PORT = _require_env_var("REDIS_PORT")
+KAFKA_BROKER_PORT = _require_env_var("KAFKA_BROKER_PORT")
```
wimax-grapl, line 124: Shower thoughts:

We may want to rejigger this into an endpoint, and for local make it like 
LOCALHOST_SENTINEL_VALUE:6379
and have Nomad do a substitute

src/rust/kafka-metrics-exporter/tests/integration_tests.rs
```diff
@@ -21,13 +22,17 @@ mod integration_tests {
     };
     use tokio_stream::StreamExt;
 
+    fn brokers() -> String {
```
twunderlich-grapl, line 25: Shouldn't this be defined in grapl-config/src/lib.rs and then called here? Also can we rename this to be kafka_brokers?

src/rust/kafka-metrics-exporter/tests/integration_tests.rs
```diff
@@ -21,13 +22,17 @@ mod integration_tests {
...
+        let bucket = std::env::var("KAFKA_ENDPOINT").expect("KAFKA_ENDPOINT");
```
twunderlich-grapl, line 26: also this variable should probably be named to something like endpoint instead of bucket

pulumi/grapl/__main__.py
```diff
@@ -194,9 +190,16 @@ def main() -> None:
                         "aws_access_key_id",
                         "aws_access_key_secret",
                         "aws_endpoint",
-                        "redis_endpoint",
                     }
                 }
+                integration_test_only_job_vars = {
```
twunderlich-grapl, line 195: I'll note that we probably will want to pass this into grapl core as well, but that can be done later

src/rust/kafka-metrics-exporter/tests/integration_tests.rs
wimax-grapl: Good call, yes.

src/rust/kafka-metrics-exporter/tests/integration_tests.rs
wimax-grapl: WOW good catch

src/rust/Dockerfile
```diff
@@ -92,7 +92,7 @@ RUN cp -ar /tmp/target/. /grapl/rust/target/ && \
     cp -ar /tmp/registry/. /usr/local/cargo/registry/ && \
     rm -rf /tmp/{target,registry}
 
-CMD [ "bash", "-c", "cargo test --features \"${RUST_INTEGRATION_TEST_FEATURES}\" --test \"*\"" ]
+CMD [ "bash", "-c", "cargo test --features \"${RUST_INTEGRATION_TEST_FEATURES}\" --test \"*\" -- --nocapture" ]
```
wimax-grapl, line 95: Torn on whether I want to disable this or not

etc/ci_scripts/dump_artifacts/nomad_artifacts.py
```diff
@@ -60,25 +61,37 @@ class NomadTask:
...
+                    self.parent.allocation_id, self.name, type=type, plain=True
+                ),
+            )
+        except URLNotFoundNomadException as e:
+            LOGGER.info(f"Couldn't get logs for {self.name}")
+            return None
+
+
+def _get_nomad_job_names(nomad_client: Nomad) -> List[str]:
+    jobs: List[str] = [j["ID"] for j in nomad_client.jobs if not j["ParameterizedJob"]]
```
twunderlich-grapl, line 77: Is the job id really the job name in Nomad?
Also why are we excluding the parameterized jobs?

nomad/local/integration-tests.nomad
```diff
@@ -317,7 +323,9 @@ job "integration-tests" {
 
       env {
         # aws vars
-        AWS_REGION                  = var.aws_region
+        AWS_REGION = var.aws_region
+        # TODO: Do the "USE_LOCALHOST_SENTINEL_VALUE" check here in case we
```
twunderlich-grapl, line 327: I think this comment belongs at the top in the locals section, but absolutely agree

nomad/local/run_integration_tests.sh
```diff
@@ -17,10 +17,19 @@ echo "--- Dispatching integration tests"
 job_id=$(nomad_dispatch integration-tests)
 echo "${job_id}"
 
-await_nomad_dispatch_finish "${job_id}" $((5 * 60))
+dispatch_timed_out=0
+await_nomad_dispatch_finish "${job_id}" $((5 * 60)) && dispatch_timed_out=1
```
twunderlich-grapl, line 21: Shouldn't this be || not &&? I also suspect that await_nomad_dispatch_finish should exit 1 if it times out, not return

src/rust/Dockerfile
twunderlich-grapl: I think its worth keeping this. It makes logging easier and then its a choice for the dev on whether they want to add logs or not

twunderlich-grapl (approved):
Minor comments but this looks good. 🚢 it

etc/ci_scripts/dump_artifacts/nomad_artifacts.py
wimax-grapl: A) That is correct
B) For parameterized jobs, you will see the following:
- integration_tests (param batch job)
- integration_tests/dispatch-12338838 (a dispatched instance of integration_tests)

We don't want to bother "getting logs" for the uninstanced `integration_tests` (but also, I'll document this)

nomad/local/run_integration_tests.sh
wimax-grapl: re: the &&/||
yes, good call, thanks.

> I also suspect that await_nomad_dispatch_finish should exit 1 if it times out, not return
if it exits, then this whole script exits, is my understanding.
However, I still want to perform the `if dispatch_timed_out` behavior if that occurs.
As such, sticking with `return`

PR #1165, opened 2021-09-10
Remove local registry

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/676

This removes all the local registry pieces since apparently Nomad doesn't need that.
Nomad will pick up local images as long as they are not tagged with latest, who knew?

You'll notice that test-integration runs in about 7 minutes and change instead of the 30+ minutes previously 🥳 

make test-integration
Nomad should run through everything without any docker pull related failures (tests may still fail since integration test suites are not passing yet)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,7 @@ build-test-typecheck: build-python-wheels
 		--file ./test/docker-compose.typecheck-tests.yml
 
 .PHONY: build-test-integration
-build-test-integration: build
+build-test-integration: build-local
 	$(WITH_LOCAL_GRAPL_ENV) \
 	$(DOCKER_BUILDX_BAKE) --file ./test/docker-compose.integration-tests.yml
 
@@ -209,6 +209,9 @@ build-docker-images: graplctl
 
 .PHONY: build
 build: build-lambda-zips build-docker-images ## Build Grapl services
+
+.PHONY: build-local
+build: build-lambda-zips build-docker-images-local ## Build Grapl services
 
 .PHONY: build-formatter
 build-formatter:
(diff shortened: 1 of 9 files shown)


nomad/local/integration-tests.nomad
```diff
@@ -34,6 +34,36 @@ variable "aws_endpoint" {
...
+  default     = "dev"
+  description = "The tagged version of the graphql_endpoint_tests we should deploy."
+}
+
+variable "rust_integration_tests_tag" {
+  type        = string
+  default     = "dev"
+  description = "The tagged version of the rust-integration-tests we should deploy."
+}
+
```
wimax-grapl, line 66: While I appreciate the effort, I don't think there's ever any case where we'll be running a non-`dev` of integration tests.

Like, I could imagine a scenario where we swap out one of our services for an older version to see if that fixed an issue - but IDK why anybody would want a different tagged version of an integration test than the one on your machine.

wimax-grapl (approved):
Approved, with 1 comment that doesn't NEED to be taken care of.

nomad/local/integration-tests.nomad
twunderlich-grapl: True, removing now

wimax-grapl (changes requested):
I think we need `$(WITH_LOCAL_GRAPL_ENV)` in `build:`

wimax-grapl (approved):

Makefile
```diff
@@ -438,7 +439,7 @@ up-detach: build ## Bring up local Grapl and detach to return control to tty
 	unset COMPOSE_FILE
 	docker-compose \
 		--file docker-compose.yml \
-		up --detach --force-recreate --always-recreate-deps --renew-anon-volumes pulumi
+		run pulumi
```
wimax-grapl, line 442: tbh I can't guarantee that `run` will have the same properties, with renew anon volumes etc

i foudn the following "works"
```

	# TODO: This could potentially be replaced with a docker-compose run, but
	#  it doesn't have all these useful flags
	docker-compose \
		--file docker-compose.yml \
		up --force-recreate --always-recreate-deps --renew-anon-volumes \
		--exit-code-from pulumi \
		pulumi
```

PR #1166, opened 2021-09-11
Clean up our "USE_LOCALSTACK_SENTINEL_VALUE" nomad stuff

wimax-grapl:
Doesn't correspond to any specific issue.

Basically, our local-grapl previously used special sentinel values like USE_LOCALSTACK_SENTINEL_VALUE or USE_REDIS_SENTINEL_VALUE (these are badly named) to say "hey, use the grapl-local-infra equivalent".

This makes 3 improvements:
- Use 1 unified sentinel value for every service that would need this: LOCAL_GRAPL_REPLACE_IP
- Just use str replaces for it
- If a var, like `var.aws_endpoint`, should actually consumed by using `local.aws_endpoint` instead, we prefix the var with a `_` to clue in the programmer that it shouldn't be directly consumed.


nomad/local/nomad_cli_tools.sh
```diff
@@ -21,36 +21,39 @@ nomad_dispatch() {
...
+    job_id=$(jq -rn --arg input "${job_id}" '$input|@uri')
+    echo "http://localhost:4646/ui/jobs/${job_id}"
+}
+
 nomad_stop_job() {
     local -r job_id="${1}"
     local -r dispatch_output=$(curl_quiet --request DELETE --data "{}" "${NOMAD_ENDPOINT}/v1/job/${job_id}")
 }
 
-nomad_get_allocation() {
```
wimax-grapl, line 29: turns out `get_per_task_results` was superior to this

pulumi/grapl/__main__.py
```diff
@@ -135,6 +135,12 @@ def main() -> None:
 
         kafka = Kafka("kafka")
 
+        # These are created in `grapl-local-infra.nomad` and not applicable to prod.
+        # Nomad will replace the LOCAL_GRAPL_REPLACE_IP sentinel value with the correct IP.
+        aws_endpoint = "http://LOCAL_GRAPL_REPLACE_IP:4566"
+        kafka_endpoint = "LOCAL_GRAPL_REPLACE_IP:19092"  # intentionally not 29092
```
twunderlich-grapl, line 141: I _think_ that these may need to move to Pulumi.local-grapl.yaml in the future since presumably we'll be passing in kafka and redis endpoint info to live services

nomad/local/nomad_cli_tools.sh
```diff
@@ -21,36 +21,39 @@ nomad_dispatch() {
     echo "${job_id}"
 }
 
+url_to_nomad_job_in_ui() {
+    local job_id="${1}"
+    # urlencode
+    job_id=$(jq -rn --arg input "${job_id}" '$input|@uri')
```
twunderlich-grapl, line 27: It might be nice to rename this var to urlencoded_job_id to clarify what this var is expecting

twunderlich-grapl (approved):
Some minor nits, but everything looks good

pulumi/grapl/__main__.py
wimax-grapl: I wasn't sure, since Redis is actually provisioned BY pulumi

PR #1167, opened 2021-09-12
Remove non-build stuff from `docker-compose.integration-tests.yml` and rename it to reflect that

wimax-grapl:
Semi related to https://github.com/grapl-security/issue-tracker/issues/659

Previously this file was peppered with like entrypoints, commands, etc that have nothing to do with building.
All of *that* is now taken care of in integration-tests.nomad.
So reduce this down to a `.build.yml` file like the main services.

make test-integration


twunderlich-grapl (approved):
LGTM

PR #1176, opened 2021-09-16
Enable ratel in local grapl, add instructions.

wimax-grapl:
Please ensure that my exposing-8080-to-host-OS is security-acceptable.

### DIFF
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -410,6 +410,11 @@ job "grapl-core" {
     content {
       network {
         mode = "bridge"
+        port "dgraph-alpha-port" {
+          # Primarily here to let us use ratel
+          static = alpha.value.http_port
+          to = alpha.value.http_port
+        }
       }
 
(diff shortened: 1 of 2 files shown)


nomad/grapl-core.nomad
```diff
@@ -410,6 +410,11 @@ job "grapl-core" {
     content {
       network {
         mode = "bridge"
+        port "dgraph-alpha-port" {
+          # Primarily here to let us use ratel
+          static = alpha.value.http_port
```
d0nut-grapl, line 415: So I never 100% understood what using both `static` and `to` here did. I assume that this just makes `static` on the host to `to` in the bridged network.

I wonder if we should just use `to` without `static` and then, when you wish to connect to dgraph w/ ratel, you look in nomad at the allocated port. This will allow us to run HA versions locally, but perhaps that's not useful enough of capability to avoid the static port.

d0nut-grapl (approved):
Approved w/ comment

PR #1178, opened 2021-09-17
Remove old Consul and Nomad infrastructure

christophermaier:
Since we're managing our Consul and Nomad clusters separately, we can remove all the old code for managing them in this repository.


wimax-grapl (approved):
goodbye old friend

PR #1185, opened 2021-09-22
Nomad merge main

wimax-grapl:
Merge main into nomad

Also, there's some funky stuff in `integration-tests.nomad` - we now run python on the host OS, no docker or anything, via pants.

PR #1186, opened 2021-09-22
Port engagement-creator from a Lambda to a service-that-retrieves-events

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/691

Oh boy, where to start
- Port engagement creator out of `lambdas.Dockerfile` and into the regular Dockerfile (which will disappear when I land https://github.com/grapl-security/grapl/pull/1185 😢 )
- Pull Analyzer Executor's nice and reusable SQS primitives out of `analyzer_executor_lib` and into `grapl_common` - thanks Pants, pretty easy!
- Refactor our BUILD files for grapl-common after a dumb thing made me lose an hour of time (you previously needed to specify every new subfolder in grapl-common/BUILD gross)
- Add an entry to Nomad for it

I can't confirm the logic works yet, but I see an every-few-seconds "queue was empty" from my Engagement Creator job in Nomad 4646.


src/python/grapl-common/BUILD
```diff
@@ -2,9 +2,6 @@ grapl_python_distribution(
     dependencies = [
```
wimax-grapl, line 2: # This is the thing I lost an hour on - I'd have needed to add `/sqs` here and that's annoying and well hidden :(

jgrillo-grapl (approved):

PR #1187, opened 2021-09-23
Ensure that manage_virtualenv.sh runs from the repository root

christophermaier:
Previously, if the script was executed from another
directory (particularly `build-support` itself), it would fail, but
leave behind an extra `venv` directory, which caused subsequent
problems with Pants.

Now, all the logic takes place from the repository root, eliminating
these issues.


twunderlich-grapl (approved):
Thanks for doing this!

PR #1189, opened 2021-09-23
Enable deploying grapl-core to nomad in AWS

ghost:
https://github.com/grapl-security/issue-tracker/issues/679

This enables the deployment of the grapl-core to Nomad in AWS.
Note: Not all jobs succeed at the moment, and it relies on having the platform-infrastructure pulumi deployed.

Specific changes

- Update the image format to support the cloudsmith format (ie the ability to add /raw in AWS)
- Conditionally pass in AWS environment variables via Nomad template stanzas. This uses template env format
- Pull artifact tags from config
- Run the deploy job to AWS
- Switches NomadJob to take a Mapping and not an Output, with type changes throughout __main__.py
- Get pulumi preview to work in buildkite. This includes an UGLY hack to deal with pulumi preview since preview doesn't complete outputs

Run the platform-infrastructure stacks
Set up local port forwarding to one of the nomad servers via `aws ssm start-session --target i-02aa25e760ae991a5 --document-name AWS-StartPortForwardingSession --parameters '{"portNumber":["4646"], "localPortNumber": ["4846"]}'` remembering to update the instance id and potentially change the localPortNumber
Add `nomad:address: localhost:4846` to your pulumi stack config
pulumi up.
Once run, go to http:localhost:4846 and you should see the grapl-core job with some nomad groups in running/completed status and some in failed status. This is expected since there is other ongoing work to set up peering, security groups etc (Most jobs should fail if they can't connect to their database of choice)


nomad/grapl-core.nomad
```diff
@@ -9,20 +9,29 @@ variable "container_registry" {
...
 
+variable "container_repo" {
+  type        = string
+  default     = ""
+  description = "The container repo inside the registry in which we can find Grapl services. Requires a trailing /"
+}
+
 variable "aws_access_key_id" {
   type        = string
+  default     = "DEFAULT_NOMAD_AWS_ACCESS_KEY_ID"
```
twunderlich-grapl, line 20: Instead of using an empty string or "test" I'm setting an easily searchable string. Let me know if you've got better ideas here.

nomad/grapl-core.nomad
```diff
@@ -305,6 +314,14 @@ locals {
...
   # enabled
   rust_backtrace = 1
+
+  heredoc = <<EOH
+GRAPL_AWS_ENDPOINT          = ${local.aws_endpoint}
+GRAPL_AWS_ACCESS_KEY_ID     = ${var.aws_access_key_id}
+GRAPL_AWS_ACCESS_KEY_SECRET = ${var.aws_access_key_secret}
+EOH
+
+  local_vars = (var._aws_endpoint != "http://LOCAL_GRAPL_REPLACE_IP:4566") ? "IN_AWS=TRUE" : local.heredoc
```
twunderlich-grapl, line 324: The template requires that some environment variable be defined. I originally was setting it to an empty string, which broke nomad's ability to parse its diff. If there's a different env var you want to be defined in AWS I'm open to suggestions

nomad/grapl-core.nomad
```diff
@@ -677,12 +696,15 @@ job "grapl-core" {
         sidecar = false
       }
 
+      template {
+        data        = local.local_vars
```
twunderlich-grapl, line 700: Contrary to Nate's original thoughts, this had to be data instead of source

nomad/grapl-core.nomad
```diff
@@ -724,14 +746,16 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image        = "${var.container_registry}grapl/node-identifier:${var.node_identifier_tag}"
-        network_mode = "grapl-network"
+        image = "${var.container_registry}grapl/${var.container_repo}node-identifier:${var.node_identifier_tag}"
```
twunderlich-grapl, line 749: This funky interpolation is done so we can support both the local grapl format (grapl/container-name:tag) and the cloudsmith format (registry/org/repo/container-name:tag)

nomad/grapl-core.nomad
```diff
@@ -724,14 +746,16 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image        = "${var.container_registry}grapl/node-identifier:${var.node_identifier_tag}"
-        network_mode = "grapl-network"
```
twunderlich-grapl, line 728: grapl-network has to be removed for AWS deploys since the network doesn't exist there

pulumi/grapl/__main__.py
```diff
@@ -102,6 +103,27 @@ def main() -> None:
         "analyzer-matched-subgraphs-bucket", analyzer_matched_emitter.bucket.bucket
     )
 
+    sysmon_generator_queue = ServiceQueue("sysmon-generator")
```
twunderlich-grapl, line 106: Moved this up since it was used by both local grapl and aws grapl

pulumi/infra/network.py
```diff
@@ -49,7 +49,7 @@ def __init__(
             cidr_block=str(cidr_block),
             enable_dns_hostnames=True,
             enable_dns_support=True,
-            tags={"Name": f"{name}-{DEPLOYMENT_NAME}"},
+            tags={"Name": f"{name}"},
```
twunderlich-grapl, line 52: This was done so that it was easier to do vpc peering

pulumi/infra/cache.py
```diff
@@ -50,6 +50,7 @@ def __init__(
                     cidr_blocks=["0.0.0.0/0"],
                 )
             ],
+            tags={"Name": f"{name}"},
```
twunderlich-grapl, line 53: Make it easy to do resource lookups from the platform-infrastructure pulumi projects

nomad/grapl-core.nomad
christophermaier: While not necessarily a change for now, I wonder if it might be better in the long term to just have this refer to some `${var.node_identifier_image}` (and similar for every other service), and pull the composition of the name up to a higher level. This is what happened when Pulumi set up our Fargate jobs before Nomad, for instance.

If things are local, we're generally just looking for `grapl/node-identifier:latest` (or whatever).

In real deployments, however, we'll have a Cloudsmith URL, which (as you point out) is a bit different, like `docker.cloudsmith.io/grapl/raw/node-identifier:20210806001956-3bd9525c`. Currently the `raw` bit is just a catch-all repository that all the containers go into, but longer term, we'll have additional repositories that we'll promote containers through as they go through the pipeline, such that production releases will be using something like `docker.cloudsmith.io/grapl/release/node-identifier:20210806001956-3bd9525c`.

Ultimately, we shouldn't be embedding that logic deep down in each job.

nomad/grapl-core.nomad
```diff
@@ -797,16 +823,18 @@ job "grapl-core" {
...
+        env         = true
       }
 
       env {
         # AWS vars
-        AWS_REGION                  = var.aws_region
-        GRAPL_AWS_ACCESS_KEY_ID     = var.aws_access_key_id
-        GRAPL_AWS_ACCESS_KEY_SECRET = var.aws_access_key_secret
-        GRAPL_AWS_ENDPOINT          = local.aws_endpoint
+        AWS_REGION = var.aws_region
```
christophermaier, line 837: We should adopt `AWS_DEFAULT_REGION` here, as pointed out above.

Otherwise, someone reading all these services is going to start asking questions about why some are one way and others are a different way.

nomad/grapl-core.nomad
```diff
@@ -830,16 +858,18 @@ job "grapl-core" {
...
```
christophermaier, line 872: Since this is a Python app, this will _have_ to be `AWS_DEFAULT_REGION`; it probably only works because we've only ever run in `us-east-1`.

We should dig into https://github.com/grapl-security/issue-tracker/issues/301 at some point 😅

nomad/grapl-core.nomad
```diff
@@ -872,15 +902,18 @@ job "grapl-core" {
...
```
christophermaier, line 916: This is also a Python app, so this also needs to be `AWS_DEFAULT_REGION`.

nomad/grapl-core.nomad
```diff
@@ -617,21 +634,23 @@ job "grapl-core" {
...
       config {
-        image        = "${var.container_registry}grapl/graph-merger:${var.graph_merger_tag}"
-        network_mode = "grapl-network"
+        image = "${var.container_registry}grapl/${var.container_repo}graph-merger:${var.graph_merger_tag}"
+      }
+
+      template {
+        data        = local.local_vars
+        destination = "graph-merger.env"
+        env         = true
```
wimax-grapl, line 643: so does this actually write stuff to a file named graph-merger.env ?
and then source it as an env?
just as a nomad template baby idk what's going on here tbh

nomad/grapl-core.nomad
```diff
@@ -923,20 +956,23 @@ job "grapl-core" {
...
+      template {
+        data        = local.local_vars
+        destination = "graphql-endpoint.env"
+        env         = true
+      }
+
       env {
         DEPLOYMENT_NAME               = var.deployment_name
         RUST_LOG                      = var.rust_log
         AWS_REGION                    = var.aws_region
```
christophermaier, line 972: Looks like the JavaScript SDK is one that requires `AWS_REGION`: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-region.html

FUN.

nomad/grapl-core.nomad
```diff
@@ -983,12 +1019,15 @@ job "grapl-core" {
...
+        env         = true
+      }
+
       env {
         # For the DynamoDB client
-        AWS_REGION                  = var.aws_region
-        GRAPL_AWS_ACCESS_KEY_ID     = var.aws_access_key_id
-        GRAPL_AWS_ACCESS_KEY_SECRET = var.aws_access_key_secret
-        GRAPL_AWS_ENDPOINT          = local.aws_endpoint
+        AWS_REGION = var.aws_region
```
christophermaier, line 1030: If this is Rust, we should stick to `AWS_DEFAULT_REGION`.

nomad/grapl-core.nomad
```diff
@@ -1026,22 +1065,25 @@ job "grapl-core" {
...
-        AWS_REGION                  = var.aws_region
-        REDIS_ENDPOINT              = local.redis_endpoint
-        GRAPL_LOG_LEVEL             = var.rust_log
-        RUST_LOG                    = var.rust_log
-        RUST_BACKTRACE              = local.rust_backtrace
+        DEST_BUCKET_NAME      = var.unid_subgraphs_generated_bucket
+        DEAD_LETTER_QUEUE_URL = var.sysmon_generator_dead_letter_queue
+        SOURCE_QUEUE_URL      = var.sysmon_generator_queue
+        AWS_DEFAULT_REGION    = var.aws_region # boto3 prefers this one
+        AWS_REGION            = var.aws_region
```
christophermaier, line 1082: Sysmon is Rust, so keep `AWS_DEFAULT_REGION`, ditch the `boto3` comment, and ditch `AWS_REGION`.

nomad/grapl-core.nomad
```diff
@@ -1058,20 +1100,22 @@ job "grapl-core" {
...
-        REDIS_ENDPOINT              = local.redis_endpoint
-        GRAPL_LOG_LEVEL             = var.rust_log
-        RUST_LOG                    = var.rust_log
-        RUST_BACKTRACE              = local.rust_backtrace
+      env {
+        DEST_BUCKET_NAME      = var.unid_subgraphs_generated_bucket
+        DEAD_LETTER_QUEUE_URL = var.osquery_generator_dead_letter_queue
+        SOURCE_QUEUE_URL      = var.osquery_generator_queue
+        AWS_DEFAULT_REGION    = var.aws_region # boto3 prefers this one
+        AWS_REGION            = var.aws_region
```
christophermaier, line 1114: Same for OSQuery

pulumi/grapl/__main__.py
```diff
@@ -118,36 +132,37 @@ def main() -> None:
     )
     pulumi.export("ux-bucket", ux_bucket.bucket)
```
wimax-grapl, line 134: apologies, i too refactored this whole section :(

wimax-grapl (approved):

pulumi/grapl/__main__.py
```diff
@@ -156,41 +171,13 @@ def main() -> None:
...
-            sysmon_generator_queue=sysmon_generator_queue.main_queue_url,
-            sysmon_generator_dead_letter_queue=sysmon_generator_queue.dead_letter_queue_url,
-            unid_subgraphs_generated_bucket=unid_subgraphs_generated_emitter.bucket,
-            user_auth_table=dynamodb_tables.user_auth_table.name,
-            user_session_table=dynamodb_tables.user_session_table.name,
-        ).apply(
+        grapl_core_job_vars = pulumi.Output.all(**nomad_inputs,).apply(
             lambda inputs: {
                 # This is a special directive to our HCL file that tells it to use Localstack
                 "_aws_endpoint": aws_endpoint,
```
christophermaier, line 177: We should probably call it `_localstack_aws_endpoint` or something similar.

pulumi/grapl/__main__.py
```diff
@@ -156,41 +171,13 @@ def main() -> None:
...
-                "deployment_name": config.DEPLOYMENT_NAME,
+                "aws_access_key_id": aws.config.access_key,
+                "aws_access_key_secret": aws.config.secret_key,
                 "grapl_test_user_name": config.GRAPL_TEST_USER_NAME,
-                "aws_region": aws.get_region().name,
                 "_redis_endpoint": redis_endpoint,
```
christophermaier, line 181: Why does this have an underscore? Is this also somehow applicable only for Localstack?

christophermaier: Is the idea that everything in this map, aside from the `inputs` values, is _only_ for Localstack use?

pulumi/grapl/__main__.py
```diff
@@ -239,115 +226,54 @@ def _get_integration_test_job_vars(
...
+                "container_registry": "docker.cloudsmith.io/",
+                "container_repo": "raw/",
+                "grapl_test_user_name": config.GRAPL_TEST_USER_NAME,
+                "_redis_endpoint": cache.endpoint,
+                # TODO: consider replacing with the previous per-service `configurable_envvars`
+                "rust_log": "DEBUG",
+                # Build Tags. We use per service tags so we can update services independently
+                "analyzer_dispatcher_tag": pulumi_config.require_object("artifacts")[
+                    "analyzer-dispatcher"
+                ],
```
christophermaier, line 243: I'd just extract the artifacts object somewhere above this and then just pull out the thing I needed, like `artifacts["analyzer-dispatcher"]`

christophermaier: Incidentally, this is where you'd stick the logic for composing the container image names: `analyzer_dispatcher_image: image_name("analyzer-dispatcher", artifacts, registry, repo)`

(though we can come up with a nicer interface.)

pulumi/infra/cache.py
christophermaier: Unless you're augmenting the resource lookup search with other information, this tag alone won't provide a sufficiently narrow scope.

It would also be good for this PR comment to be a code comment on this line; once we have a more formalized relationship between stacks, such that we can simply refer to a stack output that gives us the security group ID directly, we won't need this tag anymore.

pulumi/infra/network.py
christophermaier: How so?

nomad/grapl-core.nomad
twunderlich-grapl: Yes, exactly. I do agree that it's weird though

nomad/grapl-core.nomad
```diff
@@ -9,20 +9,29 @@ variable "container_registry" {
   description = "The container registry in which we can find Grapl services. Requires a trailing /"
 }
 
+variable "container_repo" {
+  type        = string
+  default     = ""
+  description = "The container repo inside the registry in which we can find Grapl services. Requires a trailing /"
```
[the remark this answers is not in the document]
twunderlich-grapl: Ah only if it's not empty. I'll clarify the phrasing

nomad/grapl-core.nomad
christophermaier: Heh, I had this same question... took a little reading to figure out what the heck was going on.

nomad/grapl-core.nomad
```diff
@@ -9,20 +9,29 @@ variable "container_registry" {
...
 
 variable "aws_access_key_secret" {
   type        = string
+  default     = "DEFAULT_NOMAD_AWS_ACCESS_KEY_SECRET"
   description = "The aws access key secret used to interact with AWS."
 }
 
 variable "_aws_endpoint" {
   type        = string
+  default     = "DEFAULT_NOMAD_AWS_ENDPOINT"
```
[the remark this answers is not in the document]
wimax-grapl: Yeah, I guess I'm not 100% sure what DEFAULT_NOMAD_AWS_ENDPOINT _means_

nomad/grapl-core.nomad
```diff
@@ -617,21 +634,23 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image        = "${var.container_registry}grapl/graph-merger:${var.graph_merger_tag}"
-        network_mode = "grapl-network"
+        image = "${var.container_registry}grapl/${var.container_repo}graph-merger:${var.graph_merger_tag}"
+      }
+
+      template {
+        data        = local.local_vars
```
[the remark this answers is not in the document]
wimax-grapl: +1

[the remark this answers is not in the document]
twunderlich-grapl: I'm changing it to conditionally_defined_env_vars, which I think is a bit better since we're likely going to have env vars defined only for AWS as well

pulumi/grapl/__main__.py
twunderlich-grapl: To the latter comment, yes these are localstack specific values

nomad/grapl-core.nomad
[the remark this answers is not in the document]
twunderlich-grapl: So it does need a string so that replace will still work. Instead of using an empty string, I thought it would be better to have a named variable so if it shows up it would be easy to trace. I've changed it to start as DUMMY, but I'd be open to switching it to a simple empty string if people would prefer that.

pulumi/grapl/__main__.py
twunderlich-grapl: Meh

pulumi/grapl/__main__.py
twunderlich-grapl: To the first, I've added a comment to this effect, but this is basically a hint that we should be using the local. version of the env var instead of the var. version in the hcl (because we need the mutated version of this)

nomad/grapl-core.nomad
twunderlich-grapl: If you're ok with it, I'd like to tackle this in a later PR. I do agree that I'd prefer to pass in a full image from Pulumi.

pulumi/infra/cache.py
twunderlich-grapl: Added the comment. And since we're only creating one redis security group, having it be a SG and tagged with Name= main-cache is sufficient for now.

twunderlich-grapl: Oh, I just realized that you meant that if we have multiple deployments in the same AWS account. In that case you're right, we'd need to also look up the stack as well (which is a separate tag provided by auto-tagging)

nomad/grapl-core.nomad
```diff
@@ -299,6 +308,16 @@ locals {
...
+
+  # This is used to conditionally submit env variables via template stanzas.
+  local_only_env_vars = <<EOH
+GRAPL_AWS_ENDPOINT          = ${local.aws_endpoint}
+GRAPL_AWS_ACCESS_KEY_ID     = ${var.aws_access_key_id}
+GRAPL_AWS_ACCESS_KEY_SECRET = ${var.aws_access_key_secret}
+EOH
+  # We need to submit an env var otherwise you can end up with a weird nomad state parse error.
+  aws_only_env_vars              = "DUMMY_VAR=TRUE"
+  conditionally_defined_env_vars = (var._aws_endpoint != "http://LOCAL_GRAPL_REPLACE_IP:4566") ? local.aws_only_env_vars : local.local_only_env_vars
```
christophermaier, line 320: Minor, but it would be more direct to remove the negation: "if endpoint is local, use local variables, otherwise non-local"

nomad/grapl-core.nomad
christophermaier: Another PR is great; would you mind filing an issue for it so we don't lose track of it?

nomad/grapl-core.nomad
christophermaier: Because this is the lone `AWS_REGION` usage (as opposed to `AWS_DEFAULT_REGION`), it would be worthwhile to add a comment to the effect of "No, this isn't a mistake, the JavaScript SDK only recognizes `AWS_REGION`, not `AWS_DEFAULT_REGION`"

nomad/grapl-core.nomad
```diff
@@ -980,22 +1022,24 @@ job "grapl-core" {
...
-        REDIS_ENDPOINT              = local.redis_endpoint
-        GRAPL_LOG_LEVEL             = var.rust_log
-        RUST_LOG                    = var.rust_log
-        RUST_BACKTRACE              = local.rust_backtrace
+        DEST_BUCKET_NAME      = var.unid_subgraphs_generated_bucket
+        DEAD_LETTER_QUEUE_URL = var.sysmon_generator_dead_letter_queue
+        SOURCE_QUEUE_URL      = var.sysmon_generator_queue
+        AWS_DEFAULT_REGION    = var.aws_region
+        REDIS_ENDPOINT        = local.redis_endpoint
+        GRAPL_LOG_LEVEL       = var.rust_log
```
christophermaier, line 1040: I think the `GRAPL_LOG_LEVEL` variable only applies to Python services; the Rust ones currently use `RUST_LOG`.

Long term, we should probably work to harmonize these as much as possible. In the meantime, we should not add unnecessary variables, since that just increases the cognitive overhead.

nomad/grapl-core.nomad
```diff
@@ -1012,20 +1056,21 @@ job "grapl-core" {
...
-        GRAPL_LOG_LEVEL             = var.rust_log
-        RUST_LOG                    = var.rust_log
-        RUST_BACKTRACE              = local.rust_backtrace
+      env {
+        DEST_BUCKET_NAME      = var.unid_subgraphs_generated_bucket
+        DEAD_LETTER_QUEUE_URL = var.osquery_generator_dead_letter_queue
+        SOURCE_QUEUE_URL      = var.osquery_generator_queue
+        AWS_DEFAULT_REGION    = var.aws_region
+        REDIS_ENDPOINT        = local.redis_endpoint
+        GRAPL_LOG_LEVEL       = var.rust_log
```
christophermaier, line 1071: Same here, re: `GRAPL_LOG_LEVEL`.

nomad/grapl-provision.nomad
```diff
@@ -76,6 +85,16 @@ variable "provisioner_tag" {
...
+
+  # This is used to conditionally submit env variables via template stanzas.
+  local_only_env_vars = <<EOH
+GRAPL_AWS_ENDPOINT          = ${local.aws_endpoint}
+GRAPL_AWS_ACCESS_KEY_ID     = ${var.aws_access_key_id}
+GRAPL_AWS_ACCESS_KEY_SECRET = ${var.aws_access_key_secret}
+EOH
+  # We need to submit an env var otherwise you can end up with a weird nomad state parse error
+  aws_only_env_vars              = "DUMMY_VAR=TRUE"
+  conditionally_defined_env_vars = (var._aws_endpoint != "http://LOCAL_GRAPL_REPLACE_IP:4566") ? local.aws_only_env_vars : local.local_only_env_vars
```
christophermaier, line 97: Same point re: negation.

nomad/grapl-core.nomad
twunderlich-grapl: https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/697

christophermaier (approved):

nomad/grapl-core.nomad
twunderlich-grapl: So actually rust throws errors if you use AWS_DEFAULT_REGION instead of AWS_REGION, so I've switched rust to use AWS_REGION

PR #1190, opened 2021-09-24
Nomad mount grapl root; clean up old, unused Nomad infrastructure

wimax-grapl:
For the future merge-main-into-nomad, we need to make the Grapl root available to Nomad so it can run `./pants` outside of Docker.

This is dynamically written upon every `make up-detach` now (and gitignored).


nomad/local/start_detach.sh
```diff
@@ -2,34 +2,66 @@
...
+    cat << EOF > "${VOLUME_CONFIG_FILE}"
+# Written dynamically by 'write_mount_grapl_root_volume_config'
+client {
+  host_volume "grapl-root-volume" {
+    path = "${PWD}"
+    read_only = false
+  }
+}
+EOF
+}
```
wimax-grapl, line 48: # ^ this function is the only real "new" stuff; he rest is just refactored into functions

nomad/local/local_grapl_registry.sh
```diff
@@ -1,17 +0,0 @@
-#!/bin/bash
```
d0nut-grapl, line 1: o7 ;__;

d0nut-grapl (approved):

PR #1203, opened 2021-10-07
De-Lambda-ify Provisioner, e2e tests

wimax-grapl:
Tangentially related to https://github.com/grapl-security/issue-tracker/issues/690
There are a few python_awslambdas left - dgraph_ttl and model plugin dpeloyer - that are just dead code waiting to be clawed out. 

- Make Provisioner a pex
- Make e2e tests a pex
- add a hack to e2e tests to make them work as pexs (pytest test discovery breaks a tad)
- remove all stuff about schemas in e2e tests - that was all about testing MPD and that's gone now

make test-e2e

### DIFF
--- a/src/python/e2e-test-runner/e2e_test_runner/main.py
+++ b/src/python/e2e-test-runner/e2e_test_runner/main.py
@@ -1,4 +1,6 @@
+import os
 import sys
+from pathlib import Path
 
 from grapl_common.debugger.vsc_debugger import wait_for_vsc_debugger
 from grapl_common.grapl_logger import get_module_grapl_logger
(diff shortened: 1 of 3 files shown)


src/python/e2e-test-runner/e2e_test_runner/main.py
```diff
@@ -1,32 +1,35 @@
...
     LOGGER.info(f"tests completed with status code {result}")
 
-    if result == 0:
-        return  # Lambda runtime will flip out if sys.exit(0) -_-
-    else:
-        sys.exit(result)
+    sys.exit(result)
+
+
+def _cd_to_test_directory() -> None:
```
wimax-grapl, line 26: Not saying this is *the right way* to do it, but it works for now and gets us out of weird awslambda stuff.

PR #1206, opened 2021-10-07
Update the GraphQL to build in containers via 'docker run'

inickles-grapl:
this works locally wtffff


inickles-grapl:
Notes to future self: graphql service stderr:
```
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module 'express'
Require stack:
- /graphql_endpoint/server.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/graphql_endpoint/server.js:3:17)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:79:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/graphql_endpoint/server.js' ]
}
```

howwwww. running this same docker image locally, even with non-standard image tag works just fine. that package is in the image, and it starts up fine.

I'm not sure how this part works: https://github.com/grapl-security/grapl/blob/inickles/graphql-build/pulumi/grapl/__main__.py#L308, but it looks like maybe the image tag isn't getting to nomad? If so, then it should fallback to the default, `dev`, https://github.com/grapl-security/grapl/blob/inickles/graphql-build/nomad/grapl-core.nomad#L215, but then how does an image with that tag exist on the buildkite agent in the first place? and why would it work for me locally after having 1) docker system prune -a, 2) export TAG=asdf, 3) make test-e2e.
🤔 🤔🤔🤔🤔

PR #1209, opened 2021-10-08
python integration tests now use a named-docker-volume for the Pants cache. Improve runtime of some hypothesis tests.

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/706

- Python integration tests now use a mounted Docker volume for caching. This will improve runtime in CI and locally (after an initial "slow" populate-stuff run.)
- Some slower Hypothesis tests were reduced from 100 examples to 25 examples.
- Modified behavior of stdout/stderr getting printed during integration-test time.

`make test-integration`

### DIFF
--- a/nomad/local/integration-tests.nomad
+++ b/nomad/local/integration-tests.nomad
@@ -147,8 +147,8 @@ job "integration-tests" {
 
   group "python-integration-tests" {
     restart {
-      # I guess I can let it try 2x
-      attempts = 1
+      # Make this a one-shot job. Absolute worst case, Buildkite reruns it for us.
+      attempts = 0
     }
 
     network {
@@ -193,6 +193,18 @@ job "integration-tests" {
           source   = var.grapl_root
           target   = "/mnt/grapl-root"
           readonly = false
+        }
+
+        mount {
+          type = "volume"
+          target = "/mnt/pants-cache"
+          source = "pants-cache-volume"
+          readonly = false
+          volume_options {
+            # Upon initial creation of this volume, *do* copy in the current
+            # contents in the Docker image.
+            no_copy = false
+          }
         }
       }
 
--- a/nomad/local/integration-tests.nomad
+++ b/nomad/local/integration-tests.nomad
@@ -196,9 +196,9 @@ job "integration-tests" {
         }
 
         mount {
-          type = "volume"
-          target = "/mnt/pants-cache"
-          source = "pants-cache-volume"
+          type     = "volume"
+          target   = "/mnt/pants-cache"
+          source   = "pants-cache-volume"
           readonly = false
           volume_options {
             # Upon initial creation of this volume, *do* copy in the current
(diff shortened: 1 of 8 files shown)


nomad/local/integration-tests.nomad
```diff
@@ -147,8 +147,8 @@ job "integration-tests" {
 
   group "python-integration-tests" {
     restart {
-      # I guess I can let it try 2x
-      attempts = 1
+      # Make this a one-shot job. Absolute worst case, Buildkite reruns it for us.
+      attempts = 0
```
wimax-grapl, line 151: This turned out to be a bad idea, since my timeout for `make test-integration` would have to account for the extra runtime of a potential retry.

src/python/Dockerfile
```diff
@@ -91,4 +91,5 @@ ENV PANTS_LOCAL_STORE_DIR=/mnt/pants-cache/lmdb_store
 ENV PANTS_SETUP_CACHE=/mnt/pants-cache/setup
 
 # /mnt/grapl-root is mounted by `integration-tests.nomad` at runtime.
-ENTRYPOINT ["/bin/bash", "-o", "errexit", "-o", "nounset", "-c", "cd /mnt/grapl-root && ./pants filter --filter-target-type='python_tests' :: | xargs ./pants --tag='-needs_work' test --pytest-args='-m integration_test'"]
+# -rA reports stdout/stderr for all tests, not just failed ones.
+ENTRYPOINT ["/bin/bash", "-o", "errexit", "-o", "nounset", "-c", "cd /mnt/grapl-root && ./pants filter --filter-target-type='python_tests' :: | xargs ./pants --tag='-needs_work' test --pytest-args='-m integration_test -rA'"]
```
christophermaier, line 95: Rather than using a `cd`, we should specify a `WORKDIR` in the Dockerfile. Additionally, once your entrypoint starts to get this complex (and just plain _long_), it should probably be a standalone script that is executed. If we took that route, we could get rid of all the `/bin/bash -o errexit -o nounset -c` stuff. The default shell for Linux in Dockerfiles is `/bin/sh -c`, which is certainly all that is needed for executing a script; the only reason we set such an entrypoint in our old docker-compose files was because we were overriding very non-trivial commands, and that was the most convenient and proper way of doing that. What we're doing here is a very different scenario.

In fact, I think you could just leave out `ENTRYPOINT` altogether and set `CMD` to `/custom-script-that-runs-the-tests.sh` and call it a day.

Also, are there long options for `-r` and `-A` that are more self-documenting?

src/python/grapl-tests-common/grapl_tests_common/setup_tests.py
```diff
@@ -42,12 +42,12 @@ def exec_pytest() -> int:
 
     result = pytest.main(
         [
-            "--capture=no",
+            "--capture=no",  # disable stdout capture
```
christophermaier, line 45: This comment is probably superfluous.

PR #1211, opened 2021-10-08
Delete source code for model plugin deployer and dgraph-ttl.

wimax-grapl:
(And cleaned up the Makefile a bit.)

I'm tackling metric-forwarder next, but that's a bit more in-depth.


jgrillo-grapl (approved):
👏

PR #1221, opened 2021-10-13
Autofix local nomad chrome workaround

twunderlich-grapl:
Automatically fix the chrome workaround. This will be a recurring issue whenever the chromeos linux kernel updates

I've moved the setup script, and split it into a library file and an exec file. 
I've then updated start_detach to try autofixing the nomad issue using the workaround function

### DIFF
--- a/etc/chromeos/BUILD
+++ b/etc/chromeos/BUILD
@@ -1,0 +1,1 @@
+shell_library()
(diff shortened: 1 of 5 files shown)


wimax-grapl (approved):
great refactor, thx

PR #1223, opened 2021-10-14
Use either stack reference or override for nomad address

ghost:
This sets the nomad address explicitly via a custom provider. The nomad address will take either a value from the config or will take the nomad address from the nomad stack. This lets us support two use cases:
1. We can connect directly to nomad and therefore should use the nomad address via the stack reference (ie buildkite would use this)
2. We want to use ssm port forwarding or override the address for some other reason (ie we want to run pulumi from our laptop but don't have a way to connect directly to nomad)

Not tested yet. I blew away all my pulumi changes this morning in order to test networking changes. Will test prior to merging


wimax-grapl (approved):
Please test before merging as promised :)

PR #1246, opened 2021-11-03
Create Consul Server SSM port forwarding script

ghost:
This adds a script to automate consul server port forwarding

Ran the script and went to localhost:8500

### DIFF
--- a/bin/aws/ssm_consul_server.sh
+++ b/bin/aws/ssm_consul_server.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SERVERS=$(aws ec2 describe-instances --filter Name=tag:Name,Values="Consul Server")
+CONSUL_SERVER_INSTANCE_ID=$(echo "${SERVERS}" | jq -r .Reservations[0].Instances[0].InstanceId)
+
+echo "--- Instance: ${CONSUL_SERVER_INSTANCE_ID}"
+
+aws ssm start-session \
+    --target "${CONSUL_SERVER_INSTANCE_ID}" \
+    --document-name AWS-StartPortForwardingSession \
+    --parameters '{"portNumber":["8500"], "localPortNumber": ["8500"]}'


bin/aws/ssm_consul_server.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SERVERS=$(aws ec2 describe-instances --filter Name=tag:Name,Values="Consul Server")
+CONSUL_SERVER_INSTANCE_ID=$(echo "${SERVERS}" | jq -r .Reservations[0].Instances[0].InstanceId)
```
christophermaier, line 6: You could shorten this to a single command if you like:

```sh
aws ec2 describe-instances \
    --filter Name=tag:Name,Values="Consul Server" \
    --query="Reservations[0].Instances[0].InstanceId" \
    --output=text
```
The AWS CLI uses [JMESPath](https://jmespath.org/) for the query language, which is similar to (but not the same as!) `jq`, and can be useful. Here, the main value would be simplifying the Bash (not that it's particularly complicated Bash, of course, but less Bash is often better 😅).

bin/aws/ssm_consul_server.sh
```diff
@@ -0,0 +1,13 @@
...
+
+echo "--- Instance: ${CONSUL_SERVER_INSTANCE_ID}"
+
+aws ssm start-session \
+    --target "${CONSUL_SERVER_INSTANCE_ID}" \
+    --document-name AWS-StartPortForwardingSession \
+    --parameters '{"portNumber":["8500"], "localPortNumber": ["8500"]}'
```
christophermaier, line 13: Might be useful to parameterize the local port number (with a default value of 8500), for those of us that might plausibly be connecting to servers in multiple accounts.

bin/aws/ssm_consul_server.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
christophermaier, line 1: Should this be in the platform-infrastructure repository instead?

christophermaier (approved):

PR #1247, opened 2021-11-03
Improve the SSM scripts' UX

ghost:
This adds changes to the SSM scripts suggested by Chris, which I missed before merging it in 🤦 

- Documented the scripts and improved messaging to the user
- Allowed overriding the local port for consul server and nomad server
- Removed the dependency on jq

Tested by manually by running the script


bin/aws/ssm_consul_server.sh
```diff
@@ -1,13 +1,39 @@
...
+SSM_PARAMETERS=$(
+    cat << EOF
+{
+  "portNumber": ["8500"],
+  "localPortNumber": ["${LOCAL_PORT_TO_FORWARD_TO}"]
+}
+EOF
+)
+
+echo "Connecting to a consul server in AWS PROFILE: ${AWS_PROFILE:-"No AWS Profile is set, please run export AWS_PROFILE=foo"} on port 8500 and forwarding to ${LOCAL_PORT_TO_FORWARD_TO}"
```
christophermaier, line 22: I think this is going to generate a weird message when the variable is unset.

bin/aws/ssm_consul_server.sh
```diff
@@ -1,13 +1,39 @@
...
-CONSUL_SERVER_INSTANCE_ID=$(echo "${SERVERS}" | jq -r .Reservations[0].Instances[0].InstanceId)
+LOCAL_PORT_TO_FORWARD_TO="${1:-8500}"
+
+SSM_PARAMETERS=$(
+    cat << EOF
+{
+  "portNumber": ["8500"],
+  "localPortNumber": ["${LOCAL_PORT_TO_FORWARD_TO}"]
+}
+EOF
```
christophermaier, line 19: Curious why you pulled this out... is it to get around quoting issues?

bin/aws/ssm_nomad_server.sh
```diff
@@ -1,13 +1,38 @@
 #!/usr/bin/env bash
```
christophermaier, line 1: Since all these scripts are basically the same apart from the tag used in the instance filter and the port that needs to be forwarded, it might be good to put all that into a common library that is sourced by the individual scripts, which then just parameterize the calls.

bin/aws/ssm_consul_server.sh
twunderlich-grapl: Yep. This was actually how Max had set up the nomad agent ssm script because that's where he was using passed in port variables. 
I was thinking of potentially printing this to the screen for debug purposes

bin/aws/ssm_consul_server.sh
twunderlich-grapl: Switching this to a if -z check

bin/aws/ssm_nomad_server.sh
twunderlich-grapl: Done

bin/aws/lib/ssm_tools.sh
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+    echo "Connecting to a ${SERVER_TYPE} in AWS PROFILE: ${AWS_PROFILE} on port ${REMOTE_PORT} and forwarding to ${LOCAL_PORT}"
+    echo "To connect to a ${SERVER_TYPE} in a different AWS Account change your AWS_PROFILE environment variable"
+
+    # shellcheck disable=SC2140
+    SERVER_INSTANCE_ID=$(
+        aws ec2 describe-instances \
+            --filter Name=tag:Name,Values="${SERVER_TYPE}" \
+            --filter Name="instance-state-name",Values="running" \
```
twunderlich-grapl, line 33: Added filter to ensure that the server is running

bin/aws/ssm_consul_server.sh
```diff
@@ -1,13 +1,17 @@
...
+# This will allow you to view the consul UI and interact with consul in the cli
+#
+# Input (optional): Local port to forward to. Defaults to port 8500
+
 set -euo pipefail
 
-SERVERS=$(aws ec2 describe-instances --filter Name=tag:Name,Values="Consul Server")
-CONSUL_SERVER_INSTANCE_ID=$(echo "${SERVERS}" | jq -r .Reservations[0].Instances[0].InstanceId)
+# shellcheck source=bin/aws/lib/ssm_tools.sh
+source lib/ssm_tools.sh
```
christophermaier, line 12: You can do something like this instead, which is a little less repetitive:
```shell
# shellcheck source-path=SCRIPTDIR
source "$(dirname "${BASH_SOURCE[0]}")/../lib/ssm_tools.sh"
```

bin/aws/lib/ssm_tools.sh
christophermaier: Nice catch

bin/aws/ssm_nomad_agent.sh
```diff
@@ -1,24 +1,16 @@
...
-)
+set -euo pipefail
 
-AGENTS=$(aws ec2 describe-instances --filter Name=tag:Name,Values="Nomad Agent")
-NOMAD_AGENT_INSTANCE_ID=$(echo "${AGENTS}" | jq -r .Reservations[0].Instances[0].InstanceId)
+# shellcheck source=bin/aws/lib/ssm_tools.sh
+source lib/ssm_tools.sh
 
-echo "--- Instance: ${NOMAD_AGENT_INSTANCE_ID}"
+PORT_TO_FORWARD="${1:-1234}"
```
christophermaier, line 14: Is this more to access the Grapl UI rather than the Nomad agent _per se_?

twunderlich-grapl: Yes. I've added an array to make this clearer in the user-facing messaging

christophermaier (approved):
Lots of really great improvements here! 👍

PR #1249, opened 2021-11-03
Improve chromeos install idempotency

ghost:
https://github.com/grapl-security/issue-tracker/issues/771

- Installs apt-utils which apparently is a dependency of docker install
- Adds a step to nuke $HOME/.venv if it exists since Jesse ran into issues when that did exist
- Adds message to make it clearer that nomad workaround was automatically installed

### DIFF
--- a/etc/chromeos/lib/installs.sh
+++ b/etc/chromeos/lib/installs.sh
@@ -21,7 +21,7 @@ fix_shell_completion() {
 
 install_build_tooling() {
     echo "Install build tooling"
-    sudo apt install -y build-essential libclang1
+    sudo apt install -y apt-utils build-essential libclang1
 }
 
 # potentially replace with podman in the future?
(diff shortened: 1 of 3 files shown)


etc/chromeos/lib/installs.sh
```diff
@@ -106,3 +106,14 @@ install_nomad_chromeos_workaround() {
     sudo mkdir -p "/lib/modules/$(uname -r)/"
     echo '_/bridge.ko' | sudo tee -a "/lib/modules/$(uname -r)/modules.builtin"
 }
+
+nuke_home_venv() {
+    # Apparently if you have ~/.venv that breaks the pyenv install, so we nuke it.
```
jgrillo-grapl, line 111: Oops, I mis-reported this. I had to nuke my `~/.pyenv` directory >_<. Sorry.

jgrillo-grapl (approved):

etc/chromeos/lib/installs.sh
twunderlich-grapl: Ah, that makes more sense. I'll fix now

jgrillo-grapl (approved):
👍

etc/chromeos/lib/installs.sh
```diff
@@ -46,13 +49,23 @@ install_rust_and_utilities() {
 }
 
 install_pyenv() {
-    echo "Install pyenv and set python version to 3.7.10"
+    echo "Install pyenv and set python version to ${PYENV_PYTHON_VERSION}"
     sudo apt-get install -y make libssl-dev zlib1g-dev libbz2-dev \
         libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev \
         xz-utils tk-dev libffi-dev liblzma-dev
+
+    # Check if pyenv directory exists and delete it so we can have a clean.
```
twunderlich-grapl, line 57: I made the choice here to delete the old directory, but it may make sense in the future to switch to if installed pyenv install $VERSION && pyenv global $VERSION

PR #1250, opened 2021-11-03
Clean up graplctl and e2e tests

jgrillo-grapl:
NA

This PR removes a bunch of AWS Lambda, Docker Swarm based functionality from `graplctl` and a Buildkite testing pipeline which depends on it.

Automated tests


wimax-grapl (approved):
my fav type of PR

.buildkite/pipeline.testing.yml
```diff
@@ -1,28 +0,0 @@
----
```
christophermaier, line 1: Due to the nature of the pipelines, we should leave this file in place... you can have a dummy step that just runs `echo "Do some real tests later"`.

christophermaier (changes requested):

christophermaier (approved):

PR #1251, opened 2021-11-04
Fix buildkite quoting issue breaking the pipeline

ghost:
NA, this is really a fix for https://github.com/grapl-security/grapl/pull/1250

The build is broken due to buildkite magic parsing https://buildkite.com/grapl/grapl-testing/builds/67#e88c73f4-d60b-4c14-8656-b34350f111e8
This updates the command so that we're explicit about the command being autowrapped in quotes, and uses single quotes for the echo so it doesn't get parsed funkily

Ran this through yamllint and confirmed that it wasn't changed via yamllint unlike the previous version (both were valid yaml though )

### DIFF
--- a/.buildkite/pipeline.testing.yml
+++ b/.buildkite/pipeline.testing.yml
@@ -9,4 +9,4 @@ env:
 steps:
   - label: ":aws: E2E tests in AWS"
     command:
-      - echo "TODO: make this work again"
+      - "echo 'TODO: make this work again'"


jgrillo-grapl (approved):

.buildkite/pipeline.testing.yml
```diff
@@ -9,4 +9,4 @@ env:
 steps:
   - label: ":aws: E2E tests in AWS"
     command:
-      - echo "TODO: make this work again"
+      - "echo 'TODO: make this work again'"
```
twunderlich-grapl, line 12: According to yamllint you NEED the outside quotes otherwise yaml interprets it as  a key value block , ie
"echo 'TODO" = "make this work again'"

PR #1253, opened 2021-11-04
Remove references to docker.io so containers can be built and pushed to cloudsmith

ghost:
This removes all remaining references to docker.io and the docker repository environment variable. 
This allows building the model deployer and node-idendifier-retry containers and pushing them to cloudsmith under the current process. Since this container is not in cloudsmith, it blocks deploys to nomad in AWS.

Build the containers locally (`make build` or `docker buildx bake --file docker-compose.build.yml`)
Then run docker image ls. The containers you just built won't be prefixed with docker.io

### DIFF
--- a/docker-compose.build.yml
+++ b/docker-compose.build.yml
@@ -66,7 +66,7 @@ services:
         - RUST_BUILD=${RUST_BUILD:-debug}
 
   model-plugin-deployer:
-    image: ${DOCKER_REGISTRY:-docker.io}/grapl/model-plugin-deployer:${TAG:-latest}
+    image: grapl/model-plugin-deployer:${TAG:-latest}
     build:
       context: src
       dockerfile: rust/Dockerfile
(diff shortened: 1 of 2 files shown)


docker-compose.build.yml
```diff
@@ -66,7 +66,7 @@ services:
         - RUST_BUILD=${RUST_BUILD:-debug}
 
   model-plugin-deployer:
-    image: ${DOCKER_REGISTRY:-docker.io}/grapl/model-plugin-deployer:${TAG:-latest}
+    image: grapl/model-plugin-deployer:${TAG:-latest}
```
christophermaier, line 69: I notice that the `node-identifier-retry` entry above is the only other one that still has this `DOCKER_REGISTRY` variable; does that also need to be removed here?

twunderlich-grapl: Yep, adding that now. Nice catch

christophermaier (approved):
Nice cleanup 👍

PR #1248, opened 2021-11-03
Grapl Pulumi constructs an API Gateway route to `grapl-web-ui`

wimax-grapl:
### Which issue does this PR correspond to?
https://github.com/grapl-security/issue-tracker/issues/665

### What changes does this PR make to Grapl? Why?
This PR depends on https://github.com/grapl-security/platform-infrastructure/pull/43

We construct an API Gateway route that forwards HTTP traffic all the way to grapl-web-ui; documentation on this pipeline is
in the api_gateway.py file.

### How were these changes tested?
image
also checked graphql endpoint:
```
curl -I -X POST https://n87ev5n1dg.execute-api.us-east-1.amazonaws.com/graphQlEndpoint/
HTTP/2 401 
date: Wed, 03 Nov 2021 23:41:48 GMT
content-type: text/plain; charset=utf-8
content-length: 38
set-cookie: actix-session=CwUZ2zRRO8F3ByybXBzBUCbB9MRuoGL8QxQZfkSo; HttpOnly; Secure; Path=/
apigw-requestid: IQGpeiDOIAMEJ2g=
```
correctly gives me a 401 unauth

PR #1259, opened 2021-11-09
Integration test Nomad jobs are a separate Pulumi project now.

wimax-grapl:
Precursor work for https://github.com/grapl-security/issue-tracker/issues/770
It's the Option 2 discussed here: https://grapl-internal.slack.com/archives/C02J5JYS92S/p1636149858045300

Previously, we decided "do we want to create these potential Nomad Job resources in Pulumi?" based on a `SHOULD_DEPLOY_[integration or e2e]_TESTS`. That was a short-term hack.

Now, we express this whole "optional resource" thing by making it a stack that requires output from the `grapl` stack.
The optionality is expressed by "well, have we `pulumi up`'d it?

I ran `make test-e2e` and `make test-integration` on my local machine. They likely don't work in AWS yet; for now, I just wanted to extract them and _then_ get to work on making it work right in AWS.


pulumi/integration_tests/__main__.py
```diff
@@ -0,0 +1,121 @@
...
+    # objects.
+    register_auto_tags({"grapl deployment": stack_name})
+
+    if not config.LOCAL_GRAPL:
+        # Set the nomad address. This can be either set as nomad:address in the config to support ssm port forwarding or
+        # taken from the nomad stack
+        nomad_config = pulumi.Config("nomad")
+        nomad_override_address = nomad_config.get("address")
+        # We prefer nomad:address to support overriding in the case of ssm port forwarding
+        nomad_server_stack = pulumi.StackReference(f"grapl/nomad/{stack_name}")
```
twunderlich-grapl, line 34: FYI, I've got a change that will DRY this logic up a bit in a PR

pulumi/integration_tests/__main__.py
```diff
@@ -0,0 +1,121 @@
...
+        nomad_address = nomad_override_address or nomad_server_stack.require_output(
+            "address"
+        )
+        nomad_provider = nomad.Provider("nomad-aws", address=nomad_address)
+
+    ##### Actual Logic
+
+    grapl_stack = GraplStack(stack_name)
```
twunderlich-grapl, line 42: I'm not sure I understand why you're putting this into an object. I get wanting to have a shorter alias for the StackReference.require_output(), but I'm not sure why you went and decided to do
`self.analyzer_bucket = output("analyzers-bucket")` etc.

pulumi/integration_tests/__main__.py
```diff
@@ -0,0 +1,121 @@
...
+
+    assert aws.config.access_key, "Appease typechecker"
+    assert aws.config.secret_key, "Appease typechecker"
```
twunderlich-grapl, line 45: In the grapl pulumi the assert aws.config.access_key is to ensure that AWS keys are set. I'm pretty sure that this will break when compared to a specific string

pulumi/integration_tests/__main__.py
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    integration_tests = NomadJob(
+        "integration-tests",
+        jobspec=Path("../../nomad/local/integration-tests.nomad").resolve(),
+        vars=integration_test_job_vars,
+    )
+
+
+def stackname_sans_prefix() -> str:
```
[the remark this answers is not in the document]
wimax-grapl: you're right! i've discovered this since

pulumi/integration_tests/__main__.py
```diff
@@ -0,0 +1,121 @@
...
+    assert (
+        len(split) == 2
+    ), f"Expected a stack prefix of {prefix}, found {real_stackname}"
+    return split[1]
+
+
+class GraplStack:
+    def __init__(self, stack_name: str) -> None:
+        ref_name = "local-grapl" if config.LOCAL_GRAPL else f"grapl/grapl/{stack_name}"
+        ref = pulumi.StackReference(ref_name)
```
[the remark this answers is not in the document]
wimax-grapl: they do, in fact! however, orgs don't work.

pulumi/integration_tests/__main__.py
wimax-grapl: cool! yeah, i was thinking of doing it myself

pulumi/integration_tests/__main__.py
wimax-grapl: This is actually an idea originally floated by cm:
https://github.com/grapl-security/grapl/pull/1248#discussion_r742908772

basically, "have a well-typed object that describes what imports we can get"

docker-compose.yml
```diff
@@ -13,10 +13,24 @@ services:
     entrypoint: ["/bin/bash", "-o", "errexit", "-o", "nounset", "-c"]
     command:
       - |
-        cd grapl
         pulumi login --local
+
+        cd grapl
         pulumi stack init local-grapl --non-interactive
         pulumi up --yes --skip-preview --stack=local-grapl
+        cd -
```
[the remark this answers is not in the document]
wimax-grapl: Thanks, these are both significantly better

pulumi/integration_tests/__main__.py
wimax-grapl: So, this logic actually changes in my upcoming PR (this works fine locally and breaks against AWS; this is what that whole `optionally-defined-env-vars` song and dance you did a while back is about)

pulumi/integration_tests/__main__.py
wimax-grapl: (I do agree it's a tad verbose, and perhaps make better sense in other contexts)

twunderlich-grapl: Ah interesting. That makes sense.

pulumi/integration_tests/__main__.py
twunderlich-grapl: ah right, ok.

twunderlich-grapl (approved):

PR #1261, opened 2021-11-10
A few steps closer to e2e tests in AWS

wimax-grapl:
grapl-security/issue-tracker#770

- Introduce the "optional-ness" of aws_endpoint / access_key / secret_key, which we already do in `grapl-core.nomad`
- Introduce artifacts tags.
- Disable integration tests against AWS for now - it needs some work.
- Publish E2E tests to Cloudsmith
- Sunset e2e-tests.build, add that to integration-tests.build 

locally:
make test-e2e
make test-integration

against my sandbox:
followed my initialization steps
pulumi up
manually run a `dispatch job`, upon which it fails because it doesn't know how to grab e2e tests with that `dev` tag (obviously)

### DIFF
--- a/nomad/local/e2e-tests.nomad
+++ b/nomad/local/e2e-tests.nomad
@@ -180,7 +180,7 @@ EOF
       driver = "docker"
 
       config {
-        image      = "${var.container_registry}grapl/e2e-tests:${var.e2e_tests_tag}"
+        image = "${var.container_registry}grapl/e2e-tests:${var.e2e_tests_tag}"
       }
 
       # This writes an env file that gets read by the task automatically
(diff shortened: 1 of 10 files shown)


nomad/local/e2e-tests.nomad
```diff
@@ -45,16 +51,19 @@ variable "schema_properties_table_name" {
 variable "aws_access_key_id" {
   type        = string
   description = "The aws access key id used to interact with AWS."
+  default     = "DUMMY_LOCAL_AWS_ACCESS_KEY_ID"
```
wimax-grapl, line 54: This is the same pattern we follow in the other nomad files; i don't love it, but I have much bigger fish to fry right now.

christophermaier (approved):

PR #1266, opened 2021-11-12
Propagate $TAG into Nomad

wimax-grapl:
ongoing attempt to solve grapl/merge

the current issue is that we'd build Docker images with a TAG of `ci-pipeline`, but Nomad assumed some things were still `:dev`.
- It never inspected `$TAG` at all
- e2e and integration tests were straight-up hardcoded to `dev`

This changes every place we consume a tag to use `def version_tag` (ctrl-f for it). 

ran `make test-integration` and `make test-e2e` and also variants with `TAG=derp make test-integration`
seemed to work

### DIFF
--- a/local-grapl.env
+++ b/local-grapl.env
@@ -117,7 +117,3 @@ GRAPL_TEST_USER_NAME=${DEPLOYMENT_NAME}-grapl-test-user
 
 # Locally this should be empty so that Nomad picks up local images
 DOCKER_REGISTRY=""
-
-# Locally this should be pinned to `dev`, and not read in ${TAG} or anything.
-# (Also, FYI, Nomad cannot pick up local images named `latest`.)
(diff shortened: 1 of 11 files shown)


local-grapl.env
```diff
@@ -116,6 +116,4 @@ GRAPL_TEST_USER_NAME=${DEPLOYMENT_NAME}-grapl-test-user
 # Determines the registry to build and push against.
 
 # Locally this should be empty so that Nomad picks up local images
-DOCKER_REGISTRY=""
-# Nomad won't pick up local images that are tagged with latest
-TAG="${TAG:-dev}"
+DOCKER_REGISTRY=""
```
wimax-grapl, line 119: To be clear, the `TAG` is already defaulted to dev in Makefile.

colin-grapl (approved):

PR #1267, opened 2021-11-15
Enable Consul ACLs

ghost:
https://github.com/grapl-security/issue-tracker/issues/783

This sets up consul ACL roles, policies and tokens

1. Consul ACL bootstrapping is done via a dynamic provider
2. Consul policies are defined as hcl files inside pulumi/consul-acls. These are read and generated into policies automatically.
3. We then create roles and assign a specific policy to them (roles supports multiple policies but for the moment we're just doing a 1:1)
4. These roles are then assigned to a token. These tokens can be used to log into the Consul UI as a meat-user or can be used by servers/services etc.
5. For the moment, we grant the anonymous token a number of permissions

### Why is this in the grapl repo?
1. With this being in the grapl repo its possible to test ACL changes locally.
2. Currently the private repo that sets up our stack in AWS doesn't connect to consul itself so it doesn't make sense to have this be part of the consul project. We are already connecting to consul in the grapl repo. Long-term we may want to move this into its own private project that's run after consul/nomad infra is set up, but before we deploy nomad jobs

For local grapl: `make up-detach`
For an AWS sandbox which had previously had consul acls manually bootstrapped:
Running via pulumi up.

### DIFF
--- a/pulumi/consul-acl-policies/consul-agent.hcl
+++ b/pulumi/consul-acl-policies/consul-agent.hcl
@@ -1,0 +1,19 @@
+# Policy used for consul agents, in particular nomad agents.
+
+# Allow the agent to register its own node in the Catalog and update its network coordinates, update node health checks
+# and have write access to its own agent config files.
+node "" {
+  policy = "write"
+}
+
+# Allows the agent to detect and diff services registered to itself. This is used during
+# anti-entropy to reconcile difference between the agents knowledge of registered
+# services and checks in comparison with what is known in the Catalog.
+service_prefix "" {
+  policy = "read"
+}
+
+# only necessary if using dns prepared queries
+query_prefix "" {
+  policy = "read"
+}
--- a/pulumi/consul-acl-policies/consul-agent.hcl
+++ b/pulumi/consul-acl-policies/consul-agent.hcl
@@ -2,6 +2,7 @@
 
 # Allow the agent to register its own node in the Catalog and update its network coordinates, update node health checks
 # and have write access to its own agent config files.
+# Currently we use a wildcard match via "".
 node "" {
   policy = "write"
 }
--- a/pulumi/consul-acl-policies/consul-agent.hcl
+++ b/pulumi/consul-acl-policies/consul-agent.hcl
@@ -10,7 +10,7 @@ node "" {
 # anti-entropy to reconcile difference between the agents knowledge of registered
 # services and checks in comparison with what is known in the Catalog.
 service_prefix "" {
-  policy = "read"
+  policy = "write"
 }
 
 # only necessary if using dns prepared queries
--- a/pulumi/grapl/__main__.py
+++ b/pulumi/grapl/__main__.py
@@ -1,6 +1,6 @@
 import sys
 from pathlib import Path
-from typing import Set
+from typing import Mapping, Set
 
 from typing_extensions import Final
 
@@ -23,6 +23,7 @@ from infra.bucket import Bucket
 from infra.cache import Cache
 from infra.consul_acl_policies import ConsulAclPolicies
 from infra.consul_intentions import ConsulIntentions
+from infra.docker_image_tag import version_tag
 from infra.get_hashicorp_provider_address import get_hashicorp_provider_address
 from infra.grapl_consul_acls import GraplConsulAcls
 
@@ -44,6 +45,27 @@ import pulumi
 
 def _get_subset(inputs: NomadVars, subset: Set[str]) -> NomadVars:
     return {k: inputs[k] for k in subset}
+
+
+def grapl_core_docker_image_tags(
+    artifacts: Mapping[str, str], require_artifact: bool = False
+) -> NomadVars:
+    # partial apply some repeated args
+    version_tag_alias = lambda key: version_tag(key, artifacts, require_artifact)
+
+    return dict(
+        analyzer_dispatcher_tag=version_tag_alias("analyzer-dispatcher"),
+        analyzer_executor_tag=version_tag_alias("analyzer-executor"),
+        dgraph_tag="latest",
+        engagement_creator_tag=version_tag_alias("engagement-creator"),
+        graph_merger_tag=version_tag_alias("graph-merger"),
+        graphql_endpoint_tag=version_tag_alias("graphql-endpoint"),
+        model_plugin_deployer_tag=version_tag_alias("model-plugin-deployer"),
+        node_identifier_tag=version_tag_alias("node-identifier"),
+        osquery_generator_tag=version_tag_alias("osquery-generator"),
+        sysmon_generator_tag=version_tag_alias("sysmon-generator"),
+        web_ui_tag=version_tag_alias("grapl-web-ui"),
+    )
 
 
 def main() -> None:
@@ -203,6 +225,7 @@ def main() -> None:
             vars=dict(
                 **grapl_core_job_vars_inputs,
                 **nomad_inputs,
+                **grapl_core_docker_image_tags({}),
             ),
         )
 
@@ -214,18 +237,20 @@ def main() -> None:
 
         provision_vars = _get_subset(
             dict(
+                provisioner_tag=version_tag("provisioner", {}, require_artifact=False),
                 **grapl_core_job_vars_inputs,
                 **nomad_inputs,
             ),
             {
-                "_aws_endpoint",
                 "aws_access_key_id",
                 "aws_access_key_secret",
+                "_aws_endpoint",
                 "aws_region",
                 "deployment_name",
+                "provisioner_tag",
                 "rust_log",
-                "schema_table_name",
                 "schema_properties_table_name",
+                "schema_table_name",
                 "test_user_name",
                 "user_auth_table",
             },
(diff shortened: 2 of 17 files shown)


pulumi/grapl/__main__.py
```diff
@@ -66,6 +77,90 @@ def grapl_core_docker_image_tags(
...
+
+        # As is customary in Pulumi, all inputs are available as
+        # outputs (this must include our injected Vault address and
+        # namespace values, which will be required when we try to
+        # delete the resource).
+        #
+        outs = cast(MutableMapping[str, Any], deepcopy(inputs))
+        # ... plus the accessor id and secret token
+        outs["id"] = secret_id
+        outs["accessor_id"] = accessor_id
```
twunderlich-grapl, line 125: TODO:
Figure out why accessor_id isn't showing up when exported

pulumi/grapl/__main__.py
```diff
@@ -66,6 +77,90 @@ def grapl_core_docker_image_tags(
...
+        outs["secret_token"] = secret_id
```
twunderlich-grapl, line 126: TODO mark as a secret

pulumi/grapl/__main__.py
```diff
@@ -293,17 +418,41 @@ def main() -> None:
         artifacts = pulumi_config.require_object("artifacts")
 
         # Set custom provider with the address set
-        consul_provider = get_hashicorp_provider_address(consul, "consul", consul_stack)
+        # consul_provider = get_hashicorp_provider_address(consul, "consul", consul_stack)
```
twunderlich-grapl, line 421: TODO update get_hashicorp_provider_address to use kwargs

pulumi/infra/grapl_consul_acls.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Attach ACL UI RO to the anonymous token for convenience
+        consul.AclTokenRoleAttachment(
+            f"{name}-attach-ui-ro-to-anon-token",
+            role=ui_ro_role.name,
+            token_id=self.ANONYMOUS_TOKEN_ID,
+            opts=pulumi.ResourceOptions(parent=self),
+        )
+
+        # Ugly Hack Warning: Since there's no api to assign a token to a consul agent, we assign to the anonymous token
```
twunderlich-grapl, line 75: TODO figure out how to set tokens on consul agents. There's a command https://www.consul.io/commands/acl/set-agent-token for it, but no api I can find yet.

pulumi/grapl/__main__.py
twunderlich-grapl: Done

pulumi/grapl/__main__.py
twunderlich-grapl: Done

pulumi/consul-acl-policies/consul-agent.hcl
```diff
@@ -0,0 +1,19 @@
+# Policy used for consul agents, in particular nomad agents.
+
+# Allow the agent to register its own node in the Catalog and update its network coordinates, update node health checks
+# and have write access to its own agent config files.
+node "" {
```
christophermaier, line 5: Are all these ACL rules supposed to be targeted to `""`? Is that some kind of wild-card?

pulumi/grapl/__main__.py
```diff
@@ -66,6 +77,104 @@ def grapl_core_docker_image_tags(
...
+        elif response.status_code == requests.codes.forbidden:
+            # We've already run the bootstrap process so instead let's grab the token from the config
+            if inputs["secret_token"] is not None:
+                secret_id = inputs["secret_token"]
+            else:
+                raise Exception(
+                    "If Consul ACL Bootstrapping is complete you must set consul:token with the token"
+                )
+            # Create a fake static id since we don't have the accessor_id available
+            bootstrap_id = "fake-id-since-previously-bootstrapped"
```
christophermaier, line 131: How does this "fake" data cascade through the system? Is this going to cause problems?

pulumi/grapl/__main__.py
```diff
@@ -66,6 +77,104 @@ def grapl_core_docker_image_tags(
...
+    # If this is running on a previously bootstrapped cluster, we require a token be configured and create a fake id
+    def create(self, inputs: Mapping[str, Any]) -> CreateResult:
+        response = requests.put(f"{inputs['consul_address']}/v1/acl/bootstrap")
+        if response.status_code == requests.codes.ok:
+            bootstrap_id = response.json()["AccessorID"]
+            secret_id = response.json()["SecretID"]
+        elif response.status_code == requests.codes.forbidden:
+            # We've already run the bootstrap process so instead let's grab the token from the config
+            if inputs["secret_token"] is not None:
+                secret_id = inputs["secret_token"]
```
christophermaier, line 125: Does this token really need to be set in configuration, or can we use an environment variable? It should be short-lived, correct?

pulumi/grapl/__main__.py
```diff
@@ -66,6 +77,104 @@ def grapl_core_docker_image_tags(
...
+                )
+            # Create a fake static id since we don't have the accessor_id available
+            bootstrap_id = "fake-id-since-previously-bootstrapped"
+        else:
+            response.raise_for_status()
+
+        # As is customary in Pulumi, all inputs are available as
+        # outputs (this must include our injected Vault address and
+        # namespace values, which will be required when we try to
+        # delete the resource).
```
christophermaier, line 138: A bit overzealous with the copy/paste 😜

pulumi/grapl/__main__.py
```diff
@@ -66,6 +77,104 @@ def grapl_core_docker_image_tags(
...
+    def delete(self, id: str, props: Mapping[str, Any]) -> None:
+        pass
+
+    # The function that determines if an existing resource whose inputs were
+    # modified needs to be updated or entirely replaced
+    def diff(
+        self, id: str, old_inputs: Mapping[str, Any], new_inputs: Mapping[str, Any]
+    ) -> DiffResult:
+        """
+        This is a no-op since consul acl bootstrapping can only be run once
```
christophermaier, line 155: If it's a no-op, is a `diff` implementation needed?

pulumi/grapl/__main__.py
```diff
@@ -198,11 +307,47 @@ def main() -> None:
 
         # This does not use a custom Provider since it will use either a consul:address set in the config or default to
         # http://localhost:8500. This also applies to the NomadJobs defined for LOCAL_GRAPL.
+        bootstrap = ConsulAclBootstrap(
+            name="consul-acl-bootstrap",
+            consul_address=pulumi.Config("consul").get("address")
+            or "http://localhost:8500",
+            secret_token=pulumi.Config("consul").get("token"),
```
christophermaier, line 314: Should this be `Config(...).get_secret(...)` instead? You may not need the `additional_secret_outputs` option in that case.

pulumi/grapl/__main__.py
```diff
@@ -198,11 +307,47 @@ def main() -> None:
...
+        grapl_acls = GraplConsulAcls(
+            "grapl",
+            policies=consul_acl_policies.policies,
+            opts=pulumi.ResourceOptions(provider=consul_provider),
+        )
+        pulumi.export("ui_read_only_token", grapl_acls.ui_read_only_token.id)
+        pulumi.export("ui_read_write_token", grapl_acls.ui_read_write_token.id)
+        pulumi.export(
+            "default_consul_agent_token", grapl_acls.default_consul_agent_token.id
+        )
```
christophermaier, line 341: Do all these need to be handled as secrets?

pulumi/infra/grapl_consul_acls.py
```diff
@@ -0,0 +1,84 @@
+from typing import Dict, Optional
+
+import pulumi_consul as consul
+
+import pulumi
+
+
+class GraplConsulAcls(pulumi.ComponentResource):
+
+    ANONYMOUS_TOKEN_ID = "00000000-0000-0000-0000-000000000002"
```
christophermaier, line 10: It would be good to add some documentation here with a link to https://www.consul.io/docs/security/acl/acl-system#builtin-tokens; I thought this was something you were making up yourself.

Documenting why we're using the anonymous token would also be useful.

pulumi/infra/grapl_consul_acls.py
christophermaier: Would something like https://www.consul.io/api-docs/acl/tokens#update-a-token work?

Might be worth asking on the Hashicorp forums what Terraform people do in such cases.

pulumi/grapl/__main__.py
christophermaier: So... if we provide a token, then we've basically already bootstrapped, correct? In that case, we just use that master token to do all the other provisioning.

Isn't the bootstrap process more of a cluster initiation step? It seems odd that it's baked into this "client" project. Also, the token produced has unlimited capabilities, which suggests that it should be protected centrally, and that we should probably have another mechanism by which to supply a more tightly-scoped "Grapl-only" token for this purpose.

pulumi/consul-acl-policies/consul-agent.hcl
twunderlich-grapl: Yes, I'll add comments for this since its unusual

pulumi/grapl/__main__.py
twunderlich-grapl: You're correct this is unnecessary

PR #1269, opened 2021-11-16
Pull Pulumi artifacts from testing repository

christophermaier:
First, we add a promotion step to our merge pipeline; if all our
containers build successfully, we'll use our new Cloudsmith Buildkite
plugin to promote the new artifacts out of the `raw` repository into
the `testing` one. Anything left in the `raw` repository will be
deleted over time, because anything that remains in there more than a
few minutes is by definition part of a failing commit.

Then, we parameterize our Pulumi stack with the value of this
repository, so we can instruct Nomad where to pull the containers
from. Previously, we specified this using a combination of "registry"
and "repo". However, because of how the naming in Cloudsmith works, we
always had to interject a `grapl/` into the resulting address. (That
is, `registry` was `docker.cloudsmith.io` and `repository` was `raw`,
with a resulting value of
`docker.cloudsmith.io/grapl/raw/<MY_IMAGE_NAME>:<TAG>`.

The fact that we specified a name of `grapl/<MY_IMAGE_NAME>` in our
Docker Compose build files complicated things. We did this to adhere
to Docker Hub conventions, even though we don't ever push these images
to Docker Hub. Therefore, we had to take special care to assemble the
correct name of an image, depending on whether we were deploying into
AWS or "local Grapl".

In an attempt to simplify things, we now just specify everything up to
the image name (so, e.g., `docker.cloudsmith.io/grapl/testing`) in a
Pulumi stack configuration variable, named
`container_repository`. Though it's technically a combination of a
registry and part of a repository, it does make things a bit easier to
reason about, and to configure.

Since we pin our artifacts in Pulumi using just the bare container
name and version, these will continue to work as before (because the
same names and versions will be present in the new container repository.

To make this work with local containers, we just remove the `grapl/`
prefix from the image name, making them the same as what we expect
elsewhere. This still works, and is actually a little bit safer
because now it will be impossible to accidentally push them to Docker
Hub. (I renamed our local `localstack` image to
`localstack-grapl-fork` to make it more obvious that we are using
something different than the "normal" localstack container.)

### DIFF
--- a/pulumi/grapl/__main__.py
+++ b/pulumi/grapl/__main__.py
@@ -306,8 +306,7 @@ def main() -> None:
             # The vars with a leading underscore indicate that the hcl local version of the variable should be used
             # instead of the var version.
             _redis_endpoint=cache.endpoint,
-            container_registry="docker.cloudsmith.io/",
-            container_repo="raw/",
+            container_repository=f"{config.container_repository()}/",
             # TODO: consider replacing with the previous per-service `configurable_envvars`
             rust_log="DEBUG",
             **nomad_inputs,
@@ -333,8 +332,7 @@ def main() -> None:
             dict(
                 # The vars with a leading underscore indicate that the hcl local version of the variable should be used
                 # instead of the var version.
-                container_registry="docker.cloudsmith.io/",
-                container_repo="raw/",
+                container_repository=f"{config.container_repository()}/",
                 # TODO: consider replacing with the previous per-service `configurable_envvars`
                 rust_log="DEBUG",
                 provisioner_tag=version_tag(
@@ -344,8 +342,7 @@ def main() -> None:
             ),
             {
                 "aws_region",
-                "container_registry",
-                "container_repo",
+                "container_repository",
                 "deployment_name",
                 "provisioner_tag",
                 "rust_log",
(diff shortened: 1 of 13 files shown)


pulumi/grapl/__main__.py
```diff
@@ -306,8 +306,7 @@ def main() -> None:
             # The vars with a leading underscore indicate that the hcl local version of the variable should be used
             # instead of the var version.
             _redis_endpoint=cache.endpoint,
-            container_registry="docker.cloudsmith.io/",
-            container_repo="raw/",
+            container_repository=f"{config.container_repository()}/",
```
twunderlich-grapl, line 309: Is defining a pulumi.config.get as a function in infra config our preferred pattern?

twunderlich-grapl (approved):
Looking good

pulumi/grapl/__main__.py
christophermaier: 🤔 I can probably dispense with the function... I think it actually did some more stuff in an earlier iteration, but not anymore.

christophermaier: On second thought, I'll leave it as is, since we call it in a couple places.

PR #1272, opened 2021-11-19
Promote packages to the "validated" repository in Cloudsmith

christophermaier:
Once packages make it through the testing pipeline, we can copy them
into the "validated" repository, which will be a holding place for
packages that can plausibly be released to production.

This repository has retention rules set up (out of band; not in this
PR) so that we only keep the 5 most recent versions of each package.

Additionally, we promote the packages via copy and not move because
once we're able to selectively build based on "what changed", we'll be
introducing subsets of new packages. If we were to move them all out,
then nothing would be available for the next run-through of the
pipelines.

In the future, we may be able to incorporate the "extract the current
artifacts from the Pulumi stack" logic into the Pulumi plugin, or
probably the artifacts plugin; for now, though, we'll just add it to
the pipeline directly as we work out the best patterns.

### DIFF
--- a/.buildkite/pipeline.provision.yml
+++ b/.buildkite/pipeline.provision.yml
@@ -45,6 +45,53 @@ steps:
 
   - wait
 
+  # This makes use of the Pulumi binary installed on our agents, and
+  # uses it *only* to extract information from the Pulumi YAML stack
+  # file.
+  #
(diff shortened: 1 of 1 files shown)


jgrillo-grapl (approved):
👏

PR #1285, opened 2021-11-27
Add skeleton of plugin-work-queue service

colin-grapl:
712

Adds a skeleton for the plugin-work-queue service. No business logic implemented.

No additional tests currently.


src/rust/plugin-work-queue/src/client.rs
```diff
@@ -0,0 +1,108 @@
+#![allow(warnings)]
```
christophermaier, line 1: Any reason to leave this in? Since this is a completely new file, we should be able to eliminate any warnings up front.

If it's because we've got a bunch of essentially stubbed-out implementations and there are a bunch of legitimate warnings, would it be possible to explicitly ignore those specific warnings at the site? That would presumably make it easier to remove them once the implementations are filled in. I know that's potentially a lot of annotation, but these blanket exceptions at the top of the file are really easy to forget about, and can mask a lot of other things that we legitimately don't want to allow.

src/rust/plugin-work-queue/src/server.rs
```diff
@@ -0,0 +1,144 @@
...
+    Status,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum PluginWorkQueueError {}
+
+pub struct PluginWorkQueue {}
+
+impl PluginWorkQueue {
+    async fn _put_execute_generator(
```
christophermaier, line 40: Can you talk more about this apparent underscore convention that's being used for these method names, as well as for the imported types above? It appears to be some kind of "private implementation" convention (to distinguish between the "raw" Protobuf API and a transport-independent business logic representation). If that's the case, it's a bit confusing, because it's bringing Python conventions into a Rust codebase, and also running up against Rust's use of a leading underscore to denote unused variables. 

For the types, could we do something like naming them `FooProto` instead of `_Foo`? And for the functions, could they be called explicitly, like `PluginWorkQueue::put_execute_generator(<THE_INSTANCE>, request)`?

src/rust/plugin-work-queue/src/client.rs
colin-grapl: It should be totally removed, it was just for dev purposes.

src/rust/plugin-work-queue/src/server.rs
colin-grapl: I can use `FooProto` instead of `_Foo`, no problem.

I can see if ucfs is necessary, it may be that Rust can understand how to dispatch based on the argument type.

christophermaier (approved):

PR #1289, opened 2021-11-30
Lint Dockerfiles with Hadolint

christophermaier:
Pants can lint Dockerfiles with Hadolint now... so let's use that!

(Note: we are _only_ linting with Pants; we are not changing how we build our Dockerfiles)


jgrillo-grapl (approved):
👏

PR #1290, opened 2021-11-30
Feed fully qualified Docker Image URLs into the Nomad file instead of piecemeal tags and container_repos

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/697

Previously we'd feed a `container_repository` and `tag`s into the Nomad file and do some variable interpolation inside Nomad to build the Docker Image ID ("okay, where do we find the docker image?")

I'm lifting this work up to the Pulumi level per the attached issue.
Main motivation for picking this up now: we will soon potentially be in a situation where some images come from the `grapl/raw` cloudsmith repo, and some come from the `grapl/$USERNAME` repo (if I wanted to upload 1 image for testing). This is much easier to reason about at the Pulumi level. 

against CI

### DIFF
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -18,14 +18,6 @@ variable "container_images" {
 EOF
 }
 
-# TODO delete
-variable "container_repository" {
-  type        = string
-  default     = ""
-  description = "The container registry in which we can find Grapl services. Requires a trailing / if not empty string"
-}
-
-
 variable "aws_access_key_id" {
   type        = string
   default     = "DUMMY_LOCAL_AWS_ACCESS_KEY_ID"
@@ -300,7 +292,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["dgraph"]}"
+        image = var.container_images["dgraph"]
         args = [
           "dgraph",
           "zero",
@@ -364,7 +356,7 @@ job "grapl-core" {
         driver = "docker"
 
         config {
-          image = "${var.container_images["dgraph"]}"
+          image = var.container_images["dgraph"]
           args = [
             "dgraph",
             "zero",
@@ -468,7 +460,7 @@ job "grapl-core" {
         driver = "docker"
 
         config {
-          image = "${var.container_images["dgraph"]}"
+          image = var.container_images["dgraph"]
           args = [
             "dgraph",
             "alpha",
@@ -581,7 +573,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["graph-merger"]}"
+        image = var.container_images["graph-merger"]
       }
 
       # This writes an env files that gets read by nomad automatically
@@ -637,7 +629,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["node-identifier"]}"
+        image = var.container_images["node-identifier"]
       }
 
       template {
@@ -677,7 +669,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["node-identifier-retry"]}"
+        image = var.container_images["node-identifier-retry"]
       }
 
       template {
@@ -712,7 +704,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["analyzer-dispatcher"]}"
+        image = var.container_images["analyzer-dispatcher"]
       }
 
       template {
@@ -751,7 +743,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["analyzer-executor"]}"
+        image = var.container_images["analyzer-executor"]
       }
 
       template {
@@ -810,7 +802,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["engagement-creator"]}"
+        image = var.container_images["engagement-creator"]
       }
 
       template {
@@ -863,7 +855,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["graphql-endpoint"]}"
+        image = var.container_images["graphql-endpoint"]
         ports = ["graphql-endpoint-port"]
       }
 
@@ -920,7 +912,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["model-plugin-deployer"]}"
+        image = var.container_images["model-plugin-deployer"]
         ports = ["model-plugin-deployer"]
       }
 
@@ -952,7 +944,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["web-ui"]}"
+        image = var.container_images["web-ui"]
         ports = ["web-ui-port"]
       }
 
@@ -1002,7 +994,7 @@ job "grapl-core" {
       driver = "docker"
 
       config {
-        image = "${var.container_images["sysmon-generator"]}"
+        image = var.container_images["sysmon-generator"]
       }
 
       template {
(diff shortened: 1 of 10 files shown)


nomad/grapl-core.nomad
```diff
@@ -3,10 +3,19 @@ variable "rust_log" {
   description = "Controls the logging behavior of Rust-based services."
 }
 
-variable "container_repository" {
-  type        = string
-  default     = ""
-  description = "The container registry in which we can find Grapl services. Requires a trailing / if not empty string"
+variable "container_images" {
+  type        = map(string)
```
wimax-grapl, line 7: You'll note I've made the decision to stop using bespoke variables per-service and instead just pipe in a Map[str, str]. I'm willing to revisit this decision but, honestly, I think this makes things a bit cleaner, and lets us recycle the `var.container_images` pattern across files.

nomad/local/start_detach.sh
```diff
@@ -19,7 +19,7 @@ ensure_cros_bridge_networking_workaround() {
         if ! grep -q "_/bridge.ko" "$module_filepath"; then
             echo "It looks like you're on ChromeOS, but haven't installed the Nomad bridge networking workaround."
             # shellcheck source-path=SCRIPTDIR
-            source "$(dirname "${BASH_SOURCE[0]}")/../../etc/chromeos/lib/installs.sh"
+            source "../../etc/chromeos/lib/installs.sh"
```
wimax-grapl, line 22: (this was actually failing previously)

pulumi/infra/docker_images.py
```diff
@@ -33,3 +35,19 @@ def version_tag(
...
 
     return _DEFAULT_TAG
+
+
+class DockerImageIdBuilder:
+    def __init__(self, container_repository: Optional[str]) -> None:
+        if container_repository is not None:
+            self.container_repository: str = f"{container_repository}/"
+        else:
+            self.container_repository = ""
```
christophermaier, line 45: How about:
```python
self.container_repository = f"{container_repository}/" if container_repository else ""
```

pulumi/integration_tests/__main__.py
```diff
@@ -1,30 +1,57 @@
...
+) -> str:
+    """
+    Build a map of "task name -> docker image URL".
+    See the .nomad file's var `container_images` for further documentation.
+    """
+    img_id_builder = DockerImageIdBuilder(config.container_repository())
+
+    def build_img_id_with_tag(image_name: str) -> str:
+        # A shortcut to grab the version tag from the artifacts map and build a
+        # Cloudsmith docker image url out of it.
```
christophermaier, line 32: We're not _necessarily_ building a Cloudsmith image identifier here. Also, there's actually nothing Cloudsmith-specific about any of this logic.

pulumi/grapl/__main__.py
```diff
@@ -41,24 +42,37 @@ def _get_subset(inputs: NomadVars, subset: Set[str]) -> NomadVars:
...
+    """
+    img_id_builder = DockerImageIdBuilder(config.container_repository())
+
+    def build_img_id_with_tag(image_name: str) -> str:
+        # A shortcut to grab the version tag from the artifacts map and build a
+        # Cloudsmith docker image url out of it.
+        tag = version_tag(image_name, artifacts, require_artifact)
+        return img_id_builder.build(image_name=image_name, tag=tag)
+
+    return json.dumps(
```
christophermaier, line 60: I think it'd be a bit more generic and decoupled to have this function return the `Mapping[str, str]`, and convert to a JSON string at the point of consumption that requires it to actually be a string.

pulumi/grapl/__main__.py
```diff
@@ -41,24 +42,37 @@ def _get_subset(inputs: NomadVars, subset: Set[str]) -> NomadVars:
...
-        node_identifier_tag=version_tag_alias("node-identifier"),
-        osquery_generator_tag=version_tag_alias("osquery-generator"),
-        sysmon_generator_tag=version_tag_alias("sysmon-generator"),
-        web_ui_tag=version_tag_alias("grapl-web-ui"),
+) -> str:
+    """
+    Build a map of "task name -> docker image URL".
+    See the .nomad file's var `container_images` for further documentation.
+    """
+    img_id_builder = DockerImageIdBuilder(config.container_repository())
```
christophermaier, line 52: Is there any reason to have this logic in a bare function in `__main__.py`, rather than have it be a method on `DockerImageIdBuilder` itself? It would be nice to put all the logic together into a single place.

pulumi/integration_tests/__main__.py
```diff
@@ -1,30 +1,57 @@
...
 from infra.autotag import register_auto_tags
-from infra.docker_image_tag import version_tag
+from infra.docker_images import DockerImageIdBuilder, version_tag
 from infra.nomad_job import NomadJob, NomadVars
 from infra.quiet_docker_build_output import quiet_docker_output
 
 import pulumi
 
 
+def _container_images(
```
christophermaier, line 21: The fact that this same logic is repeated in both `__main__.py` files is a further argument for consolidating it.

nomad/local/start_detach.sh
christophermaier: How was it failing? How was CI (presumably) missing this failure? How does this change fix it?

wimax-grapl: The failure was that the `source` assumed we were in a given directory, but the `cd $THIS_DIR` earlier in the script busts that assumption. The failure looked as follows:

```
wimax@graptop:~/src/grapl$ ./nomad/local/start_detach.sh 
It looks like you're on ChromeOS, but haven't installed the Nomad bridge networking workaround.
./nomad/local/start_detach.sh: line 23: ./nomad/local/../../etc/chromeos/lib/installs.sh: No such file or directory

(note: in this example I executed it manually with a `./nomad/local/start_detach.sh` instead of a `make up-detach`, but it's the same failure)
```

Since the script _already has us in that ./nomad/local directory_, obviously we no longer need to refer to PWD with that prefix.

to answer why CI missed it:
the failure case here happens if:
- you're on chrome OS (so, obviously, already this wouldn't happen in CI ever)
- the `_bridge.ko` file here didn't exist previously

pulumi/infra/docker_images.py
wimax-grapl: done

pulumi/integration_tests/__main__.py
wimax-grapl: yep, this comment is a relic of when this class was called CloudsmithUrlBuilder :)

nomad/local/start_detach.sh
christophermaier: 🤔 Rather than having assumptions like this baked into individual functions, I think it may be easier to reason about by having the functions assume they're being called relative to the root of the repository, and having any `cd` logic handled at a higher level, as seen here: https://github.com/grapl-security/grapl/blob/58d218b1cd47a374e2374f4bb926e2a66954f16d/bin/graplctl-pulumi.sh#L108-L127

wimax-grapl: (I think I may just revert this change and handle it in another PR.)

christophermaier: Sounds good 👍

pulumi/grapl/__main__.py
wimax-grapl: I've moved this logic to NomadJob; honestly this should be handled for us by the nomad-pulumi plugin, but the NomadJob abstraction is a good place to handle it.

pulumi/grapl/__main__.py
wimax-grapl: Yeah, I can do that.

pulumi/integration_tests/__main__.py
wimax-grapl: done

wimax-grapl:
okay, I think everything's been addressed. (also I _really_ like my consolidation of documentation around what a Docker Image ID is on the NewType)

christophermaier:
> I _really_ like my consolidation of documentation around what a Docker Image ID is on the NewType

#humblebrag 😂

christophermaier (approved):
I like it... it's really nice to see all that logic embedded in the Nomad jobs just disappear!

👍 👍 👍

PR #1294, opened 2021-12-03
E2E in CI automation. (It likely is not repeatable until we have org IDs)

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/770

- Add integration-tests to the pipeline.provision.yml. 
- Do the create_rc inject-docker-artifacts thing to grapl/integration-tests/testing
- Add run-paramaterized-job to pipeline.testing.yml
- Refactor the placement of some Nomad files (that were previously in nomad/local, but no longer are necessarily for local)

Hah. So. I actually have no idea how to test this besides shipping it! 
**Also, I have it as a "soft fail" because I don't believe the test is currently repeatable**, as each run would pollute the currently-running dgraph/dynamo database. We need to either `pulumi destroy` or introduce orgIDs. So I may just defer the actual "running this in CI" part (where we invoke `run_parameterized_job`).


.buildkite/pipeline.testing.yml
```diff
@@ -9,4 +9,5 @@ env:
 steps:
   - label: ":aws: E2E tests in AWS"
     command:
-      - "echo 'TODO: make this work again'"
+      - nomad/bin/run_parameterized_job.sh e2e-tests 6
+    soft_fail: true
```
wimax-grapl, line 13: ^ notable

.buildkite/pipeline.provision.yml
```diff
@@ -14,6 +14,10 @@ steps:
           command: preview
           project_dir: pulumi/grapl
           stack: grapl/testing
+      - grapl-security/pulumi#v0.1.2:
```
christophermaier, line 17: This will work, because the plugin runs `pulumi` in its [command hook](https://github.com/grapl-security/pulumi-buildkite-plugin/blob/main/hooks/command), and a Buildkite job can only have [one command](https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks).

You'll need to specify this second run in a job of its own. If it depends on `grapl/testing` having run first, then that job will need to use [depends_on](https://buildkite.com/docs/pipelines/dependencies#defining-explicit-dependencies) or [wait](https://buildkite.com/docs/pipelines/wait-step) to reflect that.

.buildkite/pipeline.provision.yml
```diff
@@ -14,6 +14,10 @@ steps:
...
+          command: preview
+          project_dir: pulumi/integration_tests
+          stack: grapl/integration-tests/testing
```
christophermaier, line 20: We should either use the 3-part fully-qualified stack identifier everywhere, or change this to `grapl/testing`. Specifying the `project_dir` does the job of setting the project. Having 2- and 3-part identifiers mixed together is likely to be confusing to new people ("Is this difference significant? _Have I been doing it wrong the whole time?_").

(I suspect this is probably only an issue in this repository, where we have a project that is the same name as our Pulumi organization, leading to some first-glance confusion: "Wait, I thought we were in the `integration-tests` project; why does it say `grapl`?... _oooooh right_, that's the _organization_ name.")

.buildkite/pipeline.provision.yml
```diff
@@ -29,6 +33,10 @@ steps:
           command: update
           project_dir: pulumi/grapl
           stack: grapl/testing
+      - grapl-security/pulumi#v0.1.2:
```
christophermaier, line 36: Same comments apply here.

.buildkite/scripts/create_new_rc.sh
```diff
@@ -12,6 +12,7 @@ source .buildkite/shared/lib/rc.sh
 # repository.
 readonly -a STACKS=(
     grapl/testing
+    grapl/integration-tests/testing
```
christophermaier, line 15: Based on how the [similar code](https://github.com/grapl-security/pipeline-infrastructure/blob/7d8ae5f93b9ae08b99fde70fe4246c178ee2373d/.buildkite/scripts/create_new_rc.sh#L15-L20) works in `pipeline-infrastructure`, I suspect this needs to be `integration-tests/testing`.

I really need to extract this code to a plugin 🤦

.buildkite/pipeline.testing.yml
christophermaier: Probably merits a comment in the code.

.buildkite/pipeline.provision.yml
wimax-grapl: i assume this is supposed to be a "this wont work"

christophermaier: 🤦 Ugh, correct... this will _not_ work.

christophermaier (approved):
:shipit:

PR #1297, opened 2021-12-05
Add ability to format BUILD files with black

ghost:
https://github.com/grapl-security/issue-tracker/issues/794

New in pants 2.8 is the ability to format BUILD files with black.
Unlike most formatting, this has its own command instead of being added to pants fmt
because it cannot use targets. See https://www.pantsbuild.org/docs/reference-update-build-files

Ran make format

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -357,8 +357,12 @@ format-prettier: build-formatter ## Reformat js/ts/yaml
 format-hcl: ## Reformat all HCLs
 	${NONROOT_DOCKER_COMPOSE_CHECK} hcl-format
 
-.PHONY: format
-format: format-python format-shell format-prettier format-rust format-hcl ## Reformat all code
+.PHONY: format-build
+format-build: ## Reformat BUILD files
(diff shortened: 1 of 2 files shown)


wimax-grapl (approved):

Makefile
```diff
@@ -357,8 +357,12 @@ format-prettier: build-formatter ## Reformat js/ts/yaml
 format-hcl: ## Reformat all HCLs
 	${NONROOT_DOCKER_COMPOSE_CHECK} hcl-format
 
-.PHONY: format
-format: format-python format-shell format-prettier format-rust format-hcl ## Reformat all code
+.PHONY: format-build
+format-build: ## Reformat BUILD files
+	./pants update-build-files
```
christophermaier, line 362: This will do more than just format build files, though. I think we need to run `.pants update-build-files --no-update-build-files-fix-safe-deprecations`.

See `./pants help update-build-files` for more.

PR #1301, opened 2021-12-07
Coerce --dump-agent-logs from str to bool, woops

wimax-grapl:
Relates to https://github.com/grapl-security/issue-tracker/issues/770

e2e tests worked fine, but dump_artifacts still tried to download agent logs. Turns out I was literally feeding in a string "False".

### DIFF
--- a/etc/ci_scripts/dump_artifacts/cli.py
+++ b/etc/ci_scripts/dump_artifacts/cli.py
@@ -6,7 +6,7 @@ import os
 import sys
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Optional
+from typing import Optional
 
 # Odd path is due to the `/etc` root pattern in pants.toml, fyi
 from ci_scripts.dump_artifacts import docker_artifacts, nomad_artifacts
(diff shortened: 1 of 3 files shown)


etc/ci_scripts/dump_artifacts/cli.py
```diff
@@ -21,33 +23,34 @@
...
+            "--dump-agent-logs",
+            dest="dump_agent_logs",
+            required=False,
+            default=False,
+            help="Dump the logs for Nomad/Consul agents (e.g. running locally). [True / False]",
+        )
+        args = parser.parse_args()
+        self.compose_project: Optional[str] = args.compose_project
+        # coerce from str -> Optional[bool] -> bool
+        self.dump_agent_logs = bool(to_bool(args.dump_agent_logs))
```
wimax-grapl, line 48: personally, I really don't like argparse's bool handling, which is `--bool-flag` vs `--no-bool-flag`, preferring --dump-agent-logs=[Tt]rue. But the `--no-bool-flag` _is_ a posixism i think

christophermaier: Counter argument: `--flag` vs `--no-flag` is inherently boolean, and you never have to wonder if the value is governed by `0` vs. `1`, `true` vs. `false`, `True` vs. `False`, `yes` vs. `no`, `Yes` vs. `No`, presence or absence, or any (potentially inconsistent!) subset of those options.

wimax-grapl: fair. again, explicitly called this out knowing others would disagree, and my position isn't particularly defensible beyond personal preference, so I'll change it :)

PR #1310, opened 2021-12-12
Make up spins up local grapl via Nomad. 

ghost:
This sets make up to spin up the local grapl environment via Nomad instead of the old docker-compose
We're keeping up-detach for the moment as an alias

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -391,9 +391,6 @@ up: build ## Bring up local Grapl and detach to return control to tty
 		pulumi
 	echo "Pulumi complete"
 
-.PHONY: up-detach
-up-detach: up ## Bring up local Grapl and detach to return control to tty
-
 .PHONY: down
 down: ## docker-compose down - both stops and removes the containers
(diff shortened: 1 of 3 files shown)


colin-grapl (approved):

Makefile
```diff
@@ -397,6 +391,9 @@ up-detach: build ## Bring up local Grapl and detach to return control to tty
 		pulumi
 	echo "Pulumi complete"
 
+.PHONY: up-detach
+up-detach: up ## Bring up local Grapl and detach to return control to tty
```
christophermaier, line 395: I'd just go ahead and delete this. There's a small enough number of engineers working on this that telling them "don't use `up-detach` anymore" is reasonable. Also, there's nothing here to indicate that this is a short-term alias; I think there would be a good chance that this ends up staying around for longer, potentially causing confusion.

christophermaier (approved):
Approved, but with a suggestion.

PR #1312, opened 2021-12-14
Hook Plugin Registry up to Nomad, and plumb through a very basic integration smoke-test.

wimax-grapl:
This is precursor work for https://app.zenhub.com/workspace/o/grapl-security/issue-tracker/issues/719
Just a very basic skeleton of a test framework, so that people can add real, valuable tests later.

- Add a `nomad job validate` that would have unblocked an issue Colin had.
- Build plugin-registry with make build
- add plugin-registry as a service to nomad
- add a smoketest integration test for plugin-registry
- add a fake implementation of GetGeneratorsForEventSourceRequest so that we can run this integration test against something. 

make test-integration

### DIFF
--- a/src/rust/plugin-registry/src/client.rs
+++ b/src/rust/plugin-registry/src/client.rs
@@ -129,8 +129,8 @@ where
         self.inner
             .get_generators_for_event_source(GetGeneratorsForEventSourceRequestProto::from(request))
             .await
-            .expect("todo");
-        todo!()
+            .map(|resp| resp.into_inner().try_into().expect("proto to rs"))
+            .map_err(|_status| PluginRegistryServiceError::TodoImplementThisEnumError())
     }
     /// Given information about a tenant, return all analyzers for that tenant
     pub async fn get_analyzers_for_tenant(
--- a/src/rust/plugin-registry/src/client.rs
+++ b/src/rust/plugin-registry/src/client.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use rust_proto::plugin_registry::{
     plugin_registry_service_client::PluginRegistryServiceClient as _PluginRegistryServiceClient,
     CreatePluginRequest,
@@ -19,9 +21,13 @@ use rust_proto::plugin_registry::{
     TearDownPluginRequestProto,
     TearDownPluginResponse,
 };
-use tonic::codegen::{
-    Body,
-    StdError,
+pub use tonic::transport::Channel;
+use tonic::{
+    codegen::{
+        Body,
+        StdError,
+    },
+    transport::Endpoint,
 };
 
 use crate::server::PluginRegistryServiceError;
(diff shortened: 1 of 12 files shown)


src/rust/plugin-registry/src/client.rs
```diff
@@ -33,6 +39,33 @@ pub struct PluginRegistryServiceClient<T> {
     inner: _PluginRegistryServiceClient<T>,
 }
 
+const HOST_ENV_VAR: &'static str = "GRAPL_PLUGIN_REGISTRY_HOST";
+const PORT_ENV_VAR: &'static str = "GRAPL_PLUGIN_REGISTRY_PORT";
+
+impl PluginRegistryServiceClient<Channel> {
```
wimax-grapl, line 45: this is basically the same code as model-plugin-deployer/src/client.rs - our other extant rust grpc service

src/rust/plugin-registry/src/client.rs
```diff
@@ -96,8 +129,8 @@ where
         self.inner
             .get_generators_for_event_source(GetGeneratorsForEventSourceRequestProto::from(request))
             .await
-            .expect("todo");
-        todo!()
+            .map(|resp| resp.into_inner().try_into().expect("proto to rs"))
+            .map_err(|_status| PluginRegistryServiceError::TodoImplementThisEnumError())
```
wimax-grapl, line 133: i just needed some concrete PluginRegistryServiceError to get this to compile

src/rust/plugin-registry/src/server.rs
```diff
@@ -135,18 +145,17 @@ impl PluginRegistryService for PluginRegistry {
...
 }
 
-pub async fn exec_service() -> Result<(), Box<dyn std::error::Error>> {
+pub async fn exec_service(socket_addr: SocketAddr) -> Result<(), Box<dyn std::error::Error>> {
     let (mut health_reporter, health_service) = tonic_health::server::health_reporter();
     health_reporter
         .set_serving::<PluginRegistryServiceServer<PluginRegistry>>()
         .await;
 
-    let addr = "[::1]:50051".parse().unwrap();
```
wimax-grapl, line 144: no more hardcoding! pass in the port

docker-compose.build.yml
```diff
@@ -83,6 +83,15 @@ services:
       args:
         - RUST_BUILD=${RUST_BUILD:-debug}
 
+  plugin-registry:
```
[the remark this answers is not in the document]
wimax-grapl: GOOD CATCH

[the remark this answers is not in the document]
wimax-grapl: i might even automate this

src/rust/plugin-registry/src/server.rs
```diff
@@ -124,7 +129,12 @@ impl PluginRegistryService for PluginRegistry {
         &self,
         _request: Request<GetGeneratorsForEventSourceRequestProto>,
     ) -> Result<Response<GetGeneratorsForEventSourceResponseProto>, Status> {
-        todo!()
+        // Stub implementation, for integration test smoke-test purposes.
+        // Replace with a real implementation soon!
+        let message = _request.into_inner();
+        Ok(Response::new(GetGeneratorsForEventSourceResponseProto {
+            plugin_ids: [message.event_source_id].into_iter().flatten().collect(),
+        }))
```
colin-grapl, line 137: It might make sense to just leave this as `todo` and mark the test as `#[should_panic]` or something

colin-grapl (approved):

src/rust/plugin-registry/src/server.rs
wimax-grapl: _yes_ but I want to test that we get a real response. should_panic could result from any number of things, like the client not working. should_panic doesn't guarantee we have a real connection

wimax-grapl: i think a response that's explicitly "yo this is a fake" would be the better solution

PR #1325, opened 2021-12-22
Implement plugin work queue. Add postgres with extensions

colin-grapl:
728

1. Adds a new docker image `postgres-ext`, which is based on postgres 13 but adds the pg_partman and pg_cron extensions.

2. Implements the plugin-work-queue service.

Integration tests are provided.


src/proto/graplinc/grapl/api/plugin_work_queue/v1beta1/plugin_work_queue.proto
```diff
@@ -39,33 +43,61 @@ message GetExecuteGeneratorRequest {}
 // A request to retrieve a new ExecutionJob for an Analyzer
 message GetExecuteAnalyzerRequest {}
 
+message NoAvailableJobs {}
+
 // A response containing the next ExecutionJob to execute against a Generator
 message GetExecuteGeneratorResponse {
   // The execution_job to be executed
-  ExecutionJob execution_job = 1;
+  oneof maybe_job {
```
wimax-grapl, line 51: I'm not 100% opposed to it, but, ultimately, this feels like a reinvention of an optional field (yes, I know everything is optional in proto3 and they're a pain to work with). Is this the preferred pattern?

src/proto/graplinc/grapl/api/plugin_work_queue/v1beta1/plugin_work_queue.proto
```diff
@@ -39,33 +43,61 @@ message GetExecuteGeneratorRequest {}
...
+    // When a job is available
+    ExecutionJob job = 1;
+    // When a job is not available
+    NoAvailableJobs no_jobs = 2;
+  };
   // A request_id, to be used in Acknowledgement
-  graplinc.common.v1beta1.Uuid request_id = 2;
+  int64 request_id = 3;
```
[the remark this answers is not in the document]
wimax-grapl: also - you know me, I love to NewType every id type I can see. Would that be worth doing in proto land

src/rust/grapl-web-ui/src/main.rs
```diff
@@ -31,7 +31,7 @@ async fn main() -> Result<(), GraplUiError> {
     HttpServer::new(move || {
         App::new()
             .wrap(actix_web::middleware::Logger::default())
-            .wrap(actix_web::middleware::Compress::default())
+            // .wrap(actix_web::middleware::Compress::default())  // todo: Reenable compression when brotli isn't vulnerable
```
wimax-grapl, line 34: For long term todos like this pls add a github issue link:
https://github.com/grapl-security/issue-tracker/issues/803

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
```diff
@@ -0,0 +1,86 @@
+CREATE TYPE status_t AS ENUM ('enqueued', 'failed', 'processed');
+
+CREATE SCHEMA plugin_work_queue;
+
+CREATE FUNCTION megabytes(bytes integer) RETURNS integer AS
+$$
+BEGIN
+    RETURN bytes * 1000 * 1000;
+END;
+$$ LANGUAGE plpgsql;
```
wimax-grapl, line 10: 1. can you document what the 256MB restriction is for?
2. one could argue its 1024 just saying

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
```diff
@@ -0,0 +1,86 @@
...
+CREATE EXTENSION pg_cron;
+
+UPDATE partman.part_config
+SET infinite_time_partitions = true,
+    retention                = '1 month',
+    retention_keep_table= true
+WHERE (
+    parent_table = 'plugin_work_queue.generator_plugin_executions' OR
+    parent_table = 'plugin_work_queue.analyzer_plugin_executions'
+);
```
wimax-grapl, line 84: been a hot second since i sqld but
where parent_table in ('first', 'second') ?

src/rust/plugin-work-queue/sqlx-data.json
```diff
@@ -0,0 +1,202 @@
+{
```
wimax-grapl, line 1: Since this is a generated file, you may want to add it to `.gitattributes`

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,332 @@
+use sqlx::{
```
wimax-grapl, line 1: ## Note to max this is how far I got reviewing

postgres/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM postgres:13.4-bullseye
+
+RUN apt update && \
+    apt -y install \
+    postgresql-13-partman \
+    postgresql-13-cron
+
+
+RUN echo "shared_preload_libraries = 'pg_cron'" >> /usr/share/postgresql/13/postgresql.conf.sample
+RUN echo "cron.database_name = 'postgres'" >> /usr/share/postgresql/13/postgresql.conf.sample
```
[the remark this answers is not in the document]
colin-grapl: Same but it seems to be the case.

src/proto/graplinc/grapl/api/plugin_work_queue/v1beta1/plugin_work_queue.proto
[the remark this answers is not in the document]
colin-grapl: I believe we already disable that lint for anything v1beta1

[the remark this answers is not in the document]
colin-grapl: The request id is based on the `serial` value in the db.

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
colin-grapl: It seems like a reasonable restriction on a given binary. We can always raise it without issue. I just don't want someone shoving 50GB into the database.

src/proto/graplinc/grapl/api/plugin_work_queue/v1beta1/plugin_work_queue.proto
```diff
@@ -19,6 +19,8 @@ message ExecutionJob {
 message PutExecuteGeneratorRequest {
   // The job to be executed
   ExecutionJob execution_job = 1;
+  // The trace id for this request
+  graplinc.common.v1beta1.Uuid trace_id = 2;
```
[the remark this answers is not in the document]
colin-grapl: I'm going to remove the trace_id for now. We can add it back in when we have a plan for tracing. For example, grpc might have some sort of native tracing that I'm unaware of.

src/proto/graplinc/grapl/api/plugin_work_queue/v1beta1/plugin_work_queue.proto
colin-grapl: It is indeed. Google provides the 'Empty' structure for this reason, but in this case I went with a custom one as it's more extendable (for example if we wanted to add a "try again after x seconds", etc).

protobuf sux

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
colin-grapl: Is that preferable? I think that would work fwiw but I have no sense of whether one is better or worse.

src/rust/plugin-work-queue/sqlx-data.json
colin-grapl: Neat, will do.

src/proto/graplinc/grapl/api/plugin_work_queue/v1beta1/plugin_work_queue.proto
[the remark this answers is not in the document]
colin-grapl: The problem with newtypes in protobuf is that protobuf is shit. It isn't free to do that. You end up with more complex generated code that has to worry about both the outer message and the inner value being optional, so it's really gross.

I hate protobuf if that isn't clear.

src/rust/plugin-work-queue/src/server.rs
```diff
@@ -24,111 +31,266 @@ use rust_proto::plugin_work_queue::{
...
         _request: GetExecuteGeneratorRequest,
     ) -> Result<GetExecuteGeneratorResponse, PluginWorkQueueError> {
-        todo!()
+        let message = self.queue.get_analyzer_message().await?;
+        let message = match message {
+            Some(message) => message,
+            None => {
+                return Ok(GetExecuteGeneratorResponse {
+                    execution_job: None,
+                    request_id: 0,
```
wimax-grapl, line 134: Is this not a NoAvailableJobs scenario? 

Even if it's not: it's not immediately obvious to me _what_ scenario this is.

src/rust/plugin-work-queue/src/server.rs
```diff
@@ -24,111 +31,266 @@ use rust_proto::plugin_work_queue::{
...
+        Ok(PutExecuteAnalyzerResponse {})
     }
 
     #[allow(dead_code)]
     async fn get_execute_generator(
         &self,
         _request: GetExecuteGeneratorRequest,
     ) -> Result<GetExecuteGeneratorResponse, PluginWorkQueueError> {
-        todo!()
+        let message = self.queue.get_analyzer_message().await?;
```
wimax-grapl, line 128: Another which-one-is-it mismatch, this should be get_generator_message.

wimax-grapl: The fact we can get this far along and still be wrong about message type worries me. 
Throwing this out there without having investigate it at all: worth it to perhaps genericize Message with a phantom type or something?

src/rust/plugin-work-queue/src/server.rs
```diff
@@ -24,111 +31,266 @@ use rust_proto::plugin_work_queue::{
...
+        })
+    }
+
+    #[allow(dead_code)]
+    async fn acknowledge_generator(
+        &self,
+        request: AcknowledgeGeneratorRequest,
+    ) -> Result<AcknowledgeGeneratorResponse, PluginWorkQueueError> {
+        self.queue
+            .ack_generator_failure(request.request_id.into())
```
wimax-grapl, line 181: Not immediately obvious to me why this should be an `ack_*_failure`

src/rust/plugin-work-queue/src/server.rs
```diff
@@ -24,111 +31,266 @@ use rust_proto::plugin_work_queue::{
...
+        service_config.plugin_work_queue_db_port,
+    );
+
+    let plugin_work_queue: PluginWorkQueue = PluginWorkQueue::from(
+        sqlx::PgPool::connect(&postgres_address)
+            .timeout(std::time::Duration::from_secs(5))
+            .await??,
+    );
+
+    sqlx::migrate!().run(&plugin_work_queue.queue.pool).await?;
```
wimax-grapl, line 293: So wait, every time this service starts it runs a migration? If so that seems wrong.

src/rust/plugin-work-queue/tests/psql_queue_tests.rs
```diff
@@ -0,0 +1,102 @@
...
+        last_updated: DateTime<Utc>,
+        visible_after: Option<DateTime<Utc>>,
+        try_count: i32,
+    }
+
+    #[tokio::test]
+    async fn test_migrate() -> Result<(), Box<dyn std::error::Error>> {
+        let pool = PgPoolOptions::new()
+            .max_connections(1)
+            .connect("postgresql://postgres:password@localhost:5432/postgres")
```
wimax-grapl, line 31: The fact this communicates to localhost HEAVILY implies to me that this test is not hooked up correctly.

Sure enough, it looks like `RUST_INTEGRATION_TEST_FEATURES=` was never updated to include these tests.

wimax-grapl: I suspect you're running this locally on your box with some form of `cargo test`. That style of development is convenient but should be considered deprecated or, worse, misleading.

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,325 @@
...
+             WHERE plugin_work_queue.analyzer_plugin_executions.execution_key = next_execution.execution_key
+             RETURNING next_execution.execution_key as "execution_key!: ExecutionId", next_execution.plugin_id, next_execution.pipeline_message, next_execution.tenant_id
+        "#).fetch_optional(&self.pool)
+            .await?;
+
+        Ok(request.map(|request| Message { request }))
+    }
+
+    #[instrument(skip(output), err)]
+    pub async fn ack_generator_success(
```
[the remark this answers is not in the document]
colin-grapl: The query for success/failure is more different than just setting success/failure. On success it also has to set the output.

src/rust/plugin-work-queue/src/server.rs
colin-grapl: It might be possible. I can experiment. But I think the ideal solution would involve having the const string be part of the type. I can probably get that working actually, I'll see what I can do.

src/rust/plugin-work-queue/src/server.rs
colin-grapl: This is a NoAvailableJobs scenario. it's not really exposed on the rust end other than as an Optional. I could bring it over to the rust side.

src/rust/plugin-work-queue/src/server.rs
colin-grapl: It shouldn't be. I realize there's an issue with the Ack concept currently.

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
wimax-grapl: i'd imagine perf wise they're the same

src/rust/plugin-work-queue/src/server.rs
colin-grapl: It will first ask the db which migrations have already applied and not run any that have already been applied. This is apparently standard practice.

src/rust/plugin-work-queue/tests/psql_queue_tests.rs
colin-grapl: Yeah, the test_migrate needs to die entirely actually. As for adding the tests, will do. And I'll have this use a from_env kinda thing.

wimax-grapl (approved):

wimax-grapl (changes requested):
jk
- please make nomad changes to make this an available service
- please make nomad changes to integration-tests.nomad to supply PLUGIN_WORK_QUEUE_stuff=

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -302,6 +261,7 @@ impl PsqlQueue {
                 WHERE execution_key = $1
             "#,
             execution_key.0,
+            status as _,
```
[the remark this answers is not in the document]
colin-grapl: Sort of. `.into()` will actually be a method call, but `as` is more like a cast.

src/rust/plugin-work-queue/tests/psql_queue_tests.rs
```diff
@@ -3,98 +3,112 @@
...
 
         // Retrieve a message
-        let msg = queue.get_message().await?.expect("No valid message");
+        let msg = queue
+            .get_generator_message()
+            .await?
+            .expect("No valid message");
         let execution_key = msg.request.execution_key;
         tracing::info!(message="Received message", execution_key=?msg.request.execution_key);
         // Acknowledge the message
```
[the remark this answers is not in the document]
colin-grapl: Done on line 68.

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+        // 1. It attempts to get a message from the queue
+        //      -> Where that message isn't over a day old
+        //      -> Where that message is "visible"
+        //      -> Where that message isn't currently being evaluated by another transaction
+        //      -> Where that message is in the 'enqueued' state
+        // 2. Updates the `try_count`
+        // 3. Updates the `visible_after`
+
+        // Note that:
+        // * messages are invisible for 10 seconds
```
christophermaier, line 137: Can you say more about the intention of "invisibility" here? Is it effectively saying that this process has a "lock" on this record for the next little bit to ensure it's not being used elsewhere?

If so, we could probably simplify this a bit with row locks, as described in https://spin.atomicobject.com/2021/02/04/redis-postgresql.

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+            SET
+                try_count  = plugin_work_queue.generator_plugin_executions.try_count + 1,
+                last_updated = CURRENT_TIMESTAMP,
+                visible_after  = CURRENT_TIMESTAMP + INTERVAL '10 seconds'
+            FROM (
+                 SELECT execution_key, plugin_id, pipeline_message, status, creation_time, visible_after, tenant_id
+                 FROM plugin_work_queue.generator_plugin_executions
+                 WHERE status = 'enqueued'
+                   AND creation_time >= (CURRENT_TIMESTAMP - INTERVAL '1 day')
+                   AND (visible_after IS NULL OR visible_after <= CURRENT_TIMESTAMP)
```
christophermaier, line 156: If `visible_after` can't be replaced with row locks, we should probably make it `NOT NULL` and populate it with an appropriate value at row insert time.

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+                 ORDER BY creation_time ASC
+                 FOR UPDATE SKIP LOCKED
```
christophermaier, line 158: Wait, you're already using row locks... was my interpretation of visibility wrong?

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+                 LIMIT 1
+             ) AS next_execution
+             WHERE plugin_work_queue.generator_plugin_executions.execution_key = next_execution.execution_key
+             RETURNING
+                 next_execution.execution_key as "execution_key!: ExecutionId",
```
christophermaier, line 163: What's the significance of `"execution_key!: ExecutionId"`; that's a rather weird column name.

(Bonus style nitpick: we should capitalize that `as`.)

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+    ) -> Result<(), PsqlQueueError> {
+        sqlx::query!(
+            r#"
+                UPDATE plugin_work_queue.generator_plugin_executions
+                SET status = $2,
+                    last_updated = CASE
+                        WHEN status != 'processed'
+                            THEN CURRENT_TIMESTAMP
+                            ELSE last_updated
+                        END
```
christophermaier, line 235: Would it be possible to invert the sense of this comparison to make it easier to think about? (negations are always a little trickier to parse):
```sql
WHEN status = 'processed'
THEN last_updated
ELSE CURRENT_TIMESTAMP
```
Also, how does the data flow work such that the existing status would _not_ be 'processed'? Leaving the `last_updated` field unchanged seems to imply that everything has already been handled.

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// Pub for testing - otherwise sqlx can't see the query
+pub async fn get_generator_status(
+    pool: &sqlx::Pool<Postgres>,
+    execution_key: &ExecutionId,
+) -> Result<Status, sqlx::Error> {
+    // The request should be marked as failed
+    let row = sqlx::query!(
+        r#"SELECT status as "status: Status"
```
christophermaier, line 279: What's the deal with these column names?

src/rust/plugin-work-queue/src/server.rs
```diff
@@ -1,12 +1,18 @@
+use grapl_utils::future_ext::GraplFutureExt;
```
christophermaier, line 1: NOTE: Chris stopped reviewing here; more tomorrow.

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
```diff
@@ -0,0 +1,80 @@
+CREATE TYPE status_t AS ENUM ('enqueued', 'failed', 'processed');
+
+CREATE SCHEMA plugin_work_queue;
```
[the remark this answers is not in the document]
colin-grapl: I could put them in the schema, just didn't think of it. We have to put our tables in schemas for pg_partman.

src/rust/plugin-work-queue/src/psql_queue.rs
colin-grapl: It means it's not null and that it's an ExecutionId so that sqlx can parse it.

src/rust/plugin-work-queue/src/psql_queue.rs
colin-grapl: https://docs.rs/sqlx/latest/sqlx/macro.query_as.html#column-type-override-infer-from-struct-field

src/rust/plugin-work-queue/src/psql_queue.rs
colin-grapl: I'm not sure what you're asking. SKIP LOCKED is basically saying "get the next row that can't be immediately locked", so if there's a query currently performing an update a separate query will skip it entirely without waiting.

Here's a post on it:
https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/

src/rust/plugin-work-queue/src/psql_queue.rs
colin-grapl: I can invert it.

As for the data flow, the status starts off as 'enqueued'. The 'ack' will change it to either 'processed' or 'failed'.

I check for '!= processed' first because we don't want to update the status for a message that raced ie: if another consumer already processed this message we want *that* 'last_updated' time.

So this basically is a way of expressing "Either update it or leave it alone", which is why I had the logic inverted - the interesting cast is when status != 'processed'.

src/rust/rust-proto/src/plugin_work_queue.rs
```diff
@@ -65,44 +84,89 @@ impl From<ExecutionJob> for ExecutionJobProto {
...
         }
     }
 }
 
-pub struct AcknowledgeResponse {}
+pub struct AcknowledgeAnalyzerResponse {}
 
-impl TryFrom<AcknowledgeResponseProto> for AcknowledgeResponse {
+impl TryFrom<AcknowledgeAnalyzerResponseProto> for AcknowledgeAnalyzerResponse {
     type Error = PluginWorkQueueDeserializationError;
```
christophermaier, line 161: Infallible, etc.

src/rust/rust-proto/src/plugin_work_queue.rs
```diff
@@ -177,44 +249,52 @@ impl From<GetExecuteGeneratorRequest> for GetExecuteGeneratorRequestProto {
...
-            .request_id
-            .ok_or(Self::Error::MissingRequiredField(
-                "GetExecuteGeneratorResponse.request_id",
-            ))?
-            .into();
-        let execution_job = value
-            .execution_job
+        let request_id = value.request_id;
+        let maybe_job = value.maybe_job.ok_or(Self::Error::MissingRequiredField(
+            "GetExecuteGeneratorResponseProto.maybe_job",
```
christophermaier, line 262: Can tracing provide the context to know what proto we're dealing with, allowing us to just say `Error::MissingRequiredField("maybe_job")`?

src/rust/rust-proto/src/plugin_work_queue.rs
```diff
@@ -177,44 +249,52 @@ impl From<GetExecuteGeneratorRequest> for GetExecuteGeneratorRequestProto {
...
+        ))?;
+        let execution_job = match maybe_job {
+            get_execute_generator_response::MaybeJob::Job(job) => Some(job),
+            get_execute_generator_response::MaybeJob::NoJobs(_) => None,
+        };
```
christophermaier, line 267: Can this `MaybeJob` -> `Option` (and back) conversion be implemented explicitly with `From`/`Into`?

I generally prefer this kind of behavior to be expressed in terms of traits, rather than buried away in individual method bodies, for clarity, explicitness, reuse, documentation, testability, etc.

src/rust/plugin-work-queue/src/psql_queue.rs
christophermaier: Wild... TIL.

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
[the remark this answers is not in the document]
colin-grapl: Makes sense, will add those to the schema.

src/rust/plugin-work-queue/src/psql_queue.rs
colin-grapl: If one worker acquires a job and hangs for the full visibility timeout another worker can acquire that same job. At that point both workers could attempt to ack the job.

src/rust/plugin-work-queue/src/server.rs
```diff
@@ -24,112 +31,292 @@ use rust_proto::plugin_work_queue::{
...
+    psql_queue::{
+        PsqlQueue,
+        PsqlQueueError,
+    },
+    PluginWorkQueueServiceConfig,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum PluginWorkQueueError {
+    #[error("PsqlQueueError")]
```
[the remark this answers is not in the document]
colin-grapl: Is there a pattern we should be following?

src/rust/plugin-work-queue/src/psql_queue.rs
christophermaier: 👍 makes sense

src/rust/rust-proto/src/plugin_work_queue.rs
```diff
@@ -65,44 +84,89 @@ impl From<ExecutionJob> for ExecutionJobProto {
...
+
+pub struct AcknowledgeGeneratorResponse {}
+
+impl TryFrom<AcknowledgeGeneratorResponseProto> for AcknowledgeGeneratorResponse {
+    type Error = PluginWorkQueueDeserializationError;
+
+    fn try_from(_value: AcknowledgeGeneratorResponseProto) -> Result<Self, Self::Error> {
+        Ok(Self {})
+    }
+}
```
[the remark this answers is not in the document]
colin-grapl: I didn't realize that Infallible was stable. Regardless though, if we ever add fields to the proto this becomes a fallible operation.

src/rust/rust-proto/src/plugin_work_queue.rs
```diff
@@ -65,44 +84,89 @@ impl From<ExecutionJob> for ExecutionJobProto {
...
+
+pub struct AcknowledgeAnalyzerRequest {
+    pub request_id: i64,
+    pub success: bool,
+}
+
+impl TryFrom<AcknowledgeAnalyzerRequestProto> for AcknowledgeAnalyzerRequest {
+    type Error = PluginWorkQueueDeserializationError;
```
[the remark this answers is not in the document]
colin-grapl: See above - the reason to use TryFrom here is because adding fields to protobuf will make this fallible. Currently there are no fields, but it seems best to express "this could fail".

src/rust/rust-proto/src/plugin_work_queue.rs
colin-grapl: Possibly, I'm not sure if orphan rules allow us to implement traits we don't own on types like Option<T>, even if we own T. At one point it was forbidden, I don't know if they've changed that - though I know there has been discussion for years.

src/rust/rust-proto/src/plugin_work_queue.rs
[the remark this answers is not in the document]
christophermaier: We should make a note to that effect here, then.

Otherwise, someone like me is going to come in a year from now, set the type to `Infallible`, and shenanigans will ensue 😅

src/rust/plugin-work-queue/README.md
```diff
@@ -0,0 +1,106 @@
...
+If the message is successfully processed or if it fails, update the row:
+1. Set the `execution_result`
+2. Set the `last_updated` to CURRENT_TIMESTAMP
+3. Set the `status` to `processed` or `failed` accordingly
+
+Otherwise, if the message is not successfully processed but can be retried,
+do nothing. It will be picked up again later.
+
+Note that just because a message has the `enqueued` status does *not* mean that it will be processed. We don't
+guarantee that, on failure, we'll mark the `status` as failed.
```
christophermaier, line 99: Why? This seems important to spell out explicitly.

Also, this sounds like two separate things. "Not being processed" sounds like we can have a message that is submitted to the queue, but is never once picked up by a worker.

Never marking something as failed seems like the second thing. We may want to think about another failure state for this. The current `failed` status arises when we receive an acknowledgement from the worker that, indeed, it tried to process this message, but it did not work. The "new" failure status seems like it would involve exceeding a retry count. That is, we keep sending this message out to workers, but we never get a response back. As a result, we continue to increment the try count. After some maximum is exceeded, we probably need to mark this message as having this new kind of failure state, at which point we can decide to handle it some other way (and stop occupying resources by repeatedly processing it).

(Incidentally, it seems like "aged out" could be another kind of failure status that should be formally noted.)

src/rust/plugin-work-queue/README.md
```diff
@@ -0,0 +1,106 @@
...
+`last_updated` is set with each update to the row
+
+`visible_after` is essentially a visibility timeout. A `visible_after` of NULL means a message is available, which
+is the default state. Every time a message is SELECT'd we update `visible_after`. See the `Visibility Timeout` section
+below.
+
+`try_count` on every receive we increment `try_count` to indicate another attempt to process this message
+
+
+## Algorithm
```
christophermaier, line 75: After reading through this algorithm, I think a few changes would help to make it significantly clearer.

I think the various concepts the algorithm mentions need to be defined first, and should be defined in generic terms. For instance, defining "visible" as "has a timestamp that is less than the current timestamp" is true in the strictest mechanical sense, but doesn't say anything about _why_ this is something that is important.

Some important concepts I can see that need description:
- age and "aging out". Why do we care about age, why is 1 day the limit that is chosen, what are the implications for messages that are too old?
- the set of statuses a message may have, and what governs the transitions between statuses. In particular, it seems that we have an implicit status of "dispatched" that is not part of the SQL status enum, but is currently implemented as "being invisible". Not being part of the enum is fine if that's not how we need to implement it, but it should be present as a concept (and our algorithm should be expressed in terms of it). (The implementation could also be reworked so this is more readily apparent in the code, even if it's not part of the enum; it could be incorporated into a view, for instance. Actually, I think views could help to clarify _a lot_ of this logic. For instance, a view could be based on both the status enum and the visibility timeout, providing a unified representation of the _conceptual_ state. Our status enum would thus be a _partial_ representation of the complete state space, and would simply be part of the implementation details.)
- the notion of "visibility". As written, the idea is spread throughout three separate parts of this file (the schema, the algorithm, and the dedicated "Visibility Timeout" section). It would be much better to consolidate all this information into a single description, along with the rest of the core concepts. The value of the interval being added to the visibility timestamp is also not described, nor the rationale for choosing its static value.
- retries. Under what conditions will a message be retried? Can a message end up being dispatched to multiple workers at the same time? What are the implications of that? Is there a maximum amount of times a message may be retried? What happens if that amount is exceeded?

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
```diff
@@ -0,0 +1,80 @@
...
+
+CREATE TYPE plugin_work_queue.status_t AS ENUM ('enqueued', 'failed', 'processed');
+
+CREATE FUNCTION plugin_work_queue.megabytes(bytes integer) RETURNS integer AS
```
christophermaier, line 5: We should add an `IF NOT EXISTS` clause to this (and to _every_ object that honors that clause).

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
```diff
@@ -0,0 +1,80 @@
...
+    PARTITION BY RANGE (creation_time);
+
+CREATE INDEX IF NOT EXISTS execution_key_ix ON plugin_work_queue.generator_plugin_executions (execution_key);
+CREATE INDEX IF NOT EXISTS creation_time_ix ON plugin_work_queue.generator_plugin_executions (creation_time);
+
+CREATE INDEX IF NOT EXISTS execution_key_ix ON plugin_work_queue.analyzer_plugin_executions (execution_key);
+CREATE INDEX IF NOT EXISTS creation_time_ix ON plugin_work_queue.analyzer_plugin_executions (creation_time);
+
+CREATE SCHEMA partman;
+CREATE EXTENSION pg_partman WITH SCHEMA partman;
```
christophermaier, line 51: `CREATE SCHEMA` and `CREATE EXTENSION` both honor the `IF NOT EXISTS` clause.

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
```diff
@@ -0,0 +1,80 @@
...
+           );
+
+SELECT partman.create_parent(p_parent_table => 'plugin_work_queue.analyzer_plugin_executions',
+                             p_control => 'creation_time',
+                             p_type => 'native',
+                             p_interval=> 'daily',
+                             p_premake => 30
+           );
+
+CREATE EXTENSION pg_cron;
```
christophermaier, line 69: `IF NOT EXISTS`

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
```diff
@@ -0,0 +1,80 @@
+CREATE SCHEMA plugin_work_queue;
+
+CREATE TYPE plugin_work_queue.status_t AS ENUM ('enqueued', 'failed', 'processed');
```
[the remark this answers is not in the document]
colin-grapl: There is a field in each table called 'status', hence 'status_t'. But I could rename that field to `current_status` to disambiguate.

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
colin-grapl: There is no IF NOT EXISTS for CREATE FUNCTION but I can add OR REPLACE. Given that a migration should never run twice I don't think it should come up but it seems fine to put in defensively.

src/rust/plugin-work-queue/sqlx-data.json
```diff
@@ -0,0 +1,217 @@
...
```
colin-grapl, line 1: It looks like the globbing is not working in .gitattributes?

src/rust/plugin-work-queue/src/client.rs
```diff
@@ -40,62 +53,87 @@ where
     }
 
     /// Adds a new execution job for a generator
+    #[tracing::instrument(skip(self, request), err)]
```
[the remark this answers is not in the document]
colin-grapl: The `tracing::instrument` macro serves two (main) purposes:

1. To add arguments and return values to the trace
2. To instrument asynchronous functions so that traces within them are correct - otherwise you end up with a single trace that has events from multiple executing futures

In general I would say 'skip anything you do not want to add as context', and generally I would say skip everything and then selectively add things back.

Not sure what you mean re: macro.

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+#[derive(sqlx::Type, Debug, Clone, PartialEq, Eq)]
+#[sqlx(type_name = "status_t", rename_all = "lowercase")]
+pub enum Status {
+    Enqueued,
+    Failed,
+    Processed,
+}
+
+#[derive(Copy, Clone, Debug, sqlx::Type)]
+#[sqlx(transparent)]
```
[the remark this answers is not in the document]
colin-grapl: Basically yeah

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+        id.0
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct NextExecutionRequest {
+    pub execution_key: ExecutionId,
+    pub plugin_id: uuid::Uuid,
+    pub tenant_id: uuid::Uuid,
+    pub pipeline_message: Vec<u8>,
```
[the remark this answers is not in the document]
colin-grapl: They can be pub(crate), but yes it's just a data object so that we can get sqlx to deserialize a given row for a specific query in a way that's ergonomic.

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum PsqlQueueError {
+    #[error("Sqlx")]
```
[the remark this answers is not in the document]
colin-grapl: I can raise more information by formatting the inner error. I had assumed that was done by default.

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+        // * messages currently do not have a maximum retry limit
+        // * The one day expiration matches our 1 day partitioning strategy
+
+        // In the future we can leverage a maximum retry limit as well as a batch version of this query
+        // A more dynamic visibility strategy would also be reasonable
+        let request: Option<NextExecutionRequest> = sqlx::query_as!(
+            NextExecutionRequest, r#"
+            UPDATE plugin_work_queue.generator_plugin_executions
+            SET
+                try_count  = plugin_work_queue.generator_plugin_executions.try_count + 1,
```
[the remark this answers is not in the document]
colin-grapl: For try_count yes. I don't think for the others, not sure.

src/rust/plugin-work-queue/README.md
```diff
@@ -0,0 +1,106 @@
...
+The query for processing a message is:
+
+1. Find the oldest message that
+    1. Has not aged out (1 day)
+    2. Is not currently being locked by another request
+    3. Is 'enqueued'
+    4. Is visible ie: the `visible_after` is null or < CURRENT_TIMESTAMP
```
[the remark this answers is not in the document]
colin-grapl: Done

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+
+#[derive(sqlx::Type, Debug, Clone, PartialEq, Eq)]
+#[sqlx(type_name = "status_t", rename_all = "lowercase")]
+pub enum Status {
+    Enqueued,
+    Failed,
+    Processed,
+}
+
+#[derive(Copy, Clone, Debug, sqlx::Type)]
```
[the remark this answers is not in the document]
colin-grapl: I wonder if we can get rustfmt to do that for us in the future. I'll do that for now - there's no significant to the ordering.

src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+        // * The one day expiration matches our 1 day partitioning strategy
+
+        // In the future we can leverage a maximum retry limit as well as a batch version of this query
+        // A more dynamic visibility strategy would also be reasonable
+        let request: Option<NextExecutionRequest> = sqlx::query_as!(
+            NextExecutionRequest, r#"
+            UPDATE plugin_work_queue.generator_plugin_executions
+            SET
+                try_count  = plugin_work_queue.generator_plugin_executions.try_count + 1,
+                last_updated = CURRENT_TIMESTAMP,
```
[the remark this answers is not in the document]
colin-grapl: It hadn't occurred to me that I can specify CURRENT_TIMESTAMP in the insert - I'll definitely do that.

src/rust/plugin-work-queue/src/psql_queue.rs
colin-grapl: I think we resolved this via slack, right? To reiterate for record, the only locks being used are the ones postgres implicitly acquires when a row is accessed and SKIP LOCKED is a feature where postgres will skip any rows that are currently locked. The visible_after serves as a deadline for the row to be evaluated, and other jobs can pick it up after, without the need to hold any transaction open.

I've added more information on this in the readme.

src/rust/plugin-work-queue/src/server.rs
```diff
@@ -24,112 +31,292 @@ use rust_proto::plugin_work_queue::{
...
+
+    #[allow(dead_code)]
+    async fn acknowledge_analyzer(
+        &self,
+        request: AcknowledgeAnalyzerRequest,
+    ) -> Result<AcknowledgeAnalyzerResponse, PluginWorkQueueError> {
+        let status = match request.success {
+            true => crate::psql_queue::Status::Processed,
+            false => crate::psql_queue::Status::Failed,
+        };
```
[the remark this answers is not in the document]
colin-grapl: I could do that, but it would mean that `Status` is a type that `rust-proto` would be aware of, as opposed to being specific to the representation in the queue. That's not so trivial - it also means sqlx becomes a dependency of rust-proto, which maybe is fine, but I think merits some discussion about where these sorts of types live.

Instead I can add a From<bool> to Status, which keeps things separate.

I can add  the use.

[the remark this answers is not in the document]
colin-grapl: On second thought I don't think that a From<bool> would be proper. It's too general. I think it's best to leave the two conversions in place for now and address this in later PRs.

src/rust/plugin-work-queue/tests/psql_queue_tests.rs
```diff
@@ -0,0 +1,125 @@
...
+        let pool = make_pool(config).await?;
+        sqlx::migrate!().run(&pool).await?;
+        let queue = PsqlQueue { pool };
+
+        // Ensure one message is queued
+        let plugin_id = uuid::Uuid::new_v4();
+        queue
+            .put_generator_message(
+                plugin_id.clone(),
+                b"some-message".to_vec(),
```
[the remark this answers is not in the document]
colin-grapl: Do you mean taking an argument like `impl Into<Vec<u8>>` ? I don't see a problem with doing that, though by requiring a Vec<u8> we push allocations up a layer. For example,

```rust
fn takes_vec(f: impl Into<Vec<u8>>) {}

fn main() {
    let my_data = vec![];
    takes_vec(&my_data); // forces an unnecessary, hidden allocation
    takes_vec(my_data); // no additional allocation
}
```
Historically idiomatic rust tends to take the 'push up' approach for owned types, so interfaces that require ownership tend to be explicit, but `impl AsRef<T>` tends to be more common.

src/rust/plugin-work-queue/tests/psql_queue_tests.rs
```diff
@@ -0,0 +1,125 @@
...
+        let status = get_generator_status(&queue.pool, &execution_key).await?;
+        assert_eq!(status, Status::Processed);
+
+        Ok(())
+    }
+
+    #[test_log::test(tokio::test)]
+    async fn test_get_and_failure() -> Result<(), Box<dyn std::error::Error>> {
+        let config = PluginWorkQueueServiceConfig::from_args();
+        tracing::info!(message = "test_get_and_failure");
```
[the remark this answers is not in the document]
colin-grapl: I've just found that having a 'sanity' event at the start of a test is often nice for when things go wrong, like something deadlocks early on or panics. It's not to label the test-run, which tracing handles, it's just a default unique message to say "This test is actually executing" whereas the test runner's "log" is much much higher up.

src/rust/rust-proto/src/plugin_work_queue.rs
```diff
@@ -65,44 +84,89 @@ impl From<ExecutionJob> for ExecutionJobProto {
...
-        Ok(Self { request_id })
+    fn try_from(value: AcknowledgeGeneratorRequestProto) -> Result<Self, Self::Error> {
+        let request_id = value.request_id;
+        let success = value.success;
+        Ok(Self {
+            request_id,
+            success,
+        })
+    }
+}
```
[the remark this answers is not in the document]
colin-grapl: In general converting from protobuf to a constrained rust type should be considered fallible, even if the implementation today is not.

src/rust/rust-proto/src/plugin_work_queue.rs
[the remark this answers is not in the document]
colin-grapl: Sounds good. I wonder if this should be something we document in rust-proto at a higher level like the readme? I think it's a good contract and pattern to always have conversion from protobuf messages to rust structures be fallible.

src/rust/rust-proto/src/plugin_work_queue.rs
colin-grapl: Looks like orphan rules now allow this, neat. Will implement.

src/rust/plugin-work-queue/src/psql_queue.rs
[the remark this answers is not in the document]
colin-grapl: Actually they have to be pub as the fields are accessed in tests.

src/rust/plugin-work-queue/src/server.rs
```diff
@@ -24,112 +31,292 @@ use rust_proto::plugin_work_queue::{
...
+
+    async fn acknowledge_analyzer(
+        &self,
+        request: Request<AcknowledgeAnalyzerRequestProto>,
+    ) -> Result<Response<AcknowledgeAnalyzerResponseProto>, Status> {
+        let request = request.into_inner();
+        let request: AcknowledgeAnalyzerRequest =
+            request.try_into().map_err(PluginWorkQueueError::from)?;
+        let response = self.acknowledge_analyzer(request).await?;
+        Ok(Response::new(response.into()))
```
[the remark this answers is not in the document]
colin-grapl: We could macro them but I don't see much point, they could just as easily diverge in the future as the RPC logic and the internal queue logic diverge. Right now all of this was designed at once so everything is very one to one, but I doubt that will always hold true.

They should be traced though.

src/rust/rust-proto/src/plugin_work_queue.rs
colin-grapl: Honestly not sure. It depends on the context of the caller and how they handle the error (as well as all callers above them). We can't guarantee it, we could possibly expect it.

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
```diff
@@ -0,0 +1,80 @@
+CREATE TYPE status_t AS ENUM ('enqueued', 'failed', 'processed');
```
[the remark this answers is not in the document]
colin-grapl: Not really. The only other state in that regard is the visibility, which I can't express in a sql enum.

src/rust/plugin-work-queue/README.md
```diff
@@ -0,0 +1,106 @@
...
+below.
+
+`try_count` on every receive we increment `try_count` to indicate another attempt to process this message
+
+
+## Algorithm
+The query for processing a message is:
+
+1. Find the oldest message that
+    1. Has not aged out (1 day)
```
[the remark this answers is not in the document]
colin-grapl: I don't think retention has been documented, but messages won't be picked up after 1 day, and they'll be dropped entirely after 1 month.

src/rust/plugin-work-queue/README.md
colin-grapl: I would rather punt on the "retry count exceeded" until another PR. This PR is already large enough. I've documented this in the README.

src/rust/plugin-work-queue/README.md
colin-grapl: I've attempted to clarify some of this in the readme.

src/rust/plugin-work-queue/src/psql_queue.rs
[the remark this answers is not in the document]
christophermaier: Unfortunately, it doesn't look like it. There _can_ be ordering dependencies for derives (I think a particular derive can only see derives that come before it) that can have implications for more complicated derives. In general, they probably _should_ be independent, though, but the `rustfmt` team seems pretty solid in _not_ implementing an ordering ability.

src/rust/plugin-work-queue/src/psql_queue.rs
christophermaier: Yeah, I think we resolved this in Slack 👍

src/rust/rust-proto/src/plugin_work_queue.rs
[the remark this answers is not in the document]
christophermaier: Sounds like a good idea.

src/rust/rust-proto/src/plugin_work_queue.rs
christophermaier: :tada:

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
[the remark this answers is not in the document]
christophermaier: I like `current_status`.

src/rust/plugin-work-queue/migrations/1640037910_plugin_registry_migration.sql
christophermaier: Ah, right, `OR REPLACE`.

Yeah, I just reflexively add idempotency clauses everywhere when writing SQL. It's also handy when you're iterating locally on a file of DDL; you can just edit and reload.

christophermaier (approved):
Looks good. There's more stuff to implement, of course, but I think this is a good foundation from which to build.

Thanks for putting up with all my comments 😅

pulumi/grapl/__main__.py
```diff
@@ -379,6 +379,15 @@ def main() -> None:
...
 
+        plugin_work_queue_postgres = Postgres(
+            name="plugin-work-queue",
+            subnet_ids=subnet_ids,
+            vpc_id=vpc_id,
+            availability_zone=availability_zone,
+            nomad_agent_security_group_id=nomad_agent_security_group_id,
+        )
+
+        pulumi.export("kafka-endpoint", "dummy_value_while_we_wait_for_kafka")
```
wimax-grapl, line 390: Remove - this is superseded by the next line

wimax-grapl (approved):

PR #1326, opened 2021-12-22
Increase grapl-core nomad job timeout conservatively for aws-deploys

wimax-grapl:
The default is 5m.
Sometimes, the update step in pipeline.provision.yml fails, e.g. https://grapl-internal.slack.com/archives/C024C7TMJ2J/p1640036760073400

Unfortunately we don't really have the setup right now to debug why it's failing.
- the `pulumi update` happens in the Buildkite plugin
- This makes it difficult to do a "finally" step after the `pulumi update` - in this case, the finally step would be to run a "for dump artifacts" step like .buildkite/scripts/grapl_testing_stack/run_parameterized_job.sh does
- there's some talk about this here https://github.com/buildkite/feedback/issues/133

I'm taking a conservative step by bumping it to 8m for now

### DIFF
--- a/pulumi/infra/nomad_job.py
+++ b/pulumi/infra/nomad_job.py
@@ -32,8 +32,8 @@ class NomadJob(pulumi.ComponentResource):
             detach=False,
             # Purge job from Nomad servers after a `pulumi destroy`
             purge_on_destroy=True,
-            opts=pulumi.ResourceOptions(
-                parent=self, custom_timeouts=opts.custom_timeouts if opts else None
+            opts=pulumi.ResourceOptions.merge(
+                opts, pulumi.ResourceOptions(parent=self)
             ),
         )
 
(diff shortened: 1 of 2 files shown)


pulumi/infra/nomad_job.py
```diff
@@ -28,11 +28,13 @@ def __init__(
             resource_name=f"{DEPLOYMENT_NAME}-{name}-job",
             jobspec=self._file_contents(str(jobspec)),
             hcl2=nomad.JobHcl2Args(enabled=True, vars=self._fix_pulumi_preview(vars)),
-            opts=pulumi.ResourceOptions(parent=self),
             # Wait for all services to become healthy
             detach=False,
             # Purge job from Nomad servers after a `pulumi destroy`
             purge_on_destroy=True,
+            opts=pulumi.ResourceOptions(
+                parent=self, custom_timeouts=opts.custom_timeouts if opts else None
```
twunderlich-grapl, line 36: What may be better than is something like
`pulumi.ResourceOptions.merge(opts, pulumi.ResourceOptions(parent=self))` that way if anything else is passed to opts it gets inherited by this as well.

wimax-grapl: i like it

twunderlich-grapl (approved):
LGTM.

PR #1327, opened 2021-12-22
Remove unused Pulumi constructs from the fargate days

wimax-grapl:
delete x3
theyre not imported from __main__ or anything


twunderlich-grapl (approved):
Can confirm that we're no longer using these

PR #1329, opened 2021-12-23
Destroy stateful resources one-by-one, and allow for failures.

wimax-grapl:

I noticed an issue starting around https://buildkite.com/grapl/grapl-provision/builds/201

My suspicion of what happened:
- a previous `pulumi update` did not succeed in creating grapl-core
- now, running destroy_stateful_resources on target urns [resource1, resource2, grapl-core] fails because, well, grapl-core never even came up beforehand

As such, I am making destroy_stateful_resources *best effort*. This is a short term solution until we have multitenancy IDs anyway.

### DIFF
--- a/.buildkite/scripts/grapl_testing_stack/destroy_stateful_resources.sh
+++ b/.buildkite/scripts/grapl_testing_stack/destroy_stateful_resources.sh
@@ -48,14 +48,26 @@ urns=$(pulumi stack output stateful-resource-urns --stack="${stack}")
 mapfile -t target_args < <(echo "${urns}" | jq -r '. | map("--target=" + .) | .[]')
 
 echo -e "--- :pulumi: Destroying stateful resources for ${stack}"
-pulumi destroy \
-    --cwd="${project_dir}" \
-    --stack="${stack}" \
-    --show-replacement-steps \
-    --non-interactive \
-    --diff \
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/grapl_testing_stack/destroy_stateful_resources.sh
```diff
@@ -48,14 +48,26 @@ urns=$(pulumi stack output stateful-resource-urns --stack="${stack}")
...
+            --show-replacement-steps \
+            --non-interactive \
+            --diff \
+            --yes \
+            --refresh \
+            --message="Destroying stateful resources" \
+            --target-dependents \
+            "${target_arg}"
+    )
+done
```
wimax-grapl, line 73: will the subshell work the way I want it to here?
I want to allow any failures inside the () to be okay; but I'm worried it'll be like "oh the subshell returned nonzero, so, that's a Command That Failed"

wimax-grapl: tested it and sure enough this never gets to hey :) 
```
set -euo pipefail
(
  set +e
  exit 1
)
echo "hey"
```

twunderlich-grapl (approved):

PR #1333, opened 2021-12-26
Grapl aws crate

colin-grapl:
802
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/802

Adds a grapl-aws crate.

The crate does a few things:
1. Provides a resource oriented API. So instead of a general 's3 client' you have buckets, which are tied to a specific bucket identity and owned. This prevents cases where a bucket is deleted, then an attacker creates the bucket again, and we start reading from /writing to that bucket.
2. Provides a capability oriented Bucket so that 'read', 'write', 'list' permissions are encoded at the type level. Object capabilities are a security boundary for cases where an attacker has control over data or control flow without violating memory safety or the program's allowed control flow graph.
3. Provides the basis for bucket encryption.

Just a basic integration test so far that probably doesn't pass.


src/rust/grapl-aws/src/s3/client.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[derive(Clone)]
+pub struct S3Client<S>
+where
+    S: S3Common,
+{
+    #[cfg(not(test))]
+    pub(crate) s3_client: S,
+    #[cfg(test)]
+    pub s3_client: S,
```
christophermaier, line 13: Can we call this member something like `inner`? Otherwise we end up with code like `s3_client.s3_client.do_something()`, which looks kinda weird and makes you do a double take.

christophermaier (approved):
Pretty cool 👍

PR #1334, opened 2021-12-27
Pre-commit hook to autoformat code

ghost:
https://github.com/grapl-security/issue-tracker/issues/805

This adds a pre-commit hook that autoformats code on changed files, without having to run make format. This does add ~2 seconds to each commit, but that seems a reasonable tradeoff considering it eliminates having to manually run a step, eliminates the mental overhead of thinking about that step, and reduces the number of commits that are just formatting changes.
I've also added this to the chromeos install script so that this will just work for new hires or new laptops.

1. Ran the commands manually to install the hook
2. Made some test commits and confirmed that they autoformatted correctly

### DIFF
--- a/etc/hooks/pre-commit.sh
+++ b/etc/hooks/pre-commit.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+FILES=$(git diff --cached --name-only --diff-filter=ACMR | sed 's| |\\ |g')
+[ -z "$FILES" ] && exit 0
+
+# Prettify all selected files
+echo "$FILES" | xargs ./pants fmt
+
+# Add back the modified files to staging
+echo "$FILES" | xargs git add
+
+exit 0
+
+# This runs pants formatting and adds it to the commit
+#./pants --changed-since=HEAD fmt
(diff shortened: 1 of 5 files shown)


etc/hooks/pre-commit.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+FILES=$(git diff --cached --name-only --diff-filter=ACMR | sed 's| |\\ |g')
```
wimax-grapl, line 3: could you comment what this sed does

twunderlich-grapl: Added a comment. Its auto escaping any files with spaces in it

etc/hooks/pre-commit.sh
```diff
@@ -0,0 +1,16 @@
...
+# Prettify all selected files
+echo "$FILES" | xargs ./pants fmt
+
+# Add back the modified files to staging
+echo "$FILES" | xargs git add
+
+exit 0
+
+# This runs pants formatting and adds it to the commit
+#./pants --changed-since=HEAD fmt
```
wimax-grapl, line 16: What's this comment about

wimax-grapl (approved):

PR #1338, opened 2022-01-03
setup_chromeos: fix nvm and pyenv installs on base chromeos; enable -euo pipefail

wimax-grapl:
I'm fixing some issues I discovered with a fresh ChromeOS install. 

- Number one is that issues silently fail due to lack of -euo pipefail.
- nvm didn't install correctly
- pyenv didn't install correctly
- added a docker-compose install via pip

### DIFF
--- a/etc/chromeos/lib/installs.sh
+++ b/etc/chromeos/lib/installs.sh
@@ -34,6 +34,9 @@ install_docker() {
     echo "Install docker"
     curl -sSL https://get.docker.com/ | sh
     sudo usermod -a -G docker "$USER"
+
+    echo "Install docker-compose"
+    pip3 install docker-compose
 }
 
 install_rust_and_utilities() {
@@ -67,7 +70,7 @@ install_pyenv() {
 
     curl -L https://raw.githubusercontent.com/pyenv/pyenv-installer/master/bin/pyenv-installer | bash
 
-    # This function is stolen from the 
+    # This function is stolen from the
     # "If your ~/.profile sources ~/.bashrc (Debian, Ubuntu, Mint)"
     # section of https://github.com/pyenv/pyenv/blob/master/README.md
     setup_pyenv_on_path() {
@@ -80,7 +83,7 @@ install_pyenv() {
         a \
         ' -e ':a' -e '$!{n;ba};}' ~/.profile
         source ~/.profile
-        echo 'eval "$(pyenv init --path)"' >>~/.profile
+        echo 'eval "$(pyenv init --path)"' >> ~/.profile
         echo 'eval "$(pyenv init -)"' >> ~/.bashrc
     }
     setup_pyenv_on_path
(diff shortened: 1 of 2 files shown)


etc/chromeos/lib/installs.sh
```diff
@@ -97,6 +139,7 @@ install_utilities() {
 install_hashicorp_tools() {
     echo "Installing hashicorp tools: consul nomad packer"
     curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add -
+    sudo apt-get install -y lsb-release software-properties-common # for `lsb_release` and `apt-add-repository``
```
twunderlich-grapl, line 142: This probably makes more sense in the install_build_tooling function

etc/chromeos/lib/installs.sh
```diff
@@ -32,6 +40,9 @@ install_docker() {
     echo "Install docker"
     curl -sSL https://get.docker.com/ | sh
     sudo usermod -a -G docker "$USER"
+
+    echo "Install docker-compose"
+    pip3 install docker-compose
```
twunderlich-grapl, line 45: We should probably be installing docker-compose via curl, instead of via pip since docker-compose version 2 was rewritten in golang

etc/chromeos/lib/installs.sh
colin-grapl: apt-get is deprecated and apt-key is too

You can do this instead like so:
```
# Let's get Consul and Nomad installed
RUN curl -s https://apt.releases.hashicorp.com/gpg | \
    gpg --no-default-keyring --keyring gnupg-ring:/etc/apt/trusted.gpg.d/hashicorp-apt.gpg --import && \
    chmod 644 /etc/apt/trusted.gpg.d/hashicorp-apt.gpg
```

etc/chromeos/lib/installs.sh
```diff
@@ -64,64 +90,98 @@ install_pyenv() {
...
+    echo_banner "Installing CNI plugins required for Nomad bridge networks"
     sudo mkdir -p /opt/cni/bin
     cd /opt/cni/bin || exit 2
     VERSION=$(get_latest_release containernetworking/plugins)
-    sudo wget "https://github.com/containernetworking/plugins/releases/download/$VERSION/cni-plugins-linux-amd64-$VERSION.tgz"
-    curl -sL https://github.com/containernetworking/plugins/releases/latest | grep "linux-amd64" | grep -v "sha" | grep -v "span" | awk '{ print $2 }' | awk -F'"' '{ print "https://github.com"$2 }' | wget -qi -
-    sudo tar -xf "cni-plugins-linux-amd64-$VERSION.tgz"
-    sudo rm "cni-plugins-linux-amd64-$VERSION.tgz"
+    TGZ_NAME="cni-plugins-linux-amd64-${VERSION}.tgz"
+    sudo wget "https://github.com/containernetworking/plugins/releases/download/$VERSION/${TGZ_NAME}"
```
twunderlich-grapl, line 171: You may as well wrap $VERSION to ${VERSION}

twunderlich-grapl (approved):
Looks good, thanks for the fixup. One minor comment that's not necessary to merge it in

PR #1347, opened 2022-01-06
etc hooks README is formatted differently by my formatter for some reason

wimax-grapl:
For some reason, `make format` formats my README differently from origin/main. Prettier is pinned. Very spooky.


christophermaier (approved):
👻

PR #1352, opened 2022-01-10
Don't close Dependabot PRs as stale

christophermaier:
We didn't like that this could allow Dependabot PRs to "disappear",
particularly over long holidays like Christmas.

While we should ensure proper assignment of such PRs to individuals,
this is an easy step we can take to ensure nothing falls through the
cracks.


twunderlich-grapl (approved):
Confirmed that's the label that Dependabot adds

wimax-grapl (approved):

PR #1354, opened 2022-01-11
Adding a debug container to nomad

wimax-grapl:
We previously had a `debugger` container that I'm co-opting for "a Nomad service that connects to arbitrary upstreams for debug purposes." 

Basically, I'm trying to debug why an upstream `nomad` doesn't work, and we didn't really have a good container to do that in. I've run in to this "I want an arbitrary container to run `curl`s against" desire anyway.

See https://discuss.hashicorp.com/t/can-i-use-consul-connect-to-proxy-into-the-nomad-service-itself-as-an-upstream/34105 for more context.

### DIFF
--- a/debugger/attach.sh
+++ b/debugger/attach.sh
@@ -15,4 +15,4 @@ docker run \
     --cap-add sys_ptrace \
     --volume="$(pwd):/from-host" \
     --workdir="/from-host" \
-    grapl/debugger
+    debugger:dev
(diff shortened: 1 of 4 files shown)


twunderlich-grapl (approved):

wimax-grapl:
gonna put this on ice for a bit discussion here 
https://grapl-internal.slack.com/archives/C017F10727P/p1641915282010500

PR #1360, opened 2022-01-11
Grapl Testing Stack dump_artifacts is now a Metahook post-command. Enabled on "pulumi update" in provision.

wimax-grapl:
For some reason, the pulumi update in `pipeline.provision.yml` times out beyond 8m. Moreover, we have zero visibility into _why_ this happens. We previously tried fixing this by bumping from 5 to 8 minutes in https://github.com/grapl-security/grapl/pull/1326

Now, thanks to metahook, we can execute a cleanup command  - `dump_artifacts` - after the command. Hooray!  We'll now be able to see which task within `grapl-core` may be slowing us down.

### DIFF
--- a/.buildkite/pipeline.provision.yml
+++ b/.buildkite/pipeline.provision.yml
@@ -53,12 +53,17 @@ steps:
       - grapl-security/vault-env#v0.1.0:
           secrets:
             - PULUMI_ACCESS_TOKEN
+      - improbable-eng/metahook#v0.4.1:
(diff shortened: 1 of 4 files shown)


christophermaier (approved):
Looks good!

My only concern was about assume-role shenanigans with the artifact uploading, but that should be sufficiently addressed by https://github.com/grapl-security/pipeline-infrastructure/pull/178; if it's not, please let me know!

PR #1361, opened 2022-01-11
Some Pulumi-centric cleanups

christophermaier:
Just removing some more old stuff.


wimax-grapl (approved):
We no longer `docker buildx bake` as part of pulumi - thank god - so yeah this looks good 2 me

PR #1369, opened 2022-01-12
Add config file to manage dependabot settings

ghost:
https://github.com/grapl-security/issue-tracker/issues/815

This adds a config file for dependabot to make management easier.
Within this file, we specifically add management of python and rust dependencies while keeping the js dependencies for graphql_endpoint

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+# Config file for dependabot security alerts and updates. For config options see
+# https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates
+version: 2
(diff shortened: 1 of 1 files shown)


christophermaier (approved):

inickles-grapl:
Would this change Dependabot to ignore other locations we care about? Like: src/js/engagament_view, src/python

jgrillo-grapl (approved):

twunderlich-grapl:
> Would this change Dependabot to ignore other locations we care about? Like: src/js/engagament_view, src/python

Dependabot looks for the dependency files so src/python should be ok unless someone's stuffed a requirements file somewhere unusual. I'll add the engagement_view, I wasn't sure if we were still using it.

PR #1283, opened 2021-11-23
Consume ccloud-bootstrap stackreference

jgrillo-grapl:
### Which issue does this PR correspond to?
https://github.com/grapl-security/issue-tracker/issues/562

### What changes does this PR make to Grapl? Why?

This PR adds automation to consume a `StackReference` to the `confluent-cloud-infrastructure` stack and create Kafka topics and ACLs based on the data contained in those stack outputs.

### How were these changes tested?

WIP: spinning up resources in AWS, making sure that topics and ACLs are created.

### See also 
https://github.com/grapl-security/confluent-cloud-infrastructure/pull/1

PR #1395, opened 2022-01-13
Internal 560 ccloud stackreference

jgrillo-grapl:
See https://github.com/grapl-security/grapl/pull/1283

See also: https://github.com/grapl-security/confluent-cloud-infrastructure/pull/1

This PR adds automation to consume a StackReference to the confluent-cloud-infrastructure stack, and consume kafka credentials for the corresponding grapl deployment stack.

Deploy to AWS sandbox and make sure stackreference consumed successfully


pulumi/infra/kafka.py
```diff
@@ -1,13 +1,137 @@
...
+        opts: Optional[pulumi.ResourceOptions] = None,
+    ):
         super().__init__("grapl:Kafka", name=name, props=None, opts=opts)
-        metrics_topic = kafka.Topic(
-            "metrics-topic", name="metrics", replication_factor=1, partitions=1
-        )
+
+        confluent_stack_output = StackReference(
+            "grapl/ccloud-bootstrap/ccloud-bootstrap"
+        ).get_output("confluent")
```
wimax-grapl, line 131: Shouldn't this be different per-stack? Most of our other StackReferences are like
```
    consul_stack = pulumi.StackReference(f"grapl/consul/{pulumi.get_stack()}")
```

so that `grapl/grapl/wimax-sandbox` hooks up to `grapl/consul/wimax-sandbox`

wimax-grapl: ok, after now looking at ccloud bootstrap I now see that the per-user stuff is handled at the Confluent level, not the pulumi stack level

wimax-grapl: could we perhaps add a comment about that

pulumi/infra/kafka.py
```diff
@@ -1,13 +1,137 @@
...
+from typing import Any, Iterable, Mapping, Optional, Sequence, Tuple, cast
+
+from infra.config import LOCAL_GRAPL
+from pulumi.stack_reference import StackReference
 
 import pulumi
 
+# this list of service names must match those in the
+# confluent-cloud-infrastructure project:
+# TODO: link
```
christophermaier, line 14: TODO

pulumi/infra/kafka.py
```diff
@@ -1,13 +1,137 @@
...
+
+
+@dataclasses.dataclass
+class CredentialOutput:
+    service_account_id: str
+    api_key: str
+    api_secret: str
+
+    @staticmethod
+    def from_json(json_: Mapping[str, str]) -> CredentialOutput:
```
christophermaier, line 32: Could we call it `input` instead of `json_`? The trailing underscore looks weird here 😅 

(Likewise elsewhere in this PR.)

pulumi/infra/kafka.py
```diff
@@ -1,13 +1,137 @@
...
+
+    def get_environment(self, environment_name: str) -> EnvironmentOutput:
+        if environment_name in self.environments:
+            return self.environments[environment_name]
+        else:
+            raise KeyError(f"{environment_name} does not exist")
+
+    @staticmethod
+    def from_json(
+        json_: pulumi.Output[Mapping[str, Any]]
```
christophermaier, line 109: Why is this a `pulumi.Output[Mapping[...]]`, but all the other `from_json` methods take just a `Mapping[...]`?

pulumi/infra/kafka.py
```diff
@@ -1,13 +1,137 @@
...
+            ),
+            services={
+                k: ServiceOutput.from_json(v) for k, v in json_["services"].items()
+            },
+            topics={k: TopicOutput.from_json(v) for k, v in json_["topics"].items()},
+        )
+
+
+@dataclasses.dataclass
+class ConfluentOutput:
```
christophermaier, line 98: Curious why all of these dataclasses are named like `FooOutput`; it seems like just `Foo` would suffice. Am I missing something?

pulumi/infra/kafka.py
```diff
@@ -1,13 +1,137 @@
...
         super().__init__("grapl:Kafka", name=name, props=None, opts=opts)
-        metrics_topic = kafka.Topic(
-            "metrics-topic", name="metrics", replication_factor=1, partitions=1
-        )
+
+        confluent_stack_output = StackReference(
+            "grapl/ccloud-bootstrap/ccloud-bootstrap"
+        ).get_output("confluent")
+
+        assert confluent_stack_output is not None
```
christophermaier, line 133: If you use [`require_output`](https://github.com/pulumi/pulumi/blob/722d8344e1013d8bf1e68f6394011e1fee12e2b2/sdk/python/lib/pulumi/stack_reference.py#L74-L85) instead of `get_output` on line 131 above, you could leave this assertion out.

pulumi/infra/kafka.py
jgrillo-grapl: `input` is also a thing in python so it would need a trailing underscore as well (I mean not strictly... you can re-bind any name you want, within the bounds of syntax, but it's not usually advisable).. I'll think of something.

pulumi/infra/kafka.py
jgrillo-grapl: This is the outermost layer of a nested structure, so only this method receives an actual `Output[T]`. All the others are operating inside an `.apply(..)`.

pulumi/infra/kafka.py
jgrillo-grapl: This is a fairly temporary situation (I hope). It will be fixed when this issue is addressed https://github.com/grapl-security/issue-tracker/issues/822

pulumi/infra/kafka.py
christophermaier: `json_input`, to cover all our bases 😅 

It's not a big deal either way, though; whatever you end up deciding is cool 👍

pulumi/infra/kafka.py
christophermaier: Ah... thought it might be something like that.

Might be worth adding a comment to that effect somewhere in this file.

christophermaier (approved):
Looks good, pending the resolution of Max's note about the test user password.

pulumi/grapl/__main__.py
```diff
@@ -79,8 +81,7 @@ def main() -> None:
     # TODO: temporarily disabled until we can reconnect the ApiGateway to the new
     # web UI.
     # jwt_secret = JWTSecret()
-
-    test_user_password = TestUserPassword()
+    # test_user_password = TestUserPassword()
```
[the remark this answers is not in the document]
jgrillo-grapl: fixed

pulumi/grapl/__main__.py
```diff
@@ -198,13 +200,18 @@ def main() -> None:
     )
     py_log_level = "DEBUG"
 
-    nomad_grapl_core_timeout = "5m"
+    # We've seen some potentially false failures from the default 5m timeout.
+    nomad_grapl_core_timeout = "8m"
```
wimax-grapl, line 204: just a heads up, you may need to rebase - this has indeed gone back to 5m

wimax-grapl (approved):

PR #1399, opened 2022-01-18
Try to fix the postgres database name issue where it was not alphanumeric

wimax-grapl:
turns out our understanding of `name=` was not what I thought it was

### DIFF
--- a/pulumi/infra/postgres.py
+++ b/pulumi/infra/postgres.py
@@ -5,6 +5,7 @@ from typing import List, Optional
 
 import pulumi_aws as aws
 import pulumi_random as random
+from importlib_metadata import re
 from packaging.version import parse as version_parse
 
 import pulumi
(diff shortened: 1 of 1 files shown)


jgrillo-grapl (approved):

pulumi/infra/postgres.py
```diff
@@ -104,9 +105,27 @@ def __init__(
...
+        #  A database name is not required when creating a DB instance.
+        #  Databases hosted by the same DB instance must have a unique name within
+        #  that instance."
+        # Since we're using 1 database per instance right now, I'm going to
+        # hardcode it to the default value of `postgres` (like the postgres
+        # docker image does).
+        #
+        # As for giving the instance a name we can see on the console, that's
+        # the `identifier=`.
+        database_name = "postgres"
```
christophermaier, line 119: While this should be technically OK, I think it would be good to give it a unique name, like `pluginregistry`, `PluginRegistry`, etc.

christophermaier (approved):

pulumi/infra/postgres.py
wimax-grapl: Potentially. I'm not _actually_ sure how important these "name"s will be in the long run if we just have one db per host.

PR #1403, opened 2022-01-20
Skeleton of the DeployPlugin rpc

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/722

This is a very barebones take on DeployPlugin.

It:
- creates a namespace
- constructs JSON job from HCL2 + variables
- submits that job

- has a test that the above three do not fail

It does not:
- interact with the plugin registry at all yet
- the submitted job doesn't run yet (mostly because the 'artifact' is the string "TODO"


src/rust/plugin-registry/src/deploy_plugin.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use crate::{
+    nomad_cli,
+    nomad_client,
+    static_files,
+};
+
+/// https://github.com/grapl-security/grapl-rfcs/blob/main/text/0000-plugins.md#deployplugin-details
+#[tracing::instrument(skip(client, plugin_id), err)]
+pub async fn deploy_plugin(
```
wimax-grapl, line 11: # This is the core of this PR

src/rust/plugin-registry/src/deploy_plugin.rs
```diff
@@ -0,0 +1,44 @@
...
+    client: nomad_client::NomadClient,
+    plugin_id: uuid::Uuid,
+) -> Result<(), nomad_client::NomadClientError> {
```
wimax-grapl, line 14: I couldn't quite figure out how to structure the Errors in this whole crate.

deploy_plugin can have errors from Nomad Client but also from Nomad CLI.

Should it just be, everything in this service gets normalized to a PluginRegistryServiceError?

src/rust/plugin-registry/src/nomad_client.rs
```diff
@@ -16,11 +19,22 @@ pub struct NomadClientConfig {
...
+/// A thin wrapper around the nomad_client_gen with usability improvements.
 pub struct NomadClient {
     pub internal_config: InternalConfig,
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum NomadClientError {
+    // Quick note: the error enums in the generated client *are not* std::error::Error
+    #[error("ParseHclError {0:?}")]
+    ParseHclError(#[from] nomad_cli::ParseHclError),
```
wimax-grapl, line 31: This really shouldn't be here, since NomadClient doesn't call nomad_cli::

wimax-grapl:
adding comments

src/rust/Dockerfile
```diff
@@ -19,8 +19,19 @@ RUN --mount=type=cache,target=/var/lib/apt/lists \
...
+        zlib1g-dev=1:1.2.11.dfsg-2 \
+        curl=7.74.0-1.3+deb11u1 \
+        unzip=6.0-26
+
+# Grab a Nomad binary, which we use in plugin-registry for parsing 
+# HCL2-with-variables into JSON.
+WORKDIR /nomad
+RUN <<EOF
+set -euo pipefail
+curl -O https://releases.hashicorp.com/nomad/1.2.4/nomad_1.2.4_linux_amd64.zip
```
colin-grapl, line 31: When using curl ideally you would specify:
`--proto '=https' --tlsv1.3 -sSf `

src/rust/plugin-registry/src/deploy_plugin.rs
colin-grapl: Yeah, generally if you have errors from two different modules you create a third error that encapsulates them.

src/rust/plugin-registry/src/deploy_plugin.rs
```diff
@@ -0,0 +1,44 @@
...
+    // Convert HCL to JSON Job model
+    let job_file_hcl = static_files::PLUGIN_JOB;
+    let job_file_vars: nomad_cli::NomadVars = HashMap::from([
+        ("plugin_id".to_owned(), format!("{}", plugin_id)),
```
colin-grapl, line 18: Instead of format you can almost certainly just put `.to_string()`

src/rust/plugin-registry/src/deploy_plugin.rs
```diff
@@ -0,0 +1,44 @@
...
+        ("tenant_id".to_owned(), "TODO".to_owned()),
+        ("plugin_artifact_url".to_owned(), "TODO".to_owned()),
+        ("aws_account_id".to_owned(), "TODO".to_owned()),
+    ]);
+    let job = nomad_cli::parse_hcl2(job_file_hcl, job_file_vars)
+        .map_err(nomad_client::NomadClientError::from)?;
+
+    // TODO: Should I "nomad job plan" to make sure we have enough memory/cpu for the task?
```
colin-grapl, line 26: Yep, probably

src/rust/plugin-registry/src/deploy_plugin.rs
```diff
@@ -0,0 +1,44 @@
...
+    // Deploy namespace
+    let namespace_name = "todo-fill-this-in-with-real-plugin-name";
+    client.create_namespace(namespace_name).await?;
+    // TODO: What if the namespace already exists?
+
+    // Start the job
+    let _job = client
+        .create_job(job, Some(namespace_name.to_owned()))
+        .await?;
+    // There's no guarantee that the job is *healthy*, just that it's been deployed.
```
colin-grapl, line 37: I think that's fine. We can add another API like 'CheckPlugin' or something for monitoring. As long as the deploy is successful I think this RPC has done its job.

src/rust/plugin-registry/src/nomad_cli.rs
```diff
@@ -0,0 +1,62 @@
...
+    #[error("IOError {0:?}")]
+    IOError(#[from] std::io::Error),
+    #[error("DeserializeJsonError")]
+    DeserializeJsonError(#[from] serde_json::Error),
+}
+
+pub type NomadVars = HashMap<String, String>;
+
+/// Convert HCL2 + variables into JSON
+fn parse_hcl2_to_json(hcl2: &str, vars: HashMap<String, String>) -> Result<String, std::io::Error> {
```
colin-grapl, line 21: Looks like an area where you'd want to use your NomadVars type alias.

src/rust/plugin-registry/src/nomad_cli.rs
```diff
@@ -0,0 +1,62 @@
...
+    // Write our static file to a temporary file
+    let mut file = NamedTempFile::new()?;
+    file.write_all(hcl2.as_bytes())?;
+    let path = file.into_temp_path(); // Useful for when another process must read the file
+
+    // Feed it to the Nomad binary.
+    // Assume nomad is available on $PATH
+    let mut command = Command::new("nomad");
+    command.arg("job").arg("run").arg("-output");
```
colin-grapl, line 30: There's `.args` for this. Also can be one line by just chainging.

src/rust/plugin-registry/src/nomad_cli.rs
```diff
@@ -0,0 +1,62 @@
...
+    for (key, value) in &vars {
+        command.arg("-var").arg(format!("{}={}", key, value));
+    }
+
+    // Specify filename
+    command.arg(&path);
+
+    let result = command.output()?;
+
+    Ok(String::from_utf8(result.stdout).expect("Invalid UTF8"))
```
colin-grapl, line 41: Maybe from_utf8_lossy is the right call her rather than crashing.

src/rust/plugin-registry/src/nomad_cli.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/// Convert HCL2 + variables into a Job
+#[tracing::instrument(skip(hcl2, vars), err)]
+pub fn parse_hcl2(hcl2: &str, vars: HashMap<String, String>) -> Result<models::Job, ParseHclError> {
```
colin-grapl, line 46: Another case for NomadVars

src/rust/plugin-registry/src/nomad_client.rs
```diff
@@ -16,11 +19,22 @@ pub struct NomadClientConfig {
     nomad_service_address: String,
```
colin-grapl, line 19: SocketAddr over String

src/rust/plugin-registry/src/static_files/plugin.nomad
```diff
@@ -0,0 +1,180 @@
...
+  description = "The number of instances of the plugin to run."
+}
+
+variable "aws_account_id" {
+  type        = string
+  description = "The account ID of the aws account that holds onto the plugin binaries."
+}
+
+locals {
+  shared_key = "grapl_secret-${uuidv4()}"
```
colin-grapl, line 34: We can drop this for now and get to it later.

src/rust/plugin-registry/src/static_files/plugin.nomad
```diff
@@ -0,0 +1,180 @@
...
+      config {
+        image = "grapl/plugin-execution-sidecar"
+        ports = [
+        "plugin_sidecar_grpc_receiver"]
+      }
+
+      env {
+        TENANT_ID     = "${var.tenant_id}"
+        PLUGIN_ID     = "${var.plugin_id}"
+        BOOTSTRAP_KEY = "${local.shared_key}"
```
colin-grapl, line 82: Can drop this I believe. Better to do the key thing later when we have a better grip on nomad + firecracker.

src/rust/plugin-registry/src/static_files/plugin.nomad
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+
+    task "tenant-plugin-bootstrap-sidecar" {
+      driver = "docker"
+
+      template {
+        data        = <<EOH
+      {{ $plugin_id := env "PLUGIN_ID" }}
+      {{ with secret "pki/issue/plugin_service_ca" (printf "common_name=%s.plugins.grapl.com" $plugin_id) "format=pem" }}
```
colin-grapl, line 93: Based on this it looks like this might issue the same certificate for the execution sidecar and the bootstrap sidecar, right? We should issue two separate certificates I think, even if they're both for the same common name.

src/rust/plugin-registry/src/nomad_cli.rs
wimax-grapl: this is due to rust format

src/rust/plugin-registry/src/deploy_plugin.rs
jgrillo-grapl: I think the most important thing is to make sure we don't surface "internal" errors (e.g. nomad CLI/client error text) to consumers of the RPC method. So at the RPC layer we should translate internal errors to something appropriate for public consumption.

jgrillo-grapl (commented):
I realize this isn't part of your PR, but I found the `PluginRegistryServiceClient` terminology very confusing. Which is it? It appears to be a gRPC service, but we're calling it a client? Why?

src/rust/plugin-registry/src/deploy_plugin.rs
wimax-grapl: Yep, that already happens

src/rust/plugin-registry/src/deploy_plugin.rs
wimax-grapl: That'll be a followup, and I'm tracking it as a checkbox in grapl-security/issue-tracker#722

src/rust/plugin-registry/src/deploy_plugin.rs
wimax-grapl: That'll be a followup, and I'm tracking it as a checkbox in grapl-security/issue-tracker#722

src/rust/plugin-registry/src/nomad_cli.rs
wimax-grapl: done

src/rust/plugin-registry/src/nomad_cli.rs
wimax-grapl: fixed

src/rust/plugin-registry/src/nomad_cli.rs
wimax-grapl: done

src/rust/plugin-registry/src/static_files/plugin.nomad
wimax-grapl: So like
pki/issue/plugin_service_execution_ca
pki/issue/plugin_service_bootstrap_ca
?

wimax-grapl:
Perhaps we can simplify it by being like
```
pub struct PluginRegistryServiceClient<T> {
    proto_client: RawProtoPluginRegistryServiceClient<T>,
}
```

colin-grapl (approved):

colin-grapl:
> Perhaps we can simplify it by being like
> 
> ```
> pub struct PluginRegistryServiceClient<T> {
>     proto_client: RawProtoPluginRegistryServiceClient<T>,
> }
> ```

That's what we do elsewhere - check out the plugin-work-queue for example.

jgrillo-grapl (approved):

PR #1407, opened 2022-01-25
Install vault as part of chromeos setup

ghost:
https://github.com/grapl-security/issue-tracker/issues/825

This installs the Hashicorp Vault locally as part of the ChromeOS setup

Installing vault locally grants CLI access to vault (assuming permission to connect to Vault). This will be
primarily used by members of Operations, but could be extended to
general secrets management such as allowing engineers to add secrets
for new services.

Running the command locally and then testing with vault --version


christophermaier (approved):
Yup

colin-grapl (approved):

PR #1408, opened 2022-01-25
We should use opt-level 1 for all of dev

colin-grapl:
None.

One line change. We should use opt-level=1 for all of dev profile, not just test. This way we avoid recompiling.

No additional testing.


jgrillo-grapl (approved):

christophermaier:
Why `1` and not `0`, out of curiosity?

PR #1411, opened 2022-01-25
Remove pulumi preview job from verify pipeline

christophermaier:
As part of a larger push to lock down our CI/CD environment, we are
removing jobs from our verify pipeline that interact with AWS
infrastructure. This pipeline is the only one that runs prior to a
merge (and is thus gated by human reviewers), and the `pulumi preview`
job is the only one that interacts with AWS.


wimax-grapl (approved):

twunderlich-grapl:
This was for issue https://github.com/grapl-security/issue-tracker/issues/830

PR #1413, opened 2022-01-26
Plugin bootstrap

colin-grapl:
738

1. Adds the bootstrap client and server implementations
2. Adds an `init` service, which will perform the actual bootstrap process
3. Adds a systemd service configuration file, which we'll eventually bake into our VM

No additional testing at this time.

### DIFF
--- a/src/rust/plugin-bootstrap/plugin.service
+++ b/src/rust/plugin-bootstrap/plugin.service
@@ -1,0 +1,52 @@
+# Contents of /etc/systemd/system/grapl-plugin.service
+[Unit]
+Description=Plugin
+After=network.target
+
+[Service]
+Type=simple
+Restart=always
+ExecStart=/usr/local/bin/grapl-plugin
+
+# All plugins should listen on port 5555 internally
+SocketBindAllow=tcp:5555
+SocketBindDeny=any
+
+RuntimeDirectory=plugin        # adds /var/run/plugin
+StateDirectory=plugin          # adds /var/lib/plugin
+CacheDirectory=plugin          # adds /var/cache/plugin
+LogsDirectory=plugin           # adds /var/log/plugin
+LoadCredential=CLIENT_CERTIFICATE:/etc/ssl/private/plugin-client-cert.pem
+
+# Some sandboxing
+# See: https://www.freedesktop.org/software/systemd/man/systemd.exec.html
+# Allocate a random user within the range: 61184–65519
+DynamicUser=yes
+CapabilityBoundingSet=
+DevicePolicy=closed
+InaccessiblePaths=-/usr/bin /usr/sbin /mnt /media /var/www
+LockPersonality=yes
+
+NoNewPrivileges=yes
+PrivateDevices=yes
+PrivateMounts=yes
+PrivateTmp=yes
+PrivateUsers=yes
+ProtectClock=yes
+ProtectControlGroups=yes
+ProtectHome=yes
+ProtectHostname=yes
+ProtectKernelLogs=yes
+ProtectKernelModules=yes
+ProtectKernelTunables=yes
+ProtectProc=invisible
+ProtectSystem=strict
+RemoveIPC=yes
+RestrictAddressFamilies=AF_UNIX AF_NETLINK AF_INET AF_INET6
+RestrictNamespaces=yes
+RestrictRealtime=yes
+ProtectHostname=yes
+RestrictSUIDSGID=yes
+SystemCallArchitectures=native
+# Note that ~ inverts the filter, meaning that we default-allow syscalls *except* for those listed
+SystemCallFilter=~@chown @clock @cpu-emulation @debug @ipc @module @mount @obsolete @privileged @raw-io @reboot @resources @setuid @swap
--- a/src/rust/plugin-bootstrap/src/client.rs
+++ b/src/rust/plugin-bootstrap/src/client.rs
@@ -1,5 +1,10 @@
-use rust_proto::plugin_bootstrap::{GetBootstrapInfoRequest, GetBootstrapInfoRequestProto, GetBootstrapInfoResponse, PluginBootstrapServiceClient as PluginBootstrapClientProto, PluginBootstrapDeserializationError};
-
+use rust_proto::plugin_bootstrap::{
+    GetBootstrapInfoRequest,
+    GetBootstrapInfoRequestProto,
+    GetBootstrapInfoResponse,
+    PluginBootstrapDeserializationError,
+    PluginBootstrapServiceClient as PluginBootstrapClientProto,
+};
 use tonic::{
     codegen::{
         Body,
@@ -16,7 +21,6 @@ pub enum PluginBootstrapClientError {
     Status(#[from] tonic::Status),
     #[error("PluginBootstrapDeserializationError")]
     PluginBootstrapDeserializationError(#[from] PluginBootstrapDeserializationError),
-
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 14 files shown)


src/rust/plugin-bootstrap/plugin.service
```diff
@@ -0,0 +1,54 @@
...
+
+# All plugins should listen on port 5555 internally
+SocketBindAllow=tcp:5555
+SocketBindDeny=any
+
+RuntimeDirectory=plugin        # adds /var/run/plugin
+StateDirectory=plugin          # adds /var/lib/plugin
+CacheDirectory=plugin          # adds /var/cache/plugin
+LogsDirectory=plugin           # adds /var/log/plugin
+LoadCredential=CLIENT_CERTIFICATE:/etc/ssl/private/plugin-client-cert.pem
```
inickles-grapl, line 21: Looks like `CLIENT_CERTIFICATE` is never used?

src/rust/plugin-bootstrap/plugin-bootstrap-init.service
```diff
@@ -0,0 +1,40 @@
+# Contents of /etc/systemd/system/plugin-bootstrap-init.service
+[Unit]
+Description=Plugin Init
+After=network.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/init-grapl-plugin
+User=root
+Group=root
```
inickles-grapl, line 10: Seems a little weird to run as root then set sandboxing stuff below.

Looks like the motivation for using root is to have read access to `/etc/ssl/private/plugin-client-cert.pem`, is that right?

colin-grapl: Yeah, it's in order to have the right file access. I agree that root is overly permissive here, but I figure that we can create a more restricted user in the future.

I *assume* that root here essentially means that escaping the sandboxing is possible, but I don't actually know.

src/rust/plugin-bootstrap/plugin.service
colin-grapl: It'll be used by the plugin code, which isn't part of this PR.

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
```diff
@@ -0,0 +1,41 @@
...
+    )?;
+
+    std::fs::set_permissions(PLUGIN_BINARY_PATH, std::fs::Permissions::from_mode(0o755))?;
+    std::fs::set_permissions(
+        CLIENT_CERTIFICATE_PATH,
+        std::fs::Permissions::from_mode(0o400),
+    )?;
+
+    let config_file = std::fs::File::create(PLUGIN_CONFIG_PATH)?;
+    config_file.set_permissions(std::fs::Permissions::from_mode(0o755))?;
```
inickles-grapl, line 38: This makes the config file executable. I think you want `0o644` or `0o600`.

inickles-grapl (changes requested):
Requesting a small file permission change.

The systemd service files seem fine, other than the running as root + sandboxing note from earlier.

As far as I can tell, there's a single plugin-bootstrap-service per plugin, and the request to get the client TLS certificate is not authenticated. That's seems fine if it's only network accessible by the plugin, just thought I'd point it out as I don't see why that's necessarily the case. Seems like it also means it's not necessary for the init process to fetch the certificate and store on disk, if instead we had plugin library code to fetch on the fly when starting up. Not that we'd want to, just checking my understanding.

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
```diff
@@ -0,0 +1,41 @@
...
+        .get_bootstrap_info(GetBootstrapInfoRequest {})
+        .await?;
+
+    std::fs::write(PLUGIN_BINARY_PATH, plugin_payload.plugin_binary)?;
+    std::fs::write(
+        CLIENT_CERTIFICATE_PATH,
+        client_certificate.client_certificate,
+    )?;
+
+    std::fs::set_permissions(PLUGIN_BINARY_PATH, std::fs::Permissions::from_mode(0o755))?;
```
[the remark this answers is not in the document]
jgrillo-grapl: Wait.. this file was already written on L25.. Do we ever need to write it again? If not then either `0o100` or `0o010` should suffice, I think?

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
jgrillo-grapl: I'm still not clear _who_ actually needs access to these files, and what access they need. Would be helpful to clarify.

jgrillo-grapl (commented):
Some questions

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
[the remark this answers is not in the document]
inickles-grapl: Just noting the user executing the above code is currently root [1], and the plugin is to be executed by a dynamic user [2], provided by systemd.

[1] https://github.com/grapl-security/grapl/pull/1413/files#diff-0781f24f5e0c738f4ac02482ba5498c3804068df157b21b4a6945a8460d99b31R9-R10
[2] https://github.com/grapl-security/grapl/pull/1413/files#diff-d35b307570920bc59095340f673266b7f73e3b38bac0d821223702499ddec833R26

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
colin-grapl: 1. Systemd should be the one executing the binary
2. Systemd and the plugin need to be able to read the certificate
3. Systemd and the plugin need to be able to read the config file

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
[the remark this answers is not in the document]
jgrillo-grapl: Hmm so the user doesn't belong to any particular group, in which case the binary needs to be `0o001` (executable by "other", readable by none, writable by none).. correct?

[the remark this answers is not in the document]
jgrillo-grapl: Another thing that came up in slack is that this "binary" may not be.. a binary (e.g. a compiled computer program). Instead it could be a `.pex` file. I think maybe the word "binary" here could be a little confusing in that case.

jgrillo-grapl (approved):
LGTM overall, some decisions to be made about file permissions but no blockers

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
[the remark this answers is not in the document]
colin-grapl: FWIW the permissions here aren't much of a security control since our isolation boundaries assume the plugin is the malicious actor - the permissions were more to let me execute binaries and that sort of thing.

So we shouldn't worry too much about it and instead just go with something minimal but working.

src/rust/plugin-bootstrap/src/client.rs
```diff
@@ -0,0 +1,76 @@
...
+        let address = std::env::var(ADDRESS_ENV_VAR).expect(ADDRESS_ENV_VAR);
+        Self::from_endpoint(address).await
+    }
+
+    /// Create a client from a specific endpoint
+    #[tracing::instrument(err)]
+    pub async fn from_endpoint(address: String) -> Result<Self, Box<dyn std::error::Error>> {
+        tracing::debug!(message = "Connecting to endpoint");
+
+        // TODO: It might make sense to make these values configurable.
```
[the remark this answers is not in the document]
colin-grapl: That's a copy/pasted todo. When we have a proper GrpcServiceConfig we can do that.

inickles-grapl (approved):

wimax-grapl:
Closes https://github.com/grapl-security/issue-tracker/issues/738
Closes https://github.com/grapl-security/issue-tracker/issues/740

PR #1419, opened 2022-01-28
Refactor PluginRegistry SQL queries into a DbClient. DeployPlugin now consumes DbClient.

wimax-grapl:
Followup to grapl-security/issue-tracker#722

Previously, the Plugin Registry postgres queries were baked into the main controller `server.rs`; this has now been refactored out to a `PluginRegistryDbClient` so that other pieces of the PluginRegistry can reuse it.

Previously, DeployPlugin was just deploying a job with garbage standin data.
Now, it consumes the PluginRegistryDbClient to populate a job with real data from a previous CreatePlugin call.

Integration tests!!!

### DIFF
--- a/src/rust/plugin-registry/src/db_client.rs
+++ b/src/rust/plugin-registry/src/db_client.rs
@@ -4,14 +4,10 @@ use rust_proto::plugin_registry::CreatePluginRequest;
 #[derive(sqlx::FromRow)]
 pub struct GetPluginRow {
     pub plugin_id: uuid::Uuid,
+    pub tenant_id: uuid::Uuid,
     pub display_name: String,
     pub plugin_type: String,
     pub artifact_s3_key: String,
-}
-
-pub struct CreatePluginResponse {
-    /// The identity of the plugin that was created
-    pub plugin_id: uuid::Uuid,
 }
 
 pub struct PluginRegistryDbClient {
@@ -24,9 +20,10 @@ impl PluginRegistryDbClient {
             r"
         SELECT
         plugin_id,
+        tenant_id,
         display_name,
         plugin_type,
-        artifact_s2_key
+        artifact_s3_key
         FROM plugins
         WHERE plugin_id = $0
             ",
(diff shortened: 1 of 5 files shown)


src/rust/plugin-registry/src/db_client.rs
```diff
@@ -0,0 +1,78 @@
...
+        .bind(&request.plugin_type.type_name())
+        .bind(&request.display_name)
+        .bind(&request.tenant_id)
+        .bind(s3_key)
+        .execute(&self.pool)
+        .await
+        .map(|_| ()) // Toss result
+    }
+
+    pub async fn new(postgres_address: &str) -> Result<Self, Box<dyn std::error::Error>> {
```
wimax-grapl, line 67: This might be an odd choice of Error type, but we have two disparate possible Errors here: a Tokio timeout or something from the sqlx migration.

It also kind of hints to the end user that this should be constructed at the top level.

src/rust/plugin-registry/src/deploy_plugin.rs
```diff
@@ -1,35 +1,40 @@
...
+            ),
         ]);
         cli.parse_hcl2(job_file_hcl, job_file_vars)?
     };
 
     // TODO: "nomad job plan" to make sure we have enough memory/cpu for the task
 
     // --- Deploy namespace
-    let namespace_name = "todo-fill-this-in-with-real-plugin-name";
+    let namespace_name = &plugin.display_name; // TODO: Do we need to regex enforce display names?
```
wimax-grapl, line 37: we likely do - nomad throws an error if you try to create a namespace with a _ in it

src/rust/plugin-registry/src/db_client.rs
```diff
@@ -0,0 +1,78 @@
+use grapl_utils::future_ext::GraplFutureExt;
```
wimax-grapl, line 1: The majority of the sqlx stuff in this file is just refactored out of previous code in server.rs.

jgrillo-grapl (approved):

PR #1422, opened 2022-01-28
Lock cache mounts

colin-grapl:
None. Hotfix.

Adds a lock to the cache mounts in src/rust/Dockerfile.

I've added the lock everywhere because, as far as I know, none of the cases where we use cache mounts are safe to share. Plus, even if they are, like in some cases with cargo build, it should be any more efficient to let cargo do the waiting vs docker.

Just rebuild.


wimax-grapl (approved):
I'm not sure we need this for cargo or rustup but whatevs. Let's make sure to keep an eye on build time impact if any.

inickles-grapl:
I think we'd want to apply this across all Dockerfiles, not just the one.

PR #1425, opened 2022-01-28
Add pulumi exports for RDS instances

ghost:
https://github.com/grapl-security/issue-tracker/issues/833

This adds pulumi exports to the AWS section for the rds instances. These exports are relied upon by the integration tests pulumi project, which is causing the Buildkite pipeline to fail.

pulumi up in the grapl project, followed by pulumi up in the integration-tests project

### DIFF
--- a/pulumi/grapl/__main__.py
+++ b/pulumi/grapl/__main__.py
@@ -387,6 +387,34 @@ def main() -> None:
             nomad_agent_security_group_id=nomad_agent_security_group_id,
         )
 
+        pulumi.export(
+            "plugin-registry-db-hostname", plugin_registry_postgres.instance.address
+        )
+        pulumi.export(
+            "plugin-registry-db-port", str(plugin_registry_postgres.instance.port)
+        )
+        pulumi.export(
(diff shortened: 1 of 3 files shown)


pulumi/grapl/__main__.py
```diff
@@ -387,6 +387,34 @@ def main() -> None:
             nomad_agent_security_group_id=nomad_agent_security_group_id,
         )
 
+        pulumi.export(
```
christophermaier, line 390: Are we exporting everything else in this project as single key/value pairs, or consolidating into objects? That is, export an object for, say, "plugin-registry-db" that has "hostname", "port", "username", etc. keys?

christophermaier (approved):

pulumi/grapl/__main__.py
twunderlich-grapl: Well as much as I'd like to do that, the reason why I'm not is because the local postgres Database object doesn't support these attributes.  We could create a plain old python object, but that feels like the same amount of work as just passing in each attribute

PR #1427, opened 2022-01-28
Upgrade MyPy and add type stubs for multiple 3rdparty libraries

wimax-grapl:
I'd hoped that this would mean we could remove the cast in
```
aws_config = cast(aws.config.vars._ExportableConfig, aws.config)
```
but, nope.

Still, this is a good upgrade - these 3rdparty libs now actually have type stubs to check against :+1:


twunderlich-grapl:
@wimax-grapl do we need to update the requirements and constraints files as well?

pants.toml
```diff
@@ -130,7 +130,7 @@ args = [ "--suppress-no-test-exit-code" ]
 
 [mypy]
 lockfile = "3rdparty/python/mypy.lockfile"
-version = "mypy==0.812"
+version = "mypy==0.910"
```
christophermaier, line 133: This is [what Pants uses now](https://www.pantsbuild.org/docs/reference-mypy#advanced-options) by default, so we could arguably remove this line.

By removing this, I think we could get away with [using the lockfile that Pants generates](https://www.pantsbuild.org/docs/reference-mypy#section-lockfile), meaning we could get rid of ours.

pants.toml
```diff
@@ -140,6 +140,14 @@ config = "src/python/mypy.ini"
...
 args = [ "--show-error-codes", "--namespace-packages" ] #, "--verbose" ]
 
+extra_type_stubs = [
+  "types-click==7.1.8",
+  "types-protobuf==3.19.6",
+  "types-redis==4.1.11",
+  "types-requests==2.27.7",
+  "types-toml==0.10.3",
+]
+
```
christophermaier, line 150: Interesting... I wonder if we can add our other typestubs to this list and remove them from requirements.txt. They would need to be manually updated, but this might also speed up other operations because we wouldn't be constantly downloading them when we set up Pants, load up a venv for Pulumi in our containers, etc.

The [documentation](https://www.pantsbuild.org/docs/reference-mypy#section-extra-type-stubs) suggests that we could get away with this if we don't need a type library at runtime.

christophermaier (approved):
Looks good; we may be able to go further, though!

wimax-grapl:
Thomas - no need, since these are done at the plugin level

pants.toml
wimax-grapl: We don't have any other other explicitly defined `types-` in requirements.txt, just constraints (meaning other libraries declare a dependency).

christophermaier: Does it have to be `types-*` only? Do the `mypy_boto_*` libs not count, or are those something different?

wimax-grapl: Oh. Huh! Interesting point. Let me give it a whirl!

PR #1428, opened 2022-01-29
Move the preview integration-test step after the grapl project has been run

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/835

This moves the preview integration-test until after we pulumi up the grapl project.

We're moving this instead of temporarily skipping the step, because we're going to run into similar problems in the future

### DIFF
--- a/.buildkite/pipeline.provision.yml
+++ b/.buildkite/pipeline.provision.yml
@@ -19,20 +19,20 @@ steps:
 
   - wait
 
-  - label: ":pulumi: Preview grapl/integration-tests/testing environment changes in Staging account"
-    plugins:
-      - grapl-security/vault-login#v0.1.0
-      - grapl-security/vault-env#v0.1.0:
-          secrets:
-            - PULUMI_ACCESS_TOKEN
-      - grapl-security/pulumi#v0.1.2:
-          command: preview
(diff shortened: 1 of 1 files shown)


wimax-grapl (approved):

.buildkite/pipeline.provision.yml
```diff
@@ -67,6 +52,21 @@ steps:
...
+            - PULUMI_ACCESS_TOKEN
+      - grapl-security/pulumi#v0.1.2:
+          command: preview
+          project_dir: pulumi/integration_tests
+          stack: grapl/testing
+    agents:
+      queue: "pulumi-staging"
+
+  - wait
+
```
christophermaier, line 69: 🤔 Does previewing the integration tests provide much benefit?

christophermaier (approved):
The logic of the move looks good... I just wonder if we need to keep the preview at all; seems like something we could dispense with.

If not, though, no biggie.

.buildkite/pipeline.provision.yml
twunderlich-grapl: Maybe by making it clear where something fails? I'll keep it for now, but I'm open to removing it

PR #1431, opened 2022-01-30
Have Consul Intentions use the consul provider

ghost:
https://github.com/grapl-security/issue-tracker/issues/835
https://github.com/grapl-security/issue-tracker/issues/837

Allow the passed in consul provider to be used by consul intentions

Previously we were overriding the opts argument for each consul
intention. This wasn't discovered since during testing we wanted the
default localhost:8500, but now that its being executed from a
pipeline, the pipeline uses the wrong address.

### DIFF
--- a/pulumi/infra/consul_intentions.py
+++ b/pulumi/infra/consul_intentions.py
@@ -36,5 +36,7 @@ class ConsulIntentions(pulumi.ComponentResource):
                         kind=intention["Kind"],
                         name=intention["Name"],
                         config_json=json.dumps({"Sources": intention["Sources"]}),
-                        opts=pulumi.ResourceOptions(parent=self),
+                        opts=pulumi.ResourceOptions.merge(
+                            opts, pulumi.ResourceOptions(parent=self)
+                        ),
                     )


pulumi/infra/consul_intentions.py
```diff
@@ -36,5 +36,7 @@ def __init__(
                         kind=intention["Kind"],
                         name=intention["Name"],
                         config_json=json.dumps({"Sources": intention["Sources"]}),
-                        opts=pulumi.ResourceOptions(parent=self),
```
twunderlich-grapl, line 39: So because we're setting the opts here, consul uses the default address (localhost:8500).

pulumi/infra/consul_intentions.py
```diff
@@ -36,5 +36,7 @@ def __init__(
...
+                        opts=pulumi.ResourceOptions.merge(
```
twunderlich-grapl, line 39: This merges the resource options so we can use the provider option we're passing in. That way we use the real address for consul.

colin-grapl (approved):

jgrillo-grapl (approved):

PR #1435, opened 2022-01-31
Pin mypy_boto3 versions

wimax-grapl:
Followup to https://github.com/grapl-security/grapl/pull/1427/files

```
extra_type_stubs = [
  # These could technically get specific versions, but, eh
  "mypy_boto3_cloudwatch",
  "mypy_boto3_dynamodb",
```
famous last words


jgrillo-grapl (approved):
💯

PR #1437, opened 2022-02-01
Remove Jupyter Notebook

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/840

Jupyter notebook is being replaced, most likely by an api. Since this
is now dead code, we're deleting it so we don't need to waste time
building the image or maintaining the code

N/A


christophermaier (commented):
There appear to be some other references (judging from `git grep notebook`), in particular some code in `src/js/engagement_view`.

Probably also need to tweak the CODEOWNERS file, too.

jgrillo-grapl (approved):
👏

christophermaier:
@twunderlich-grapl Looks like this also needs to be rebased on `main`... that would probably address a few of the CI failures.

PR #1438, opened 2022-02-01
Disable coverage in integration tests

wimax-grapl:
its a 2 line change

relates to this weird transient error
```

2:04
18:34:54.82 [ERROR] 1 Exception encountered:

  ValueError: Failed to create directory /mnt/grapl-root/dist/coverage/python: Failed to create dir "/mnt/grapl-root/dist/coverage/python" due to Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
```
@ https://grapl-internal.slack.com/archives/C02JCBKS97V/p1643742241129789

### DIFF
--- a/test/run/integration-tests-from-docker-container.sh
+++ b/test/run/integration-tests-from-docker-container.sh
@@ -13,4 +13,5 @@ run_pants_python_tests() {
 
 # -rA means "report stdout/stderr for all tests, not just failing tests"
 # -m integration_test limits it to integration tests.
-run_pants_python_tests "-m integration_test -rA"
+# --no-cov disables coverage (since it's integration tests, makes sense!)
+run_pants_python_tests "-m integration_test -rA --no-cov"


test/run/integration-tests-from-docker-container.sh
```diff
@@ -13,4 +13,5 @@ run_pants_python_tests() {
 
 # -rA means "report stdout/stderr for all tests, not just failing tests"
```
wimax-grapl, line 14: (yes, `-rA` is the longest form available)

jgrillo-grapl (approved):

test/run/integration-tests-from-docker-container.sh
christophermaier: Pre-empting my usual code review comments, I see :joy:

test/run/integration-tests-from-docker-container.sh
```diff
@@ -13,4 +13,5 @@ run_pants_python_tests() {
...
 # -m integration_test limits it to integration tests.
-run_pants_python_tests "-m integration_test -rA"
+# --no-cov disables coverage (since it's integration tests, makes sense!)
+run_pants_python_tests "-m integration_test -rA --no-cov"
```
christophermaier, line 17: Curious if we should do the same for the e2e tests? Or is that handled some other way?

christophermaier (approved):

test/run/integration-tests-from-docker-container.sh
wimax-grapl: totally different setup unfortunately

PR #1439, opened 2022-02-02
Add a `make regenerate-sqlx-data` target. Add an empty sqlx-data for plugin-registry, which currently doesn't use the macros.

wimax-grapl:
My next task for https://github.com/grapl-security/issue-tracker/issues/722 requires that I add a new sql query, so I figured now was as good a time as ever to finally add this long-overdue script.

One weird aspect of it: each migrate needs its own fresh instance of postgres!
Otherwise you get something like
```

--- Sqlx Preparing Rust libs

 --- Sqlx Prepare on src/rust/plugin-work-queue
Applied 1640037910/migrate plugin work queue migration (944.70359ms)
   Compiling plugin-work-queue v0.1.0 (/home/wimax/src/repos/grapl/src/rust/plugin-work-queue)
    Finished dev [optimized + debuginfo] target(s) in 1.77s
query data written to `sqlx-data.json` in the current directory; please check this into version control

 --- Sqlx Prepare on src/rust/plugin-registry
error: migration 1640037910 was previously applied but is missing in the resolved migrations
```


bin/regenerate_sqlx_data.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
```
christophermaier, line 1: Maybe this should go into `src/rust/bin` instead? Keep it closer to the source?

bin/regenerate_sqlx_data.sh
```diff
@@ -0,0 +1,65 @@
...
+set -euo pipefail
+
+################################################################################
+# Our Rust code uses a library called `sqlx` to interface with Postgres.
+# It *statically* checks the validity of queries by running them against a real,
+# migrated SQL database and storing that validity in sqlx-data.json.
+#
+# Long story short: Changing Rust sql queries? You probably want to run this
+# script with `make regenerate-sqlx-data`.
+################################################################################
```
christophermaier, line 11: @andrea-grapl Do you happen to have that script that we came up with that does this? As I recall, I think we were able to get away with something a little simpler?

bin/regenerate_sqlx_data.sh
wimax-grapl: very reasonable

bin/regenerate_sqlx_data.sh
```diff
@@ -0,0 +1,65 @@
...
+}
+
+# Stop the container if any failures occur.
+trap stop_postgres EXIT
+
+sqlx_prepare src/rust/plugin-work-queue
+sqlx_prepare src/rust/plugin-registry
+
+# Undo the above trap
+trap - EXIT
```
christophermaier: Removing the trap here isn't really necessary

wimax-grapl: Well, it causes a double-`docker stop` in the happy path, which is a little odd. True, the `|| true` will make it not fail, but the echos are a little weird

wimax-grapl:
> It's a bit odd to have this located in src/rust/bin, but then change back to the repository root, only to reference back down into src/rust further down in the script.

So, on the face of it, I agree. However!
One of my great qualms with bash scripting is:
- Scripts that have to be run from one specific directory in order to work correctly
- Scripts that stop working when they're moved around

So this is a hedge against it.

Nonetheless, the top-level `cd` is weird so I'm making it less weird.

etc/chromeos/lib/installs.sh
```diff
@@ -216,3 +216,8 @@ install_git_hooks() {
     GIT_ROOT=$(git rev-parse --show-toplevel)
     ln --symbolic --relative --force "$GIT_ROOT/etc/hooks/pre-commit.sh" "$GIT_ROOT/.git/hooks/pre-commit"
 }
+
+install_sqlx_prepare_deps() {
+    cargo install sqlx-cli --no-default-features --features postgres,rustls
+    sudo apt install --yes netcat # used for `nc`
```
christophermaier, line 222: The `used for nc` comment is superfluous.

christophermaier (approved):

etc/chromeos/lib/installs.sh
wimax-grapl: (I find the discrepancy between the two package names deeply annoying tbh)

PR #1444, opened 2022-02-03
Invoke `curl` in a more secure way

christophermaier:
Basically adds `--proto "=https" --tlsv1.2` to all the `curl`
invocations I could find.

Also translates short options to long options for readability.

### DIFF
--- a/etc/chromeos/lib/installs.sh
+++ b/etc/chromeos/lib/installs.sh
@@ -19,9 +19,12 @@ echo_banner() {
 }
 
 get_latest_release() {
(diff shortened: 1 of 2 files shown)


twunderlich-grapl (approved):

etc/chromeos/lib/installs.sh
```diff
@@ -162,7 +193,12 @@ install_utilities() {
 
 install_hashicorp_tools() {
     echo_banner "Installing hashicorp tools: consul nomad packer"
-    curl --proto '=https' --tlsv1.3 -sSf https://apt.releases.hashicorp.com/gpg |
```
colin-grapl, line 165: It was 1.3 before :P

PR #1445, opened 2022-02-03
Add a `plugin_deployment` table. Register DeployPlugin success/fail in the table. 

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/722

Satisfies the "[If success, mark plugin as being deployed in plugins table" ](https://github.com/grapl-security/grapl-rfcs/blob/main/text/0000-plugins.md#nomad-job) requirement.

I'm not really that convinced that this is _useful_, since we just track that a deployment was successfully attempted - not that it is healthy - but I can see it being useful as an audit-log.

### DIFF
--- a/src/rust/plugin-registry/migrations/20220202225101_create_plugin_deployment_table.sql
+++ b/src/rust/plugin-registry/migrations/20220202225101_create_plugin_deployment_table.sql
@@ -8,9 +8,11 @@ DO $$ BEGIN
 END $$;
 
 
+-- This table is append-only - so you'll see multiple entries for `plugin_id`
+-- and the latest state will be the last entry.
 CREATE TABLE IF NOT EXISTS plugin_deployment
 (
-    id               SERIAL                    PRIMARY KEY,
+    id               BIGSERIAL                 PRIMARY KEY,
     plugin_id        uuid                      NOT NULL,
     deploy_time      timestamptz               NOT NULL DEFAULT CURRENT_TIMESTAMP,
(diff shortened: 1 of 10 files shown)


src/rust/plugin-registry/migrations/20220202225101_create_plugin_deployment_table.sql
```diff
@@ -0,0 +1,19 @@
+-- Satisfies the "If success, mark plugin as being deployed in plugins table"
+-- requirement of RFC 0000 DeployPlugin.
+
+DO $$ BEGIN
+    IF to_regtype('plugin_deployment_status') IS NULL THEN
+        CREATE TYPE plugin_deployment_status AS ENUM ('fail', 'success');
+    END IF;
+END $$;
+
```
wimax-grapl, line 9: Please do sweat the details on `plugin_deployment`. 
Should I use BIGSERIAL?
Should this also have a tenant-id or is just-plugin-id OK?

jgrillo-grapl (approved):
LGTM

src/rust/plugin-registry/migrations/20220202225101_create_plugin_deployment_table.sql
jgrillo-grapl: plugin-id should be unique enough.. `BIGSERIAL` might not be a bad idea though just to be extra safe? I mean it feels unlikely we'll have 2.1 billion plugin deployments ever but who knows lol

PR #1447, opened 2022-02-03
Split chromeos install script into two scripts

ghost:
This splits the chromeos install script into a required script and an optional script.

We're doing this split to make it clear what are required dependencies
and what are optional helpers. The primary reason for doing this now
is that the current git hook breaks usage of git patches (ie when only
part of a file should be committed instead of the whole thing).
Specifically this breaks CM's workflow


wimax-grapl (approved):

christophermaier (approved):
:bow: THANK YOU! :bow:

PR #1451, opened 2022-02-07
Issue 733 generator executor

colin-grapl:
733

This adds the business logic of the generator executor. The service polls the work queue for new generator work and dispatches that work over to the generator.

All of the mTLS is handled already in the generator client.

The next steps for this service would be to set up the nomad job and get it the cert it needs but in order to keep the PR small I figured I'd start here.

There's no additional testing at this point. As the rest of the plugin system comes together we can start building out more testing.

### DIFF
--- a/src/rust/generator-executor/src/lib.rs
+++ b/src/rust/generator-executor/src/lib.rs
@@ -1,5 +1,3 @@
-
-
 use async_stream::stream;
 use futures_util::pin_mut;
 use generator_sdk::{
@@ -12,16 +10,12 @@ use generator_sdk::{
     ClientConfig as GeneratorClientConfig,
 };
 use grapl_utils::future_ext::GraplFutureExt;
-use plugin_work_queue::client::{
-    PluginWorkQueueServiceClient,
-};
-use rust_proto::{
-    plugin_work_queue::{
-        AcknowledgeGeneratorRequest,
-        ExecutionJob,
-        GetExecuteGeneratorRequest,
-        GetExecuteGeneratorResponse,
-    },
+use plugin_work_queue::client::PluginWorkQueueServiceClient;
+use rust_proto::plugin_work_queue::{
+    AcknowledgeGeneratorRequest,
+    ExecutionJob,
+    GetExecuteGeneratorRequest,
+    GetExecuteGeneratorResponse,
 };
 use structopt::StructOpt;
 use tokio::time::{
--- a/src/rust/generator-executor/src/lib.rs
+++ b/src/rust/generator-executor/src/lib.rs
@@ -1,15 +1,10 @@
-use std::{
-    future::Future,
-    net::SocketAddr,
-    pin::Pin,
-};
+
 
 use async_stream::stream;
 use futures_util::pin_mut;
 use generator_sdk::{
     client::{
         CacheBuilder,
-        ClientCache,
         GeneratorClient,
         GeneratorClientError,
         TokioAsyncResolver,
@@ -19,10 +14,8 @@ use generator_sdk::{
 use grapl_utils::future_ext::GraplFutureExt;
 use plugin_work_queue::client::{
     PluginWorkQueueServiceClient,
-    PluginWorkQueueServiceClientError,
 };
 use rust_proto::{
-    plugin_sdk::generators::RunGeneratorResponse,
     plugin_work_queue::{
         AcknowledgeGeneratorRequest,
         ExecutionJob,
@@ -123,7 +116,7 @@ pub async fn process_message_loop(
 
     while let Some((job, request_id)) = jobs.next().await {
         let ExecutionJob {
-            tenant_id,
+            tenant_id: _,
             plugin_id,
             data,
         } = job;
--- a/src/rust/generator-executor/src/lib.rs
+++ b/src/rust/generator-executor/src/lib.rs
@@ -1,0 +1,180 @@
+use std::{
+    future::Future,
+    net::SocketAddr,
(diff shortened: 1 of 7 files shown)


src/rust/generator-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+fn make_job_stream(
+    mut work_queue_client: PluginWorkQueueServiceClient<tonic::transport::Channel>,
+) -> impl Stream<Item = (ExecutionJob, i64)> {
+    stream! {
+        loop {
+            let message = match work_queue_client.get_execute_generator(GetExecuteGeneratorRequest {}).await {
+                Ok(message) => message,
+                Err(e) => {
+                    tracing::error!(message="Failed to retrieve generator job", error=?e);
+                    // check what kind of error it is and retry it accordingly
```
christophermaier, line 50: Is this actually a `TODO`? If not, how is the checking and retrying actually achieved by this code?

src/rust/generator-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+
+fn make_job_stream(
+    mut work_queue_client: PluginWorkQueueServiceClient<tonic::transport::Channel>,
+) -> impl Stream<Item = (ExecutionJob, i64)> {
+    stream! {
+        loop {
+            let message = match work_queue_client.get_execute_generator(GetExecuteGeneratorRequest {}).await {
+                Ok(message) => message,
+                Err(e) => {
+                    tracing::error!(message="Failed to retrieve generator job", error=?e);
```
christophermaier, line 49: Is it considered an error if the work queue just doesn't have any work to distribute now?

src/rust/generator-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    work_queue_client: &mut PluginWorkQueueServiceClient<tonic::transport::Channel>,
+    request_id: i64,
+    success: bool,
+) {
+    if let Err(e) = work_queue_client
+        .acknowledge_generator(AcknowledgeGeneratorRequest {
+            request_id,
+            success,
+        })
+        .timeout(Duration::from_secs(2))
```
christophermaier, line 78: It would be good to extract these timeout values as constants, at the very least, if not (perhaps later on, and not in this PR) adjustable configuration values or dynamically responsive values.

src/rust/generator-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        }
+    }
+}
+
+// Attempts to ack the message. On failure, will retry in a background task
+async fn try_ack(
+    work_queue_client: &mut PluginWorkQueueServiceClient<tonic::transport::Channel>,
+    request_id: i64,
+    success: bool,
```
christophermaier, line 71: Something I've often found useful from a readability / maintenance standpoint is to replace bare `bool` values with descriptive enums.

`try_ack(client, id, Status::Success)` and `try_ack(client, id, Status::Failure)` read much more easily than the identical code with simply `true` and `false`.

Alternatively (and probably better, in this case, honestly), you could have `ack_success` and `ack_failure` as thin facades over `try_ack` that handle the booleans internally, leaving the broader code easier to read.

(Idiomatically, it also seems odd to call this function `try_XXX` when it doesn't return a `Result`.)

src/rust/generator-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+            data,
+        } = job;
+
+        // based on the error, either retry later or ack as a failure
+        let response = generator_client
+            .run_generator(data, plugin_id.to_string())
+            .await;
+
+        match response {
+            Ok(_) => {
```
christophermaier, line 124: Do we really not care about what's inside the `Ok`?

src/rust/generator-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    config: GeneratorExecutorConfig,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let cert = std::fs::read(config.generator_certificate_path)?;
+    let cert = tonic::transport::Certificate::from_pem(cert);
+
+    let work_queue =
+        PluginWorkQueueServiceClient::connect(config.plugin_work_queue_address).await?;
+    let generator_client = GeneratorClient::new(
+        CacheBuilder::from(config.generator_client_config.client_cache_config).build(),
+        cert, // load certificate
```
christophermaier, line 161: What is the significance of the `// load certificate` comment? Is it a `TODO`? Does passing the value load it?

src/rust/generator-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+) -> Result<(), Box<dyn std::error::Error>> {
+    let jobs = make_job_stream(work_queue_client.clone());
+    pin_mut!(jobs);
+
+    while let Some((job, request_id)) = jobs.next().await {
+        let ExecutionJob {
+            tenant_id: _,
+            plugin_id,
+            data,
+        } = job;
```
christophermaier, line 116: If you don't care about `tenant_id`, I think you can just do:
```rust
let ExecutionJob { plugin_id, data, ... } = job;
```

jgrillo-grapl: If in fact we don't care about the tenant_id it should be removed from the protobuf also. The semantics of an "optional" tenant ID are pretty concerning from a safety standpoint, and `plugin_id` should be "unique enough" I think?

PR #1460, opened 2022-02-10
Upgrade yarn version to v3

ghost:
https://github.com/grapl-security/issue-tracker/issues/844

Upgrades the engagement_view to use yarn v3.

We're upgrading to yarn v3 so that
a. we fix an issue we're seeing in CI.
b. we're on the latest and have a path to enabling zero installs and
other features.
c. Get the yarn versions of local developement and the docker container
to match
Automated test suite, manual testing


andrea-grapl (approved):

src/js/engagement_view/.yarnrc.yml
```diff
@@ -10,4 +8,4 @@ installStatePath: .yarn/state/install-state.gz
 
 nodeLinker: node-modules
 
-yarnPath: .yarn/releases/yarn-berry.cjs
+yarnPath: .yarn/releases/yarn-3.1.1.cjs
```
inickles-grapl, line 11: I think you want to just remove this line as well since it's no longer being sourced from that location.

twunderlich-grapl: So while true that the docker container has the yarn version set correctly, anyone running yarn locally won't necessarily be using the right yarn version. So far, this seems to be the preferred way to set the version although I do agree its stupid to basically have this twice. 
If you know of a config option I've missed to force local development to use the right yarn version please let me know since its been a while since I've done serious yarn spelunking

PR #1463, opened 2022-02-15
Nodejs audit

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/817

Add audit capabilities for js/ts.
This adds two pipelines one which runs npm audit, the other runs yarn audit.

Resurrecting this branch now that compliance is finished.

https://buildkite.com/grapl/grapl-yarn-audit
https://buildkite.com/grapl/grapl-npm-audit

The results should match up roughly with the dependabot alerts: https://github.com/grapl-security/grapl/security/dependabot?page=2&q=is%3Aopen

### DIFF
--- a/.buildkite/pipeline.npm-audit.sh
+++ b/.buildkite/pipeline.npm-audit.sh
@@ -27,7 +27,7 @@ cat << EOF
 steps:
   - label: ":nodejs: npm audit"
     command:
-      - cd src/js/graphql_endpoints
+      - cd src/js/graphql_endpoint
       - npm audit
     plugins:
       - docker#v3.8.0:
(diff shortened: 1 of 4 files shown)


christophermaier (approved):
So, the only reason we're doing Yarn and NPM is because we've got a project managed by each of them, right?

I wish there was a good way to abstract the commonalities between these two and the `cargo audit` job... might not be worth it just yet, but if we add any other jobs like this in the future (perhaps the `pip audit` one?), that would be a good time to try to DRY things up a bit.

Looks good!

twunderlich-grapl:
@christophermaier yep. I think the plan is to convert over to yarn for everything, but that felt like a change that should be done by someone who actually uses the project. Once that's done we can unify all of this.

PR #1466, opened 2022-02-16
some protobuf updates

jgrillo-grapl:

https://github.com/grapl-security/issue-tracker/issues/580

This PR makes some changes to the pipeline protobofus. ~This entails a breaking change to the `Envelope`~ A `NewEnvelope` protobuf was added to a new `v1beta2` package. This is temporary. After `rust-proto` is rewritten the `v1beta1.Envelope` will be removed entirely and replaced with the `v1beta2.NewEnvelope` (which will of course be renamed as well), but right now our rust code is too brittle to survive the change. There are some non-breaking additions to the `Metadata` protobuf. A `RawLogs` message was added. It also adds a dependency on some `google.protobuf` types, which are provided by the protobuf compiler and prost, so they don't need to be vendored.

In addition to the protobufs, this PR updates the `python-proto` interface library:
- made `Envelope[I]` generically parametrized by the type `I` of the inner message
- Updated the base class `SerDe[P]` to be aware of its protocol buffer message type `P`
- Added another base class `SerdeWithInner[P, I]` which is aware of the inner message's type as well.
- Added pipeline type for `RawLog`
- Added/updated hypothesis strategies for generating the new data
- Added tests for everything

I put a lot less effort into the rust code. It needs substantially more work, and that will come in additional PRs over time.

automated tests


src/python/python-proto/python_proto/common.py
```diff
@@ -0,0 +1,122 @@
...
+    def from_datetime(datetime_: datetime.datetime) -> Timestamp:
+        duration = datetime_ - EPOCH
+        return Timestamp(
+            seconds=duration.days * SECONDS_PER_DAY + duration.seconds,
+            nanos=duration.microseconds * 1000,
+        )
+
+    def into_datetime(self) -> datetime.datetime:
+        return EPOCH + datetime.timedelta(
+            seconds=self.seconds, microseconds=self.nanos // 1000
```
wimax-grapl, line 75: we _may_ want to just add a comment about any loss of accuracy (if there is any) - if this is just an implementation detail then nvm

src/python/python-proto/python_proto/pipeline.py
```diff
@@ -1,111 +1,111 @@
...
+        inner_message = _Any()
+        inner_message.Pack(
+            self.inner_message.into_proto(), type_url_prefix="".encode("utf-8")
+        )
+        proto_envelope.inner_message.CopyFrom(inner_message)
         return proto_envelope
+
+
+@dataclasses.dataclass(frozen=True)
+class RawLog(SerDe[_RawLog]):  # FIXME: test
```
wimax-grapl, line 94: is this to be done asap or just soon

src/python/python-proto/python_proto/tests/strategies.py
```diff
@@ -33,16 +36,49 @@
...
+
+INT64_MIN = -(2 ** 63) + 1
 INT64_MAX = 2 ** 63 - 1
+
+INT32_MIN = -(2 ** 31) + 1
+INT32_MAX = 2 ** 31 - 1
+
+# see https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration
+DURATION_SECONDS_MIN = -315_576_000_000
+DURATION_SECONDS_MAX = 315_576_000_000
```
wimax-grapl, line 61: i so rarely see this lol

src/python/python-proto/python_proto/common.py
jgrillo-grapl: yeah that's a good point, I'll add a docstring

src/python/python-proto/python_proto/tests/strategies.py
```diff
@@ -57,23 +93,71 @@ def uuids(
...
 
 
 def envelopes(
     metadatas: st.SearchStrategy[Metadata] = metadatas(),
-    inner_messages: st.SearchStrategy[bytes] = st.binary(),
-    inner_types: st.SearchStrategy[str] = st.text(),
+    inner_messages: st.SearchStrategy[SerDe] = uuids()
+    | timestamps()
+    | durations()
+    | raw_logs(),
```
wimax-grapl, line 155: TIL SearchStrategy supports `__or__`

src/python/python-proto/python_proto/tests/test_common.py
```diff
@@ -0,0 +1,39 @@
+import datetime
+import uuid
+
+import pytest
+
+pytest.register_assert_rewrite("python_proto.tests.helpers")
```
wimax-grapl, line 6: slick i had no idea this exists

wimax-grapl (approved):

src/python/python-proto/python_proto/tests/strategies.py
```diff
@@ -33,16 +36,49 @@
     Label,
     MetricWrapper,
 )
-from python_proto.pipeline import Envelope, Metadata, Uuid
+from python_proto.pipeline import Envelope, Metadata, RawLog
+
+#
+# constants
+#
+# These values are used to parametrize the strategies defined below. Some of
```
colin-grapl, line 44: typo parameterize.

colin-grapl (approved):

src/python/python-proto/python_proto/pipeline.py
jgrillo-grapl: This was actually done, I added strategies and tests for this type. Will remove comment.

src/python/python-proto/python_proto/tests/strategies.py
jgrillo-grapl: Not a typo ;). The following spellings are all equivalent (and accepted):
- parameterize
- parametrize
- parameterise
- parametrise
The latter two are more often used in British English, the former two are used interchangeably (albeit, hopefully, consistently) in American English.

src/python/python-proto/python_proto/tests/test_common.py
jgrillo-grapl: yeah it makes the failures nice to look at

src/python/python-proto/python_proto/tests/strategies.py
jgrillo-grapl: I learned by accident here that this works in python as well, I know rust supports this kind of notation but had no idea python does too.

src/python/python-proto/python_proto/tests/strategies.py
jgrillo-grapl: yeah it's very convenient and expressive

src/python/python-proto/python_proto/common.py
jgrillo-grapl: done

src/python/python-proto/python_proto/pipeline.py
jgrillo-grapl: fixed

src/rust/sqs-executor/src/lib.rs
```diff
@@ -292,10 +293,13 @@ async fn process_message<
 
             let mut envelopes = vec![];
             for event in events {
+                let wrapped_event = prost_types::Any {
```
jgrillo-grapl, line 296: should these changes be here?

src/rust/sysmon/Cargo.toml
```diff
@@ -11,6 +11,9 @@ readme = "README.md"
 
 [dependencies]
 failure = "0.1.8"
+prost = "0.9.0"
+prost-types = "0.9.0"
+rust-proto = { path = "../rust-proto/" }
```
inickles-grapl, line 16: Not used and shouldn't be here.

src/rust/generators/sysmon-generator/src/serialization.rs
```diff
@@ -30,12 +31,19 @@ impl CheckedError for SysmonDecoderError {
     }
 }
 
+impl From<prost::DecodeError> for SysmonDecoderError {
+    fn from(err: prost::DecodeError) -> Self {
+        SysmonDecoderError::DeserializeError(err.to_string())
+    }
+}
+
```
inickles-grapl, line 39: The generator shouldn't care about grapl service proto, it should only need to work with log data, but even still, this should be its own variant of `SysmonDecoderError` that uses thiserror's `#[from] prost::DecodeError`, instead of packing it into the `DeserializeError` variant.

src/rust/generators/sysmon-generator/src/serialization.rs
```diff
@@ -1,6 +1,7 @@
 use std::str::FromStr;
 
 use grapl_service::decoder::decompress::PayloadDecompressionError;
+use prost;
```
inickles-grapl, line 4: Can also be removed.

src/rust/generators/sysmon-generator/Cargo.toml
```diff
@@ -15,6 +15,8 @@ name = "generator_bench"
 harness = false
 
 [dependencies]
+prost = "0.9.0"
+prost-types = "0.9.0"
```
inickles-grapl, line 19: Can also be removed.

inickles-grapl (changes requested):
Only looked at the Rust bits. Looks like a few dangling imports that are no longer used, for the most part.

src/rust/generators/sysmon-generator/src/serialization.rs
jgrillo-grapl: Ah, nvm just saw your comment above. Will fix.

src/rust/generators/sysmon-generator/src/serialization.rs
jgrillo-grapl: Removed

src/rust/sysmon/Cargo.toml
jgrillo-grapl: Fixed

src/rust/generators/sysmon-generator/Cargo.toml
jgrillo-grapl: Fixed

PR #1467, opened 2022-02-16
Install Docker Compose v2 CLI plugin

christophermaier:
This ensures that both the old and new versions of Docker Compose are
present on our workstations.

We still use the old Docker Compose (invoked as `docker-compose`) in
our Makefiles, scripts, etc., but this will enable us to start
migrating over to the new Docker Compose (invoked as `docker compose`;
note the space).


jgrillo-grapl (approved):
breaking changes are fun 👏

PR #1469, opened 2022-02-16
Pulumi firecracker kernel

wimax-grapl:

https://github.com/grapl-security/issue-tracker/issues/742 kinda

Now that we can generate a Firecracker kernel (or reference it from Cloudsmith), we want to upload that
asset to S3 with Pulumi. This does that.

The next step is to hook it into the Plugin Registry, but that's another big refactor :)

did an e2e local and inspected the env variable passed to plugin-registry


pulumi/infra/artifacts.py
```diff
@@ -0,0 +1,59 @@
...
+        If no such artifact exists, and require_artifact, throw error
+
+        We generally set `require_artifact=True` for production deployments.
+        """
+        artifact_version = self.artifacts.get(key)
+        if artifact_version is not None:
+            return artifact_version
+        if self.require_artifact:
+            raise KeyError(
+                f"Expected to find an artifacts entry for {key} in Pulumi.stackname.yaml"
```
christophermaier, line 37: It would be good to interpolate the actual stack name in `Pulumi.stackname.yaml`.

pulumi/infra/config.py
```diff
@@ -213,18 +213,26 @@ def require_artifact(artifact_name: str) -> Any:
...
-    return value of this function.
+    The value will be something like `grapl/testing`.
+    """
+    return pulumi.Config().get("cloudsmith-repository-name")
+
+
+def container_repository() -> Optional[str]:
+    """The repository from which to pull container images from.
 
     Not specifying a repository will result in local images being used.
```
christophermaier, line 232: "... but only for 'local grapl' deployments"

pulumi/infra/firecracker_assets.py
```diff
@@ -0,0 +1,100 @@
...
+        super().__init__("grapl:FirecrackerS3BucketObjects", name, None, opts)
+        child_opts = pulumi.ResourceOptions(parent=self)
+
+        kernel_s3obj = aws.s3.BucketObject(
+            "firecracker_kernel",
+            key=FIRECRACKER_KERNEL_FILENAME,
+            bucket=plugins_bucket.bucket,
+            source=firecracker_assets.kernel_asset,
+            opts=child_opts,
+        )
```
christophermaier, line 55: I think this is going to effectively mean that we'll only ever have a single kernel image in use at a time, which may not be a good idea.

Also, since the default is to create a new resource and then delete the old resource, this will also have the effect of deleting the kernel image from the bucket entirely once we get a second kernel version in play (_"Ask me how I know", he said bitterly_). At the very least, we'll need to have a `delete_before_replace=True` resource option.

However, this this is also why I think we'll want to be able to have multiple kernel images in the bucket at a time; imagine a Nomad job trying to spin up right between the deletion of the old image and the uploading of the new image. It won't find any image, and wackiness will ensue.

pulumi/infra/firecracker_assets.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def cloudsmith_cdn_url(
+    repository_name: str,
+    package_name: str,
+    version: str,
+) -> str:
+    return (
+        f"https://dl.cloudsmith.io/public/{repository_name}"
+        f"/raw/versions/{version}/{package_name}"
```
christophermaier, line 72: Might want to emphasize that "raw" here is referring to a Cloudsmith _package type_ (i.e., not an RPM, not a container, but a generic file), and not to our "grapl/raw" Cloudsmith _repository_.

Also, why two lines? Is this some kind of automatic formatting?

pulumi/infra/firecracker_assets.py
```diff
@@ -0,0 +1,100 @@
...
+            bucket=plugins_bucket.bucket,
+            source=firecracker_assets.kernel_asset,
+            opts=child_opts,
+        )
+        self.kernel_s3obj_url = get_s3url(kernel_s3obj)
+
+
+def get_s3url(obj: aws.s3.BucketObject) -> pulumi.Output[str]:
+    return pulumi.Output.all(bucket=obj.bucket, key=obj.key).apply(
+        lambda inputs: f"https://{inputs['bucket']}.s3.amazonaws.com/{inputs['key']}"
```
christophermaier, line 61: How does `BucketObject` not have a native way to return its own URL?!

christophermaier (changes requested):

pulumi/infra/artifacts.py
wimax-grapl: Done. Introduced a STACK_CONFIG_FILENAME and use it in error msgs

pulumi/infra/config.py
wimax-grapl: fixed

pulumi/infra/firecracker_assets.py
wimax-grapl: i try and keep stuff under 80 chars - not religiously but as a general rule

pulumi/infra/firecracker_assets.py
wimax-grapl: yeah i know lol

pulumi/infra/firecracker_assets.py
wimax-grapl: > since the default is to create a new resource and then delete the old resource
I'm not sure how that works with S3BucketObject here, since - _theoretically_ - an update would just be an S3 PUT request that overwrites the object already there, no?

Some other thoughts:
- Do you mean `delete_before_replace=False`? Because what we want to avoid is having _nothing_ at that key, right?
- It looks like Nomad doesn't cache its artifacts, what the heck?! we gotta download another 20mb from s3 every time we run a plugin? https://github.com/hashicorp/nomad/issues/5217

> this is also why I think we'll want to be able to have multiple kernel images in the bucket at a time
- You've made a good case for why we'd want it
- I don't know how to do this today with our existing artifacts setup
- I'm open to ideas
- I'd ask if this is something that needs to be solved ASAP, or could be kicked down the road

christophermaier: I mean `delete_before_replace=True`, because the artifacts are going to be _at the same location_ in the bucket.

If we have `delete_before_replace=False`, then this is what will happen:
- We upload the new kernel to `<BUCKET>/firecracker_kernel` (the "replace" step)
- Now that the new resource is in place, we can delete the old one; now we delete `<BUCKET>/firecracker_kernel`
- Now we have no kernel

If we have `delete_before_replace=True`, this is what happens:
- Delete the old resource `<BUCKET>/firecracker_kernel`
- Upload the new resource to `<BUCKET>/firecracker_kernel`
- Now we have a new kernel.

The latter is preferable to the former, but it's still not perfect because there is still a time where there is no kernel available; anything that spins up in those few seconds is going to have a bad time.

For now, I think this is fine (let's get to the point where we're actually running things in Firecracker first, you know 😅), but we will definitely need to address this before production. An issue to track this would be great.

(Incidentally, I don't think it should be too terribly hard, or significantly different than what we already do with our Cloudsmith artifacts... we just incorporate the version into the key, so rather than downloading from `<BUCKET>/firecracker_kernel`, we go to `<BUCKET>/firecracker_kernel/kernel_x.y.z`... probably good to also include some metadata files alongside that -- checksums, etc. But yes, all this could come after this PR. We used to use S3 as a glorified artifact store like this at my last job.)

(Is the reason we're putting these in S3 rather than Cloudsmith data transfer fees? Or do we know enough about that to make a call at this point?)

wimax-grapl: Okay, that makes sense.

S3-vs-Cloudsmith-wise: I mostly went this route because the plugin.nomad file provided in the spec said
```

      artifact {
        source      = var.kernel_artifact_url
        destination = "local/vmlinux"
        headers {
          x-amz-expected-bucket-owner = var.aws_account_id
          x-amz-meta-client-id        = "nomad-deployer"
        }
      }
```
and I was blindly following it instead of questioning 😇 
but this does suggest that we want to have some acl checks on it

christophermaier: 🤔 If that's just a URL with HTTP headers, we should be able to use Cloudsmith just as easily.

(Just food for thought, and possibly for the follow-up issue.)

pulumi/infra/config.py
```diff
@@ -9,14 +9,18 @@
 import pulumi
 
 # This will be incorporated into various infrastructure object names.
-DEPLOYMENT_NAME = pulumi.get_stack()
+DEPLOYMENT_NAME: Final[str] = pulumi.get_stack()
```
christophermaier, line 12: Wonder if it's time to rename this to `STACK_NAME` instead of `DEPLOYMENT_NAME` to finally sever all ties with our CDK past?

(Doesn't need to be in this PR; just thinking aloud.)

christophermaier (approved):

pulumi/infra/config.py
wimax-grapl: image

PR #1470, opened 2022-02-17
remove redundant .clone() in Rust Files from deleting --allow clippy:…

andrea-grapl:
…:redundant_clone from src/rust/bin/lint

https://github.com/grapl-security/issue-tracker/issues/853

This specific issue is pertaining to errors that manifest as a result of removing --allow clippy::redundant_clone from src/rust/bin/lint. This effort was straightforward, consisting of removing redundant `.clone()` when running `make lint-rust`. 

`make up` && `make test-e2e`

These are not functional changes and should not effect behavior.


christophermaier (approved):
:tada: Thanks for removing this!

PR #1484, opened 2022-02-18
Also simplify the aws-local-only-env-vars

wimax-grapl:
Some simplifications:
- Unify the 3 environment variables as one; previously we had this weird behavior of "uh yeah if 1 of them exists we use all 3"
- As such, I moved the template string up to the Pulumi level
- Export the fully templated variable so other stacks can use it (i.e. pulumi/integration_tests)
- Remove any defaults from Nomad, and pull them up to Pulumi (this makes exporting it easier)

### DIFF
--- a/nomad/e2e-tests.nomad
+++ b/nomad/e2e-tests.nomad
@@ -44,22 +44,17 @@ variable "schema_properties_table_name" {
   description = "What is the name of the schema properties table?"
 }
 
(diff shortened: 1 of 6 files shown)


christophermaier (approved):
Exporting everything as a single string from Pulumi looks kinda weird, but this seems like a definite improvement over the status quo 👍

PR #1488, opened 2022-02-22
"Deployment name" becoming "stack name"

wimax-grapl:
i replaced like 90% of the occurences, and it's completely gone at the pulumi/nomad level. just a few function param names


pulumi/integration_tests/__main__.py
```diff
@@ -53,14 +53,14 @@ def _integration_container_images(
...
-    stack_name = pulumi.get_stack()
+    stack_name = config.STACK_NAME
 
     pulumi_config = pulumi.Config()
     artifacts = ArtifactGetter.from_config(pulumi_config)
 
     # These tags will be added to all provisioned infrastructure
     # objects.
-    register_auto_tags({"grapl deployment": stack_name})
+    register_auto_tags({"grapl_stack_name": stack_name})
```
christophermaier, line 63: If we want our tags to be consistent with all our our other deployments, this should be:
```python
register_auto_tags(
    {"pulumi:project": pulumi.get_project(), "pulumi:stack": pulumi.get_stack()}
)
```

pulumi/grapl/__main__.py
```diff
@@ -320,25 +320,28 @@ def main() -> None:
...
-                "stack_name",
-                "py_log_level",
-                "schema_properties_table_name",
-                "schema_table_name",
-                "test_user_name",
-                "user_auth_table",
-            },
-        )
+        local_provision_vars: Final[NomadVars] = {
+            "stack_name": config.STACK_NAME,
```
christophermaier, line 323: I thought we removed `stack_name` in this commit?

pulumi/grapl/__main__.py
```diff
@@ -191,6 +205,7 @@ def main() -> None:
         analyzer_matched_subgraphs_bucket=analyzer_matched_emitter.bucket_name,
         analyzer_dispatcher_dead_letter_queue=analyzer_dispatcher_queue.dead_letter_queue_url,
         aws_region=aws.get_region().name,
+        container_images=_container_images(artifacts),
```
christophermaier, line 208: Why is this change in this PR?

pulumi/grapl/__main__.py
```diff
@@ -182,6 +182,20 @@ def main() -> None:
...
+            "DEBUG",
+            "h2::codec=WARN",
+            "hyper=WARN",
+            "rusoto_core=WARN",
+            "rustls=WARN",
+            "serde_xml_rs=WARN",
+        ]
+    )
+    py_log_level = "DEBUG"
+
```
christophermaier, line 198: Why is this change in this PR?

local-grapl.env
```diff
@@ -76,8 +76,4 @@ GRAPL_LOG_LEVEL=DEBUG
 # logs. This turns that buffering off.
 PYTHONUNBUFFERED=1
 RUST_BACKTRACE=1
-RUST_LOG=DEBUG
-
-# Miscellaneous
-########################################################################
-MG_ALPHAS=${DGRAPH_HOST}:${DGRAPH_ALPHA_GRPC_EXTERNAL_PUBLIC_PORT}
+RUST_LOG=DEBUG
```
christophermaier, line 79: Why is this change in this PR?

christophermaier (commented):
Rebasing on `main` would get rid of the `codecov/patch` noise.

pulumi/grapl/__main__.py
wimax-grapl: Hah - don't read in to my commit messages that much, this will be a squash - in this case, I did a :%s/DEPLOMENT_NAME/STACK_NAME/g but forgot to save the file before committing

pulumi/grapl/__main__.py
wimax-grapl: I mean, it's moving a preexisting thing around by 40 lines - but I can revert since it doesn't match the PR title

wimax-grapl:
gonna not do the codecov/patch rebase because a force push will screw up all your commit-specific messages (or at least I think so?)

pulumi/grapl/__main__.py
christophermaier: I get that... I was just wondering why rearranging of logging code got pulled into variable renaming.

christophermaier:
> gonna not do the codecov/patch rebase because a force push will screw up all your commit-specific messages (or at least I think so?)

🤔 I think it might, but Github is always improving this kind of stuff behind the scenes... maybe try a rebase/force-push and see what happens? For science? 😅

christophermaier (approved):

PR #1493, opened 2022-02-24
Assorted Makefile cleanups and refactorings

christophermaier:
In the course of reworking some Docker-related stuff, it became apparent that our existing Makefile targets were a bit more complicated than necessary. They have now been rearranged to better reflect the true underlying dependency structure.

I also added some additional logging output for the benefit of our Buildkite runs of our integration and e2e tests, as well as enabled our remote build cache for those tests, which should help speed things up a bit.

Read the individual commits for further details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,10 @@ build-test-integration: build
 
 .PHONY: build-test-e2e
 build-test-e2e: build
-	./pants package ./src/python/e2e-test-runner/e2e_test_runner:pex
+# Any PEX tagged with `e2e-test-pex` is required for our image. This
+# seems like the most straightforward way of capturing these
+# dependencies at the moment.
+	./pants --tag="e2e-test-pex" package ::
 	$(DOCKER_BUILDX_BAKE) \
 		--file ./test/docker-compose.integration-tests.build.yml \
 		e2e-tests
--- a/Makefile
+++ b/Makefile
@@ -127,10 +127,6 @@ help: ## Print this help
 
 ##@ Build 🔨
 
-.PHONY: build-service-pexs
-build-service-pexs: ## Build all PEX files that are used by our service processes
-	./pants --tag="service-pex" package ::
-
 .PHONY: build-test-unit
 build-test-unit:
 	$(DOCKER_BUILDX_BAKE) \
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -148,25 +144,72 @@ build-test-unit-js:
...
 
 .PHONY: build-test-e2e
-build-test-e2e: build
-	./pants package ./src/python/e2e-test-runner/e2e_test_runner:pex
+build-test-e2e:
+	@echo "--- Building e2e testing image"
+# Any PEX tagged with `e2e-test-pex` is required for our image. This
+# seems like the most straightforward way of capturing these
+# dependencies at the moment.
+	./pants --tag="e2e-test-pex" package ::
```
wimax-grapl, line 160: so today i learned how to comment makefiels correctly without it showing up at runtime 😢

Makefile
```diff
@@ -358,10 +386,15 @@ format: format-python format-shell format-prettier format-rust format-hcl format
 ##@ Local Grapl 💻
 
 .PHONY: up
-up: build ## Bring up local Grapl and detach to return control to tty
+up: ## Bring up local Grapl and detach to return control to tty
+up: build-grapl-services build-local-services _up
+
+# NOTE: Internal target to decouple the building of images from the
```
jgrillo-grapl, line 398: awesome

jgrillo-grapl (approved):

Makefile
```diff
@@ -354,10 +392,16 @@ format: format-python format-shell format-prettier format-rust format-hcl format
...
+# running of them. Do not invoke this directly.
+.PHONY: _up
+_up:
```
wimax-grapl, line 401: maybe `_up_no_build`?

Makefile
```diff
@@ -148,25 +144,72 @@ build-test-unit-js:
...
+.PHONY: build-ui
+build-ui: ## Build the web UI
+build-ui: build-engagement-view
+	@echo "--- Building the Web UI"
+	$(DOCKER_BUILDX_BAKE) --file docker-compose.build.yml \
+		grapl-web-ui
+
+.PHONY: build-grapl-services
+build-grapl-services: ## Build all the services for the Grapl SaaS
+build-grapl-services: build-ui
```
wimax-grapl, line 182: I guess I'm not fully sure what we gain from indicating all of these here, versus splitting up `docker-compose.build.yml` into local and nonlocal.

We have enough issues remembering to add new services to build_and_upload_containers.sh !

wimax-grapl (approved):

Makefile
christophermaier: That's all going to be addressed in the PR I'm working on now, using the HCL file for `buildx bake`. This was a precursor to that, when I was trying to sort out the best way to build things in `build_and_upload_containers.sh`.

PR #1494, opened 2022-02-24
Less stackname interpolation

wimax-grapl:
Eliminate some of the last places that consume STACK_NAME. Instead, pass in the TestUserPassword secretid verbatim.

https://github.com/grapl-security/issue-tracker/issues/865

### DIFF
--- a/pulumi/integration_tests/__main__.py
+++ b/pulumi/integration_tests/__main__.py
@@ -77,7 +77,7 @@ def main() -> None:
         "aws_region": aws.get_region().name,
         "container_images": _e2e_container_images(artifacts),
         # Used by graplctl to determine if it should manual-event or not
-        "stack_name": grapl_stack.upstream_stack_name,  
+        "stack_name": grapl_stack.upstream_stack_name,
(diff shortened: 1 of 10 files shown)


christophermaier (approved):

pulumi/integration_tests/__main__.py
```diff
@@ -76,7 +76,8 @@ def main() -> None:
         "aws_env_vars_for_local": grapl_stack.aws_env_vars_for_local,
         "aws_region": aws.get_region().name,
         "container_images": _e2e_container_images(artifacts),
-        "stack_name": grapl_stack.upstream_stack_name,
+        # Used by graplctl to determine if it should manual-event or not
```
jgrillo-grapl, line 79: what does "manual-event" mean?

jgrillo-grapl (approved):

pulumi/integration_tests/__main__.py
wimax-grapl: image

basically it's the whole "when we put something in S3, send a message to the associated SQS queue"
that's automatic in AWS

PR #1496, opened 2022-02-24
Switch to HCL file for driving image builds

christophermaier:
Buildkit-based builds can be driven by Docker Compose YAML files (as
we have been doing for a while now), or by a HCL file, which has
several benefits.

In no particular order:

- We can define "targets" for each image we want to build, and each
target is _only_ for building an image, not building and maybe
running, as with Docker Compose services.

- Targets can inherit configuration from other targets (similar to
Docker Compose's old (and recently re-introduced) `extends` syntax.

- You can formally define variables in the HCL file.

- You can actually use _functions_ in the file, which Docker Compose
YAML files have never had.

- You can define your own functions, something also missing from
Docker Compose YAML.

- Targets can be combined into groups. Groups can also be combined,
allowing you to create arbitrary collections of images for whatever
purpose you need. This has never been part of Docker Compose (though
its `profiles` feature is in the vicinity of it).

- It's introspectable; by passing `--print` to your `docker buildx
bake` command, you get a JSON representation of what your invocation
would actually build, with all variables, targets, and groups fully
expanded.

- You can define multiple tags for an image, rather than relying on
the single tag that Docker Compose allows you.

- Configuration options are selectively overridable at build-time via
the CLI.

- It's HCL and not YAML :)

We can use all these features to accomplish quite a few neat tricks.

- We define some "base targets" (one for Rust images and one for
Python images). These ensure that all services built from our Rust or
Python Dockerfiles are using the correct context and Dockerfile, and
are passed the proper value of the `RUST_BUILD` build argument (for
Rust images, naturally).

- We create several groups of images; one for all Rust services, one
for all Python services, one for all integration test images, one for
all images that are only used for "local Grapl" runs, one for all the
images that we'll ultimately push to Cloudsmith, and so on. You can
then operate on these groups as logical entities (need to build all
the images to run Grapl? Just run `docker buildx bake grapl-services`;
no need to manually curate a list to pass the command each time!)

- We use functions to know whether or not we're doing a "release
build" (i.e., something we'll be pushing to Cloudsmith for real
eventual production usage) or just a "local build". If the value of
the `VERSION` variable is unset (""), "latest", or "dev", then we're
not doing a release build. This means that our Rust builds will be run
with a `debug` profile, rather than a `release` profile, none of our
image tags will start with `docker.cloudsmith.io/grapl/raw`. If they
don't start with that, there's no way we can push them to
Cloudsmith. On the other hand, if we pass something else as
`VERSION` (like `20220223182016-70b8e4f`), we'll build Rust services
with the `release` profile (yielding binaries that are ~10% their
`debug`-built sizes), and we'll add the correct image tags. All this
is expressed in a handful of related variables and simple
functions (standard and custom-built).

- We also use custom functions to formally denote which images are
destined for Cloudsmith and which are for local usage only. Images for
local usage have their tags defined via a `local_only_tag` function,
which _never_ includes `docker.cloudsmith.io/grapl/raw`, regardless of
the value of `VERSION` (or anything else). Images that can be pushed
to Cloudsmith use the `upstream_aware_tag` custom function.

All this allows us to consolidate build instructions for a number of
containers into a single place, and treat them as a coherent
collection of artifacts, regardless of which Dockerfile they come
from.

(There are a handful of related refactorings that go along with this; as always, read individual commits for further details)


Makefile
```diff
@@ -5,7 +5,14 @@
...
 
 -include .env
-TAG ?= dev
+
+# This variable is used in a few places, most notably
+# docker-bake.hcl. You can read more about it there, but the TL;DR is
+# that you'll need to set this to a proper version (not "", "latest",
+# or "dev") in order to generate release builds of our services to use
+# in a production deployment. Defaulting to `dev` is fine for
+# day-to-day development and local testing, though.
```
twunderlich-grapl, line 14: Do you want to add a note that the reason why we default to dev locally is because nomad doesn't pick up the latest tag?

Makefile
```diff
@@ -143,73 +157,70 @@ build-test-unit-js:
...
 
-.PHONY: build-for-push
-build-for-push: ## Build only the images we push to our registry
-build-for-push: build-grapl-services build-test-e2e
+.PHONY: build-test-integration
+build-test-integration:
+	@echo "--- Building integration test images"
+	$(DOCKER_BUILDX_BAKE) --file "${BUILDX_BAKE_HCL_FILE}" integration-tests
+
+.PHONY: build-local-infrastructure
```
twunderlich-grapl, line 217: Can you move build-local-infrastructure above the build-test-* targets that way its easier to find the command?

docker-bake.hcl
```diff
@@ -0,0 +1,434 @@
...
+#   that a .dockerignore file exists in each location
+
+# Variables
+########################################################################
+# Variables in this section are intended to be set by users to
+# influence the build.
+
+# IMAGE_TAG is the "master variable"; if it is set to something other
+# than "latest", we will build images suitable for production
+# usage. If it is unset, or "latest", we'll be creating images for
```
twunderlich-grapl, line 20: Can you update this with to mention that dev will create images for local production. ie  `if it is unset, "dev", or "latest"...`
You have dev mentioned in the comments below, but since this is the variable that users should be updating they're more likely to read this comment

docker-bake.hcl
```diff
@@ -0,0 +1,434 @@
...
+# what you want.
+group "default" {
+  targets = [
+    "all"
+  ]
+}
+
+# The services that will ultimately be deployed for Grapl in AWS
+group "grapl-services" {
+  # NOTE: Please keep this list sorted in alphabetical order
```
twunderlich-grapl, line 123: 🥳

twunderlich-grapl (changes requested):
Some relatively minor comments, otherwise looks good.

christophermaier:
@twunderlich-grapl I think I've addressed your requests; how do things look now?

wimax-grapl (approved):

twunderlich-grapl (approved):
@christophermaier  thanks for doing that, looks good!

PR #1498, opened 2022-02-25
Add a Dockerized way to build a rootfs. Currently, the rootfs just contains a dummy file.

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/742

Provide a mechanism to generate a `dist/firecracker_rootfs.tar.gz`

Ran `make dist/firecracker_rootfs.tar.gz` in a few different scenarios (files changed, or dist file removed).

I've mounted the resulting file (ungzipped) and it contains the hello.txt as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -507,3 +507,7 @@ generate-sqlx-data:  # Regenerate sqlx-data.json based on queries made in Rust c
 # Intentionally not phony, as this generates a real file
 dist/firecracker_kernel.tar.gz: firecracker/generate_firecracker_kernel.sh
 	./firecracker/generate_firecracker_kernel.sh
+
+FIRECRACKER_ROOTFS_DEPENDENCIES = $(shell find firecracker/rootfs)
+dist/firecracker_rootfs.tar.gz: $(FIRECRACKER_ROOTFS_DEPENDENCIES)
+	./firecracker/rootfs/generate_firecracker_rootfs.sh
--- a/firecracker/rootfs/bake.hcl
+++ b/firecracker/rootfs/bake.hcl
@@ -1,0 +1,7 @@
+target "rootfs-build" {
+  context    = "."
+  dockerfile = "Dockerfile.rootfs"
+  tags = [
+    "rootfs-build:dev"
+  ]
+}
--- a/firecracker/rootfs/docker_runtime/create_rootfs_image.sh
+++ b/firecracker/rootfs/docker_runtime/create_rootfs_image.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+set -euo pipefail
+set -o xtrace
+
+readonly IMAGE="/tmp/rootfs.ext4"
+readonly MOUNT_POINT="/tmp/rootfs-mount"
+readonly DIST_FILE="/dist/firecracker_rootfs.tar.gz"
+
+########################################
+# Create image and mount it.
+########################################
+# make a 50mb empty file
+dd if=/dev/zero of="${IMAGE}" bs=1M count=50
+
+# format that filesystem
+/sbin/mkfs.ext4 "${IMAGE}"
+
+# make a mount-point
+mkdir "${MOUNT_POINT}"
+
+# Mount
+mount -t fuse-ext2 -o rw+ "${IMAGE}" "${MOUNT_POINT}"
+
+########################################
+# Mutate image
+########################################
+(
+    cd "${MOUNT_POINT}"
+    echo "hello :)" > hello.txt
+)
+
+########################################
+# Copy rootfs into dist.
+########################################
+# -u is unmount; there is no -- version
+fusermount -u "${MOUNT_POINT}"
+tar --create --gzip --file="${DIST_FILE}" "${IMAGE}"
--- a/firecracker/rootfs/generate_firecracker_rootfs.sh
+++ b/firecracker/rootfs/generate_firecracker_rootfs.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+THIS_DIR=$(dirname "${BASH_SOURCE[0]}")
+readonly THIS_DIR
+
+REPOSITORY_ROOT=$(git rev-parse --show-toplevel)
+readonly REPOSITORY_ROOT
+
+(
+    cd "${THIS_DIR}"
+    docker buildx bake -f bake.hcl rootfs-build
+)
+
+########################################
+# Generate the rootfs.
+# We use Fuse to mount inside Docker.
+########################################
+docker run \
+    --rm \
+    --device /dev/fuse \
+    --cap-add SYS_ADMIN \
+    --volume "${REPOSITORY_ROOT}/dist:/dist" \
+    rootfs-build:dev
--- a/firecracker/rootfs/Dockerfile.rootfs
+++ b/firecracker/rootfs/Dockerfile.rootfs
@@ -1,0 +1,14 @@
+# syntax=docker/dockerfile:1.3-labs
+
+FROM debian:bullseye-slim AS rootfs-build
+
+RUN --mount=type=cache,target=/var/lib/apt/lists,sharing=locked,id=rootfs-apt <<EOF
+    apt update
+    apt install --yes \
+        fuse\
+        fuseext2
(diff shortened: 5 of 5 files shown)


firecracker/rootfs/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR /home/grapl
+
+RUN --mount=type=cache,target=/var/lib/apt/lists,sharing=locked,id=rootfs-apt <<EOF
+    apt-get update
+    apt-get install --yes --no-install-recommends \
+        fuse=2.9.9-5 \
+        fuseext2=0.4-1.2+b1
+EOF
+
+COPY docker_runtime docker_runtime
```
twunderlich-grapl, line 14: Am I missing this file/directory? Does this need to be committed? Also I don't see this being used

wimax-grapl: It's there:
https://github.com/grapl-security/grapl/pull/1498/files#diff-5ac450fc9b2ea12729c5e5f795d6bbce5ea457229e6e24ce3f129f741fa5c7e2

and it's where we write the script that's actually done at `docker run` time: create_rootfs_image

Makefile
```diff
@@ -507,3 +507,7 @@ generate-sqlx-data:  # Regenerate sqlx-data.json based on queries made in Rust c
 # Intentionally not phony, as this generates a real file
 dist/firecracker_kernel.tar.gz: firecracker/generate_firecracker_kernel.sh
 	./firecracker/generate_firecracker_kernel.sh
+
+FIRECRACKER_ROOTFS_DEPENDENCIES = $(shell find firecracker/rootfs)
```
twunderlich-grapl, line 511: 1. This should be := not =
2. Why do we need this find anyway?

firecracker/rootfs/bake.hcl
```diff
@@ -0,0 +1,7 @@
+target "rootfs-build" {
```
christophermaier, line 1: This file is canonically called `docker-bake.hcl`.

firecracker/rootfs/generate_firecracker_rootfs.sh
```diff
@@ -0,0 +1,24 @@
...
+########################################
+# Generate the rootfs.
+# We use Fuse to mount inside Docker.
+########################################
+docker run \
+    --rm \
+    --device /dev/fuse \
+    --cap-add SYS_ADMIN \
+    --volume "${REPOSITORY_ROOT}/dist:/dist" \
+    rootfs-build:dev
```
christophermaier, line 24: I'm curious why you need to build an image and then run it to make the rootfs tarball... could this all be done in the image build itself and then copied out using the `output =["type=local,dest=."]` trick in the bake HCL file?

Also curious why you need `fuse`.

Makefile
wimax-grapl: re: := / =, I was following the pattern of `ASSET_DIRS / ASSET_FILES` in the other makefile

The `find` is a way to list the directory and every file underneath it; we do this so that we don't needlessly rebuild the .tar.gz unless something in that directory has changed.

There's no great way to have make targets depend on a directory: https://stackoverflow.com/questions/4440500/depending-on-directories-in-make

firecracker/rootfs/generate_firecracker_rootfs.sh
wimax-grapl: # Why fuse
We need `fuse` because mounting loop devices isn't really supported on crostini:
https://grapl-internal.slack.com/archives/C0174A8QV2S/p1645807029980149?thread_ts=1645720167.748819&cid=C0174A8QV2S

I believe "an image file" qualifies as a loop device.

# Why not at `bake` time?
Since we need fuse, we need `--device /dev/fuse` to use it from within Docker.
Since we need `--device /dev/fuse` - and `--cap-add SYS_ADMIN` - I presumably can't do that with `buildx bake`. I can't prove that they're not supp

christophermaier: It looks like you're wanting to create a directory structure and then package that up as a file; I'm just wondering if you can do that by literally creating the directory structure you want and then feeding that into a process that creates the desired file format. If so, that seems like it might be simpler for this usecase.

If not, that's fine!

twunderlich-grapl: @wimax-grapl I think it would be useful to take the Why fuse section here and put it in either a comment or in a doc.

wimax-grapl: I haven't been able to find such a process-that-creates-ext4; I was mostly going off of [this firecracker guide](https://github.com/firecracker-microvm/firecracker/blob/main/docs/rootfs-and-kernel-setup.md). 

I believe the "format," so to speak, is a "KVM image"

colin-grapl: https://unix.stackexchange.com/questions/503211/how-can-an-image-file-be-created-for-a-directory

Looks like what Max is doing is probably the best bet

firecracker/rootfs/docker_runtime/create_rootfs_image.sh
```diff
@@ -0,0 +1,40 @@
...
+dd if=/dev/zero of="${IMAGE}" bs=1M count=50
+
+# format that filesystem
+/sbin/mkfs.ext4 "${IMAGE}"
+
+# make a mount-point
+mkdir "${MOUNT_POINT}"
+
+# Mount
+mount -t fuse-ext2 -o rw+ "${IMAGE}" "${MOUNT_POINT}"
```
colin-grapl, line 22: ext2?

firecracker/rootfs/docker_runtime/create_rootfs_image.sh
```diff
@@ -32,6 +32,9 @@ mount -t fuse-ext2 -o rw+ "${IMAGE}" "${MOUNT_POINT}"
 ########################################
 # Copy rootfs into dist.
 ########################################
-tar --create --gzip --file="${DIST_FILE}" "${IMAGE}"
+(
+    cd /tmp
+    tar --create --gzip --file "${DIST_FILE}" --directory "/tmp" "rootfs.ext4"
```
christophermaier, line 37: I think the `cd /tmp` is redundant with the `--directory "/tmp"` option on `tar`.

firecracker/rootfs/docker_runtime/create_rootfs_image.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Mount
+mount -t fuse-ext2 -o rw+ "${IMAGE}" "${MOUNT_POINT}"
+
+########################################
+# Mutate image
+########################################
+(
+    cd "${MOUNT_POINT}"
+    echo "hello :)" > hello.txt
```
christophermaier, line 29: Is it possible (at this stage of all this Firecracker work) to have a test where we can spin this up under Firecracker and make assertions about the content of the filesystem?

firecracker/rootfs/docker_runtime/create_rootfs_image.sh
wimax-grapl: yeah believe it or not 
image

firecracker/rootfs/docker_runtime/create_rootfs_image.sh
wimax-grapl: believe it or not the `--directory` doesn't seem to do it itself

firecracker/rootfs/docker_runtime/create_rootfs_image.sh
wimax-grapl: I think an "unsquash/mount it and run goss tests" is absolutely worth doing _but_ not as part of this PR

firecracker/rootfs/docker_runtime/create_rootfs_image.sh
christophermaier: Oh, try moving `--directory` further over to the left (before `--file`, at least):
> -C, --directory=DIR
              Change to DIR before performing any operations.  This option is order-sensitive, i.e. it affects all options that follow.

PR #1501, opened 2022-03-01
Add 'sysmon_parser' to replace the 'sysmon'.

inickles-grapl:


.github/CODEOWNERS
```diff
@@ -81,7 +81,7 @@ BUILD @christophermaier
 /src/rust/rust-proto/ @grapl-security/wg-data-infra
 /src/rust/rust-toolchain.toml @inickles-grapl
 /src/rust/sqs-executor/ @inickles-grapl @grapl-security/wg-data-infra
-/src/rust/sysmon/ @inickles-grapl
+/src/rust/sysmon_parser/ @inickles-grapl
```
colin-grapl, line 84: sysmon-parser pls

src/rust/generators/sysmon-generator/src/serialization.rs
```diff
@@ -26,70 +26,66 @@ impl CheckedError for SysmonDecoderError {
...
+    fn decode(&mut self, body: Vec<u8>) -> Result<Vec<SysmonEvent<'static>>, Self::DecoderError> {
         let decompressed = grapl_service::decoder::decompress::maybe_decompress(body.as_slice())?;
 
         let mut first_deserialization_error: Option<SysmonDecoderError> = None;
 
-        /*
-           This iterator is taking a set of bytes of the logs, splitting the logs on newlines,
-           converting the byte sequences to utf-8 strings, and then filtering on supported event
-           types: Process Creation, Network Connection, and File Creation.
+        use sysmon_parser::system::EventId;
```
colin-grapl, line 46: even if a use is in a fn body it's global, so it should go at the top

src/rust/generators/sysmon-generator/src/serialization.rs
```diff
@@ -26,70 +26,66 @@ impl CheckedError for SysmonDecoderError {
...
-        */
-        let events: Vec<_> = decompressed
-            .split(|byte| *byte == b'\n')
-            .map(String::from_utf8_lossy)
-            .map(|s| s.to_string())
-            .filter_map(|event_str| {
-                let parsed_event = Event::from_str(&event_str);
-                match parsed_event {
-                    Ok(event) => {
+        let xml = String::from_utf8_lossy(decompressed.as_slice());
```
colin-grapl, line 48: I think you can use str::from_utf8_lossy? but idk if that's better

.github/CODEOWNERS
inickles-grapl: I know that'd match other crates in this repo, but there's this https://rust-lang.github.io/rfcs/0940-hyphens-considered-harmful.html and the idea that we're moving this off this repo next anyway.

Thoughts?

src/rust/sysmon_parser/benches/benches.rs
```diff
@@ -0,0 +1,49 @@
...
+use sysmon_parser::SysmonEvent;
+
+pub fn bulk_bench(c: &mut Criterion) {
+    let events6 = std::fs::read_to_string("tests/data/events6.xml").unwrap();
+
+    c.bench_function("bulk - events6", |b| {
+        b.iter(|| {
+            let results = sysmon_parser::parse_events(&events6);
+            for result in results {
+                result.unwrap();
```
colin-grapl, line 17: might wanna black box the result.unwrap()

src/rust/sysmon_parser/src/error.rs
```diff
@@ -0,0 +1,30 @@
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// Represents all possible errors that can occur when parsing Sysmon XML events.
+#[derive(Error, Debug, PartialEq)]
```
colin-grapl, line 6: https://blog.rust-lang.org/2019/12/19/Rust-1.40.0.html#non_exhaustive-structs-enums-and-variants

colin-grapl (approved):
some minor comments

.github/CODEOWNERS
colin-grapl: hyphens own

src/rust/generators/sysmon-generator/src/models/file/create.rs
```diff
@@ -24,28 +27,27 @@ use crate::{
...
+        .with_process_name(get_image_name(&event_data.image))
         .with_last_seen_timestamp(timestamp);
 
     let mut file = FileNode::new(FileNode::session_strategy());
-    file.with_asset_id(file_create.system.computer.computer.clone())
-        .with_file_path(strip_file_zone_identifier(
-            &file_create.event_data.target_filename,
-        ))
+    file.with_asset_id(&system.computer)
+        .with_file_path(strip_file_zone_identifier(&event_data.target_filename))
```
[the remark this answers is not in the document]
inickles-grapl: Yeah, sounds good.

src/rust/generators/sysmon-generator/src/models/mod.rs
```diff
@@ -114,21 +103,19 @@ fn strip_file_zone_identifier(path: &str) -> &str {
 }
 
 /// Gets the name of the process given a path to the executable.
-fn get_image_name(image_path: &str) -> Option<String> {
+fn get_image_name(image_path: &str) -> String {
```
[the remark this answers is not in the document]
inickles-grapl: I don't even really like that we're doing this, I think we should drop this and pass up the full path, especially because it's Windows specific, but the reason why I changed the return type is because the generator had calls to `.unwrap()` after calling this, which was not acceptable.

.github/CODEOWNERS
inickles-grapl: I'll use dashes for now, we can consider switching if/when we do move.

src/rust/generators/sysmon-generator/src/serialization.rs
inickles-grapl: Sounds good.

src/rust/generators/sysmon-generator/src/serialization.rs
inickles-grapl: There is not a lossy variant for str, only String. FWIW `String::from_utf8_lossy` returns a Cow.

I'd rather not even use lossy, but I got a Utf8Error trying to use str::from_utf8. I thought I made event6.xml UTF8, so I was wondering if it was an artifact of the compression, which seems like it shouldn't be the case, but I also figured I shouldn't worry about changing that in this PR. The previous uses the same `String::from_utf8_lossy`.

src/rust/sysmon_parser/src/error.rs
inickles-grapl: TIL this looks like what I want in other places too.

colin-grapl (approved):

PR #1504, opened 2022-03-02
Assorted Makefile cleanups

christophermaier:
This is a bit of a grab-bag of tweaks to our Makefile. In general, I've tried to split things into targets as atomically as possible, in order to more clearly reveal the dependencies between them (as well as make it possible to run particular tasks by themselves, which in some cases was not possible until now). I've also tried to format things consistently to make the file easier to read and navigate.

Probably the most interesting thing is the addition of standard labels to all our images, which we can use to make the various Docker-related `clean-*` targets function properly again.

It's best to review this PR commit-by-commit.


jgrillo-grapl (approved):
👏 🎆 💯

PR #1506, opened 2022-03-03
Custom Header to enable ingress gateway tracing

ghost:
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/862

This is part of a larger distributed tracing project.
This particular piece adds a custom header to calls to the backend. This custom header is required in order to include ingress gateways in the trace.
Per https://www.consul.io/docs/connect/distributed-tracing#considerations

 `Requests through [Ingress Gateways](https://www.consul.io/docs/connect/gateways/ingress-gateway) will not be traced unless the header x-client-trace-id: 1 is set (see [hashicorp/consul#6645](https://github.com/hashicorp/consul/issues/6645)).`

1. Tested as part of a larger tracing branch, where traces include the ingress gateway.
2. Regression tests should ensure that this doesn't break any functionality

### DIFF
--- a/src/js/engagement_view/src/serviceWorker.ts
+++ b/src/js/engagement_view/src/serviceWorker.ts
@@ -106,7 +106,12 @@ function registerValidSW(swUrl: string, config?: Config) {
 function checkValidServiceWorker(swUrl: string, config?: Config) {
     // Check if the service worker can be found. If it can't reload the page.
     fetch(swUrl, {
-        headers: { "Service-Worker": "script" },
+        headers: {
+            "Service-Worker": "script",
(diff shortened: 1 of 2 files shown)


twunderlich-grapl:
@andrea-grapl I _think_ I've added the header to all calls to our backend, however please let me know if there's somewhere else that we need to add this header to.

andrea-grapl (approved):
Looks good!

andrea-grapl (commented):
@twunderlich-grapl  Not sure where we want tracing, but we also have headers in `engagement_view/src/serviceWorker.ts`, will leave up to you if we want to add `x-client-trace-id` there too.

PR #1507, opened 2022-03-03
Use system zstd library for compiling zstd crate

christophermaier:
Rather than compiling `libzstd` during our `cargo build` process, we
can just link against a system library.

This change appears to shave off ~20 seconds from the overall compile
time, as judged from the output of `cargo +nightly build --timings`.

### DIFF
--- a/etc/chromeos/lib/installs.sh
+++ b/etc/chromeos/lib/installs.sh
@@ -44,7 +44,9 @@ install_build_tooling() {
         apt-utils
         build-essential
         libclang1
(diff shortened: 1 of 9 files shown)


jgrillo-grapl (approved):

christophermaier:
Closing this for now; we would like to keep our distroless deployment container base, and don't really want to have to manually juggle adding various system libraries to it to ensure our binaries continue to work. If we come up with a better solution to that later on, we'll revisit this.

PR #1509, opened 2022-03-03
Send local pulumi traces to Jaeger

ghost:

This sends pulumi's traces to Jaeger when running local grapl.

Note that this relies on https://github.com/grapl-security/grapl/pull/1508 being merged in first

`make up` on a branch with the observability job merged in.
Go to http://localhost:16686/search, you should see the pulumi traces.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@ services:
     command:
       - |
         pulumi login --local
-        export TRACING_BACKEND_URL="http://host.docker.internal:9411/api/v1/spans"
+        TRACING_BACKEND_URL="http://host.docker.internal:9411/api/v1/spans"
 
         (
           cd grapl
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
         (
           cd grapl
           pulumi stack init local-grapl --non-interactive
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
           # Ideally this stack name would be scoped by organization, but
           # login --local doesn't support organizations
           pulumi stack init local-grapl-integration-tests --non-interactive
-          pulumi up --yes --skip-preview --stack=local-grapl-integration-tests
+          pulumi up --yes --skip-preview --stack=local-grapl-integration-tests --tracing http://host.docker.internal:9411/api/v1/spans
```
christophermaier, line 34: Since this command is basically one big chunk of bash script, can we set this URL as a variable up at the top and just reference it in both these places?

christophermaier (approved):

docker-compose.yml
```diff
@@ -14,11 +14,12 @@ services:
     command:
       - |
         pulumi login --local
+        export TRACING_BACKEND_URL="http://host.docker.internal:9411/api/v1/spans"
```
christophermaier, line 17: Tiny nitpick: I don't think `export` is required here; the variable doesn't need to be available to subprocesses, since you're passing it directly to the process via the command line option.

PR #1513, opened 2022-03-03
Instrument graphql-endpoint for tracing

ghost:

This instruments the graphql-endpoint for tracing.

This relies on OTEL_EXPORTER_ZIPKIN_ENDPOINT being set as an environment variable in Nomad. This will be added in a different PR.

Tested as part of a larger local-trace branch using make test-e2e and then checking Jaeger to confirm that the graphql_endpoint traces exist.
This by itself doesn't enable any functionality


andrea-grapl (approved):
Looks good, thanks for taking care of this. I know we had the existing `console.debug` which got changed to `console.log` - wondering if we should even be logging this? Will leave to your discretion, otherwise, looks good!

PR #1515, opened 2022-03-03
Instrument python code for tracing

ghost:

Instrument python code for tracing

We're manually instrumenting our code for the moment, since most of our
python code isn't using frameworks. This does rely on setting an
endpoint via env vars, which should make operations easier between
local and AWS

From the larger local-tracing branch.
The automated test suite should ensure that this doesn't break the code even without a true endpoint


3rdparty/python/requirements.txt
```diff
@@ -1,13 +1,20 @@
 # Any Python dependencies used by our code should be declared here.
 # !!!!! Don't forget to regenerate the constraints from this! (see readme.md)
 
+# Please keep this alphabetical order
```
christophermaier, line 4: 🙏 🙏 🙏

src/python/analyzer_executor/src/run.py
```diff
@@ -23,28 +25,31 @@ async def _main() -> None:
...
-            queue_url=queue_url,
-            receipt_handle=sqs_message.receipt_handle,
-            message_id=sqs_message.message_id,
-            coroutine=message_handle_coroutine,
-        )
-        try:
-            await timeout_manager.keep_alive()
-        except SqsTimeoutManagerException:
-            LOGGER.error("SQS Timeout Manager exception", exc_info=True)
+    with TRACER.start_as_current_span("_main"):
```
christophermaier, line 28: Might be better to use `_main.__name__` instead of `"_main"` for this span name, just to guard against refactoring-induced errors in the future.

src/python/engagement-creator/engagement_creator/run.py
```diff
@@ -5,41 +5,44 @@
...
-            queue_url=queue_url,
-            receipt_handle=sqs_message.receipt_handle,
-            message_id=sqs_message.message_id,
-            coroutine=message_handle_coroutine,
-        )
-        try:
-            await timeout_manager.keep_alive()
-        except SqsTimeoutManagerException:
-            LOGGER.error("SQS Timeout Manager exception", exc_info=True)
+    with TRACER.start_as_current_span("engagement_creator"):
```
christophermaier, line 24: Why is this `"engagement_creator"`, rather than `"_main"`, as it is above?

src/python/grapl-common/grapl_common/grapl_tracer.py
```diff
@@ -0,0 +1,54 @@
...
+    # Long-term we'll switch to OTEL
+    zipkin_exporter = ZipkinExporter(
+        # version=Protocol.V2
+        # optional:
+        # endpoint="http://localhost:9411/api/v2/spans",
+        # local_node_ipv4="192.168.0.1",
+        # local_node_ipv6="2001:db8::c001",
+        # local_node_port=31313,
+        # max_tag_value_length=256
+        # timeout=5 (in seconds)
```
christophermaier, line 46: Are these additional arguments to eventually add later?

src/python/provisioner/provisioner/app.py
```diff
@@ -98,22 +101,23 @@ def _retrieve_test_user_password(
...
-
-    LOGGER.info("creating test user")
-    _create_user(
-        dynamodb=dynamodb,
-        username=username,
-        cleartext=password,
-        role="owner",
-    )
-    LOGGER.info("created test user")
+    with TRACER.start_as_current_span("provision"):
```
christophermaier, line 104: Same comments here re: span name.

src/python/grapl-common/grapl_common/grapl_tracer.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+    processor = BatchSpanProcessor(zipkin_exporter)
+    provider.add_span_processor(processor)
+    trace.set_tracer_provider(provider)
+
+    BotocoreInstrumentor().instrument()
```
christophermaier, line 52: As this code stands, this is basically just going to instrument all AWS calls, right?

(I have no issue with that, to be clear; just clarifying my understanding!)

src/python/grapl-common/grapl_common/grapl_tracer.py
```diff
@@ -0,0 +1,54 @@
...
+
+    :param service_name: App/Service name ie provisioner
+    :param module_name: This should be __name__
+    :return: Tracer:
+    """
+    provider = TracerProvider(
+        resource=Resource.create(
+            {
+                "service.name": service_name,
+                # TODO add things like instance_id, etc here
```
christophermaier, line 32: Will we eventually have some kind of schema for what kind of information we want to capture here? Or is there already some commonly-agreed-upon schema out in the wild that we could use (or are currently using in this code)?

christophermaier (approved):
I have some relatively minor questions / suggestions, but otherwise looks good!

src/python/grapl-common/grapl_common/grapl_tracer.py
twunderlich-grapl: Correct. Each AWS call will be a new span in the trace

src/python/grapl-common/grapl_common/grapl_tracer.py
twunderlich-grapl: Yep. We'll be primarily using the opentracing schemas, although I could see us adding a few things that are grapl unique
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md

src/python/grapl-common/grapl_common/grapl_tracer.py
twunderlich-grapl: These were the default arguments. I'm deleting them since I don't think we'll be adjusting them in here. More likely, we'll control opentelemetry with environment variables going forward

twunderlich-grapl:
@christophermaier 
I've set the span name to always be the method/function name via `__name__`. I think that should be the pattern going forward.

jgrillo-grapl (approved):

PR #1519, opened 2022-03-04
Internal 858 refactor rust proto

jgrillo-grapl:

https://github.com/grapl-security/issue-tracker/issues/858

This PR makes the following changes to our wire format:
- add `Duration` and `Timestamp` types to `graplinc.common.v1beta1` which replace our usage of `google.protobuf.{Timestamp, Duration}`. The reason for this is that Google's types use signed integers to represent time, which is a little weird--especially for a `Duration` which is a measurement of _length_.
- Update messages which previously used `google.protobuf.Timestamp` to instead use `graplinc.common.v1beta1.Timestamp`

As a consequence, our `python-proto` library has been updated to reflect these changes.

Additionally, this PR introduces a `rust-proto-new` library which is spiritually the same as `python-proto` in that it encapsulates all the protocol buffer code (e.g. code generated by prost) and provides a simple API to the user that allows them to either deserialize a struct from bytes or serialize a struct to bytes. We will port services over to this new `rust-proto-new` as we migrate them to kafka.

Note that currently we have not achieved feature parity with `python-proto`. The changes in this PR represent the minimum functionality required to build the `pipeline-ingress` service. Subsequent pull requests will add more protocol buffers as we migrate pipeline services to kafka.

Automated tests. See in particular `rust-proto-new/tests`.


src/python/python-proto/python_proto/pipeline.py
```diff
@@ -86,7 +86,8 @@ def into_proto(self) -> _Envelope:
         proto_envelope.metadata.CopyFrom(self.metadata.into_proto())
         inner_message = _Any()
         inner_message.Pack(
-            self.inner_message.into_proto(), type_url_prefix="".encode("utf-8")
+            self.inner_message.into_proto(),
+            type_url_prefix="graplsecurity.com".encode("utf-8"),
```
colin-grapl, line 90: You could also just do b"graplsecurity.com"

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
```diff
@@ -0,0 +1,197 @@
...
+            msb: u64::from_le_bytes(upper),
+        }
+    }
+}
+
+impl type_url::TypeUrl for Uuid {
+    const TYPE_URL: &'static str = "graplsecurity.com/graplinc.common.v1beta1.Uuid";
+}
+
+impl SerDe for Uuid {
```
colin-grapl, line 59: You could just write a blanket impl for T: Message + url::TypeUrl, no need to have this impl duplicated for every type.

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
```diff
@@ -0,0 +1,197 @@
...
+    BufMut,
+};
+use prost::Message;
+pub use uuid::Uuid;
+
+use crate::{
+    protobufs::graplinc::common::v1beta1::{
+        Duration as _Duration,
+        Timestamp as _Timestamp,
+        Uuid as _Uuid,
```
colin-grapl, line 21: I think the consensus is that rather than using underscore for the protos we should just add Proto at the end. This came up in the rust-proto PR.

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
jgrillo-grapl: what is a blanket impl?

jgrillo-grapl: Oh, sorry, I misunderstood your comment. You were referring to `SerDe` not `TypeUrl`. I'll look into that.

src/rust/rust-proto-new/src/graplinc/grapl/pipeline/v1beta1.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+impl TryFrom<_Metadata> for Metadata {
+    type Error = SerDeError;
+
+    fn try_from(metadata_proto: _Metadata) -> Result<Self, Self::Error> {
+        let tenant_id = if let Some(tenant_id) = metadata_proto.tenant_id {
+            tenant_id
+        } else {
+            return Err(SerDeError::MissingField("tenant_id absent".to_string()));
```
colin-grapl, line 46: I'd remove the "absent" bit from all of these. The string should just be the field name. Something like "absent" belongs in the Error's Display or Debug impls.

src/rust/rust-proto-new/src/graplinc/grapl/pipeline/v1beta1.rs
```diff
@@ -0,0 +1,236 @@
...
+    metadata: Metadata,
+    inner_type: String,
+    inner_message: Bytes,
+}
+
+impl TryFrom<_Envelope> for Envelope {
+    type Error = SerDeError;
+
+    fn try_from(envelope_proto: _Envelope) -> Result<Self, Self::Error> {
+        let metadata = if let Some(metadata) = envelope_proto.metadata {
```
colin-grapl, line 141: In cases like this you can write:
```rust
let metadata = envelope_proto.metadata.ok_or(SerDeError::MissingField("metadata absent".to_string()))?;
```

colin-grapl (commented):
Some minor comments.

src/python/python-proto/python_proto/pipeline.py
jgrillo-grapl: I'm not 100% sure these things are actually identical? A `b""` is a bytes literal but `"".encode("utf-8")` is saying "encode my string in UTF-8 bytes". That is, the important thing here is the encoding. While I _suspect_ this source file is UTF-8 encoded, I'm not sure whether that's specified anywhere?

jgrillo-grapl: hmm there's some more clarity here https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

in particular,

> The source character set is defined by the encoding declaration; it is UTF-8 if no encoding declaration is given in the source file; see section [Encoding declarations](https://docs.python.org/3/reference/lexical_analysis.html#encodings).

and

> Bytes literals are always prefixed with `'b'` or `'B'`; they produce an instance of the [`bytes`](https://docs.python.org/3/library/stdtypes.html#bytes) type instead of the [`str`](https://docs.python.org/3/library/stdtypes.html#str) type. They may only contain ASCII characters; bytes with a numeric value of 128 or greater must be expressed with escapes.

So, given that ASCII is a subset of UTF-8 a `b".."` here is equivalent to `"..".encode("utf-8")`. Will fix.

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
jgrillo-grapl: The reason I used an underscore here was to signify that these are _internal_ to this crate, and it's the same as what we do in `python-proto`. In python a leading underscore has a special meaning whereas in rust we can (and do here) use a more powerful mechanism to restrict visibility. We can change it to say `Proto` at the end, but does it actually make anything clearer? I kind of like the underscore here because it makes these things _ugly_ (which they are) and clearly marks them as "not for public consumption".

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
jgrillo-grapl: I'm not sure this is quite so simple, because some types are `From` and some are `TryFrom`. I've implemented `From` for infallible conversions and `TryFrom` otherwise. I'm not sure how I can accommodate both in one blanket implementation.

src/rust/rust-proto-new/src/graplinc/grapl/pipeline/v1beta1.rs
jgrillo-grapl: fixed

src/rust/rust-proto-new/src/graplinc/grapl/pipeline/v1beta1.rs
jgrillo-grapl: fixed

src/python/python-proto/python_proto/pipeline.py
jgrillo-grapl: fixed

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
jgrillo-grapl: Actually, thinking about this further.. are you suggesting adding a type parameter to `SerDe`? Like parametrize `SerDe` by the protobuf stub type? This is how the python code works, due to inheritance (I think?) but I was able to get by _without_ doing that here..

src/proto/graplinc/common/v1beta1/types.proto
```diff
@@ -11,3 +11,30 @@ message Uuid {
   // Most significant bytes
   fixed64 msb = 2;
 }
+
```
wimax-grapl, line 14: Can we add a comment about why we're rolling our own time primitives? (I'm not questioning your motives, I just want future users to know what's up)

jgrillo-grapl: done

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
```diff
@@ -0,0 +1,230 @@
...
+// SystemTime (a.k.a. Timestamp)
+//
+
+impl TryFrom<_Timestamp> for SystemTime {
+    type Error = SerDeError;
+
+    fn try_from(timestamp_proto: _Timestamp) -> Result<Self, Self::Error> {
+        match timestamp_proto.duration {
+            Some(
+                crate::protobufs::graplinc::common::v1beta1::timestamp::Duration::BeforeEpoch(
```
wimax-grapl, line 130: can we use _Duration::BeforeEpoch? Same for Since

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
```diff
@@ -0,0 +1,230 @@
...
+impl TryFrom<SystemTime> for _Timestamp {
+    type Error = SystemTimeError;
+
+    fn try_from(timestamp: SystemTime) -> Result<Self, Self::Error> {
+        if timestamp >= UNIX_EPOCH {
+            let duration = timestamp.duration_since(UNIX_EPOCH)?;
+            let duration_proto: _Duration = duration.into();
+            Ok(_Timestamp {
+                duration: Some(
+                    crate::protobufs::graplinc::common::v1beta1::timestamp::Duration::SinceEpoch(
```
wimax-grapl, line 157: yeah this is also a _Duration?

wimax-grapl (commented):
Looks very good, love the proptest stuff

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
jgrillo-grapl: for some reason we can't do this, because we get the following error (which admittedly I don't understand): ```
error[E0599]: no associated item named `BeforeEpoch` found for struct `protobufs::graplinc::common::v1beta1::Duration` in the current scope
   --> rust-proto-new/src/graplinc/common/v1beta1.rs:167:32
    |
167 |                     _Duration::BeforeEpoch(
    |                                ^^^^^^^^^^^ associated item not found in `protobufs::graplinc::common::v1beta1::Duration`
    |
   ::: /home/jgrillo/src/grapl/src/rust/target/debug/build/rust-proto-new-f6e7cf9c0a6e6e98/out/graplinc.common.v1beta1.rs:28:1
    |
28  | pub struct Duration {
    | ------------------- associated item `BeforeEpoch` not found for this
```
I also tried to import `BeforeEpoch` and `SinceEpoch` directly e.g. ```
use crate::protobufs::graplinc::common::v1beta1::Duration::{
    BeforeEpoch as _BeforeEpoch,
    SinceEpoch as _SinceEpoch
};```
but that yielded the following error:
```
error[E0432]: unresolved imports `crate::protobufs::graplinc::common::v1beta1::Duration`, `crate::protobufs::graplinc::common::v1beta1::Duration`
  --> rust-proto-new/src/graplinc/common/v1beta1.rs:20:9
   |
20 |         Duration::BeforeEpoch as _BeforeEpoch,
   |         ^^^^^^^^ `Duration` is a struct, not a module
21 |         Duration::SinceEpoch as _SinceEpoch,
   |         ^^^^^^^^ `Duration` is a struct, not a module```
so.. I agree fully qualifying here is _ugly_ but I don't know what else to do.

src/python/python-proto/python_proto/common.py
```diff
@@ -122,3 +95,72 @@ def into_proto(self) -> _Duration:
...
+    @staticmethod
+    def from_proto(proto_timestamp: _Timestamp) -> Timestamp:
+        assert proto_timestamp.WhichOneof("duration") is not None
+        if proto_timestamp.WhichOneof("duration") == "since_epoch":
+            proto_duration = proto_timestamp.since_epoch
+            return Timestamp(
+                duration=Duration.from_proto(proto_duration),
+                before_epoch=False,
+            )
+        elif proto_timestamp.WhichOneof("duration") == "before_epoch":
```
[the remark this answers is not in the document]
jgrillo-grapl: done

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
colin-grapl: I don't personally care either way - I felt similar when I used underscore. Just saying that the convention we've landed on is to add Proto.

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
colin-grapl: Hm, yeah nevermind, not worth figuring it out. I think there's a way to perform write it like this is all:
`impl SerDe for T: Message + TypeUrl {`

But it's not important, and I see that your impl uses From/TryFrom, which is probably for the best

colin-grapl (approved):

src/rust/rust-proto-new/src/graplinc/common/v1beta1.rs
jgrillo-grapl: renamed everything e.g. `_Foo` --> `FooProto`

wimax-grapl (approved):

PR #1520, opened 2022-03-04
Build Rust unit tests in a container, rather than an image

christophermaier:
This is part of a larger effort to make our development process a bit
more friendly. Here, we introduce a very minimal Docker image that we
use to run Rust unit tests. Note that we are not running the tests
during the *building* of the image; rather, we spawn a container from
that image and use *that* as our build environment.

We make use of volumes in order to re-use compilation artifacts from
previous runs, making subsequent runs much quicker.

Another aspect of this change that will speed things up is that we
have split out a dedicated `coverage` task for our Rust
tests. Previously, we exclusively ran unit tests under Tarpaulin to
gather coverage statistics. Unfortunately, due to how Tarpaulin
operates, this required recompilation of *everything* each time tests
were run. Since coverage statistics are mainly of interest only in CI,
it makes little sense to impose this burden on developers in their
code-test-debug cycle. Now, invoking `make test-unit-rust` from the
repository root will simply invoke the equivalent of `cargo test`,
which will run much more quickly. Should developers wish to capture
coverage statistics for their Rust code locally, they can run `make
test-unit-rust-coverage`, which runs the tests under Tarpaulin. This
target is what is run in our CI pipeline now.

A Makefile has been added to our Rust source directory to facilitate
localized management of Rust-related tasks. The top-level Makefile has
been updated to delegate to this new Makefile when appropriate, so
developers can continue to use the top-level Makefile to drive their
day-to-day interactions with the code.

Note that this only addresses Rust *unit* tests; integration tests and
general release builds are still handled as they were prior to this
commit. Bringing those tasks in line with this new approach is future
work. (The `build` target in `src/rust/Makefile` is only used here in
the context of the existing `grapl-template-generator` test script,
and is not yet used to build our final deployable artifacts.)

Interesting techniques used in this commit include:
- Using a "sentinel file" in `src/rust/Makefile` to trigger build
environment image rebuilds only when necessary, rather than on every
target invocation.
- Introspection of `src/rust/rust-toolchain.toml` to automatically
keep the build environment image in sync with our Rust toolchain.
- Dynamic generation of makefile targets for cleaning each volume used
by the build environment container.
- Introspection of the build environment image to automatically
determine the appropriate mount points for the volumes, based on
exported environment variables.


.github/CODEOWNERS
```diff
@@ -71,19 +71,17 @@ BUILD @christophermaier
 /src/rust/grapl-service/ @inickles-grapl
 /src/rust/grapl-utils/ @inickles-grapl
 /src/rust/grapl-web-ui/ @inickles-grapl
+/src/rust/Makefile @christophermaier
 /src/rust/model-plugin-deployer/ @wimax-grapl
+/src/rust/node-identifier/ @grapl-security/wg-data-infra
```
twunderlich-grapl, line 76: FYI this team doesn't appear to exist according to github. Can you double-check the name here?

jgrillo-grapl (approved):
🎉

Makefile
```diff
@@ -264,6 +262,7 @@ test-unit-js: test-unit-graphql-endpoint
 test-unit-js: ## Build and run unit tests - JavaScript only
 
 .PHONY: test-unit-graphql-endpoint
+test-unit-graphql-endpoint: | dist
```
twunderlich-grapl, line 265: What does the pipe in front of the dist target do?

twunderlich-grapl (approved):

Makefile
christophermaier: It marks `dist` as an [order-only prerequisite](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html) for this target. This is basically the prototypical usecase for such prerequisites (detailed toward the bottom of the linked page).

PR #1521, opened 2022-03-06
Add `cargo udeps` job to CI

christophermaier:
It's good to periodically check to ensure we're using all our Rust
dependencies.

It runs each time `Cargo.toml` or `Cargo.lock` files are changed; any
failures are purely advisory. Like our `cargo audit` job, it also runs
on a weekly basis with non-advisory failures.


twunderlich-grapl (approved):
I wish python had a good way to do this

colin-grapl (approved):

PR #1531, opened 2022-03-10
React dependency upgrades

andrea-grapl:
Update react scripts, typescript errors, add a proxy file instead of a proxy line in `package.json` 

`make up` `yarn run build`

NOTE: We changed the proxy in our `package.json` flie so this might fail


src/js/engagement_view/node_modules/.gitignore
```diff
@@ -1,6 +0,0 @@
-# This exists to keep the existence of the node_modules directory in git.
```
christophermaier, line 1: Probably best to keep this file in the repository for the time being; I'll be digging into it in my PR that'll follow this one.

src/js/graphql_endpoint/package-lock.json
```diff
@@ -14,8 +14,14 @@
         "@grpc/grpc-js": "^1.2.2",
         "@opentelemetry/api": "^1.0.4",
         "@opentelemetry/exporter-zipkin": "^1.0.1",
+        "@opentelemetry/instrumentation-express": "^0.28.0",
```
christophermaier, line 17: Should this file be in this PR?

(I ask because nothing else in this PR does anything with the `graphql_endpoint` code.)

andrea-grapl: Probably not, but it fixes a yarn audit before I knew it was a soft fail I'm taking out a separate PR to fix all yarn audits soon. Can downgrade though.

.gitignore
```diff
@@ -25,6 +25,7 @@ __pycache__/
 target/
 
 # Node packages
+# This exists to keep the existence of the node_modules directory in git.
```
twunderlich-grapl, line 28: So this comment is incorrect. This actually means that you cannot commit any changes within the node_modules directory to git

src/js/engagement_view/.yarnrc.yml
```diff
@@ -7,3 +7,6 @@ httpTimeout: 60000
 installStatePath: .yarn/state/install-state.gz
 
 nodeLinker: node-modules
+
+#Uncomment the following line for local development
+#yarnPath: .yarn/releases/yarn-3.1.1.cjs
```
twunderlich-grapl, line 12: I'm about to put a PR up that will remove the need for this. You can delete it or leave it

colin-grapl (approved):

PR #1532, opened 2022-03-10
Local tracing

ghost:

This sets up the necessary bits in order to get local tracing working.
Specifically:
This sets the protocol for everything (defaulting to grpc with overrides for http)
This creates a jaeger config. Notably the trace_id_128bit is required for Jaeger
Sets the tracing endpoint with the private ip address associated with eth0 interface
Sets tracing addresses via env vars

make up 
Go to localhost:1234, log in and go to the engagements (to get some data that hits graphql)
Go to http://localhost:16686/search (Jaeger)
There should be traces for web-ui, ingress service etc which are sent via the consul config


pulumi/infra/consul_config.py
```diff
@@ -0,0 +1,83 @@
...
+                        {
+                            "http": {
+                                "name": "envoy.tracers.zipkin",
+                                "typedConfig": {
+                                    "@type": "type.googleapis.com/envoy.config.trace.v3.ZipkinConfig",
+                                    "collector_cluster": "zipkin",
+                                    "collector_endpoint_version": "HTTP_JSON",
+                                    "collector_endpoint": "/api/v2/spans",
+                                    "shared_span_context": False,
+                                    "trace_id_128bit": True,
```
twunderlich-grapl, line 64: This is apparently the magic config required to get tracing working with Jaeger. Without it, this fails silently

pulumi/infra/consul_config.py
```diff
@@ -0,0 +1,83 @@
...
+                                },
+                            }
+                        }
+                    ),
+                    "prometheus_bind_addr": "0.0.0.0:9102",
+                    "protocol": "grpc",
```
twunderlich-grapl, line 70: We're setting the default protocol for EVERYTHING to grpc here

pulumi/infra/config.py
```diff
@@ -234,3 +235,11 @@ def container_repository() -> Optional[str]:
     if repo_name:
         return f"docker.cloudsmith.io/{repo_name}"
     return None
+
+
+# This should be equivalent to what HOST_IP_IN_NOMAD resolves to. However, we need this for the consul proxy defaults,
+# so we can't use var interpolation. This should be temporary until we get consul dns working within containers
+def get_nomad_ip() -> str:
```
twunderlich-grapl, line 242: I forgot that this runs inside a docker container so this won't work. It will need to be set manually

christophermaier (approved):

wimax-grapl (approved):

pulumi/grapl/__main__.py
```diff
@@ -307,6 +302,33 @@ def main() -> None:
...
+        protocol="http",
+        opts=pulumi.ResourceOptions(provider=consul_provider),
+    )
+
+    ConsulServiceDefault(
+        "graphql-endpoint",
+        service_name="graphql-endpoint",
+        protocol="http",
+        opts=pulumi.ResourceOptions(provider=consul_provider),
+    )
```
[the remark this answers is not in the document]
twunderlich-grapl: I'm undecided at the moment, but we definitely could.

PR #1535, opened 2022-03-11
Always use the correct yarn version

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/877

Set a single source of truth for yarn version via corepack.
Set a node version which will generate a warning when the wrong node version is used (unless someone has engine-strict set in npmrc)

Test suite for testing the container changes.

For running node locally, 
1. run the chromeos script to install the new node version and enable corepack (you can run the commands manually).
2. `cd src/js/engagement_view && yarn run build`


src/js/engagement_view/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "grapl-fe",
   "version": "0.1.0",
   "private": true,
+  "engines": {
```
twunderlich-grapl, line 5: This sets a warning if the wrong version of node is used

jgrillo-grapl (approved):

src/js/engagement_view/package.json
andrea-grapl: Looks good, wondering if you can take a look at the last line of `src/js/engagement_view/package.json`. When `packageManager: yarn@3.1.1.` is removed, @cm and I noticed the project still builds with tests passing locally, wondering if this is necessary?

andrea-grapl (approved):
Well done, thank you for fixing this so quickly!

src/js/engagement_view/package.json
twunderlich-grapl: So the engines section is not necessary since it will only warn not error out.
The packageManager setting is necessary since that's what corepack uses to figure out the yarn version. Without that set, you'll use whatever the default yarn version is for your system. Since I set up yarn 3.1.1 with you manually, it's probably using the correct version. Alternatively, it may just build correctly even with yarn 1.x.x

christophermaier (approved):
I hate that this is so complicated 😭

PR #1539, opened 2022-03-15
Update .dockerignore to match all sub-directories under src/rust named 'target'

inickles-grapl:

This updates the top-level .dockerignore file to match not just `src/rust/target`, but all directories named `target` under `src/rust`, which can happen when building one of the crates individually.

I realized I had large (hundreds of MB) context transfers for Rust Dockerfile stages, which didn't seem right. Turns out this was due to a directory `src/rust/generators/sysmon-generator/target` I had from working with that crate directly.

With this change I noticed these context transfer sizes were much closer to the size of the sources under proto and rust.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,5 +7,9 @@
 .pids
 .yarn
 build-support/grapl-venv
-src/rust/target/**
+
+# match any directory named 'target', which can result from building creates
+# individually as one might do during local development.
+src/rust/**/target/**
+
 test_artifacts/**


twunderlich-grapl (approved):

christophermaier (commented):
This is fine, but I can't actually replicate the original problem; no matter what crate directory I go into, I only end up with a `target` directory in the workspace root.

inickles-grapl:
> This is fine, but I can't actually replicate the original problem; no matter what crate directory I go into, I only end up with a `target` directory in the workspace root.

Looks like the RLS extension for VS Code is creating it. I closed my instance of VS Code, removed my `src/rust/generators/sysmon-generator/target`, started VS Code to load `src/rust/generators/sysmon-generator` and the target directory is back again.

inickles-grapl:
> > This is fine, but I can't actually replicate the original problem; no matter what crate directory I go into, I only end up with a `target` directory in the workspace root.
> 
> Looks like the RLS extension for VS Code is creating it. I closed my instance of VS Code, removed my `src/rust/generators/sysmon-generator/target`, started VS Code to load `src/rust/generators/sysmon-generator` and the target directory is back again.

I switched back to using the rust-analyzer extension, disabling the RLS extension, which is what I was using up to a couple days ago (until it was having issues) and I'm no longer able to reproduce. Still seems worthwhile to keep this IMO, but I wouldn't mind if someone wanted to revert.

PR #1545, opened 2022-03-17
Tracing docker buildx commands

ghost:

This adds commands to enable/disable tracing for docker buildx bake

Run `make up` (so you have Jaeger locally)
Run `make buildx-tracing`
Run `make build-local-infrastructure` or other build command as long as it uses buildx
Go to http://localhost:16686/ and look at the buildkite traces.
Run `make disable-buildx-tracing`
Run the build again 
Refresh Jaeger. No new traces will be there


Makefile
```diff
@@ -633,6 +633,19 @@ generate-nomad-rust-client: ## Generate the Nomad rust client from OpenAPI
 # should use that here.
 	$(MAKE) format-prettier
```
wimax-grapl, line 635: 1. Could we get a quick README or something on how to use and consume this?
2. Not necessarily for _this_ PR but would it make sense to exercise this in CI as a "hey here's proof the jaeger setup all still works"?

wimax-grapl (approved):

Makefile
wimax-grapl: Also, in general, it seems odd to me that this is a
1. enable
2. do things
3. disable

and not a 
1. WITH_JAEGER_TRACING=1 do things

wish we had context managers or RAII in bash...

christophermaier: I had a similar thought... I wonder if we could have a one-time "setup" target that creates the tracing-enabled builder, and then a `WITH_TRACING` macro / flag / etc. that would add the `--builder=XXX` option to the `docker buildx bake` invocations.

``` make
ifdef WITH_TRACING
buildx_builder_args := --builder=grapl-tracing
endif

DOCKER_BUILDX_BAKE_HCL := docker buildx bake --file=docker-bake.hcl $(buildx_builder_args)
```
Then you can do `WITH_TRACING=1 make local-infrastructure` to get tracing enabled (given the builder exists, of course).

(If there was anything else to add tracing to, we could stick that behind the `WITH_TRACING` variable, too.)

It's still not airtight, though, because it still relies on a) having the builder set up and b) that Jaeger is running.

Make is kinda weird for this 🤷 

(These are just thoughts... either way is fine; I'm thinking out loud here)

Makefile
```diff
@@ -633,6 +633,19 @@ generate-nomad-rust-client: ## Generate the Nomad rust client from OpenAPI
...
 
+.PHONY: buildx-tracing
+buildx-tracing: ## Set tracing up for docker buildx, this assumes that Jaeger is already up
+	docker buildx create \
+      --name builder \
```
christophermaier, line 639: It'd be good to name this something more descriptive, like "grapl-tracing", or "grapl-jaeger", etc.

christophermaier (approved):

Makefile
twunderlich-grapl: So since this only sets tracing for docker (we set up tracing for the grapl services without this), I think it makes more sense to rename this docker-tracing than grapl-tracing.

christophermaier: I meant that, among the _n_ potential builders you might one day have on your system, this would be the one that Grapl has set up for tracing builds. It's inherently for Docker since it's a Docker builder 🙃

Makefile
christophermaier: Taking another look, I think the existing code conflates the _creation_ of a builder with the _using_ of that builder. In particular, disabling the builder does not also delete that builder (I don't think that it _should_ do that, just to be clear), so I think we'll have an error the next time we run `make buildx-tracing` because it will try to create another builder with the same name as one that already exists.

Decoupling the creation of the builder from the using or not-using of that builder seems like a worthwhile change.

Makefile
```diff
@@ -29,11 +29,16 @@ export
 export EVERY_COMPOSE_FILE=--file docker-compose.yml \
 	--file ./test/docker-compose.unit-tests-js.yml \
 
+# This is used to send docker traces to Jaeger. This is primarily useful for debugging build time performance
+ifdef WITH_TRACING
+buildx_builder_args := --builder=builder --load
```
christophermaier, line 34: Doesn't `--load` output a tarball? 

https://github.com/moby/buildkit#docker-tarball

christophermaier: 🤔 Hrmm... maybe not; I don't see that it does anything different 🤔

Makefile
twunderlich-grapl: 1. I've added a README in the docs/development folder
2. I've updated the code to use an environment variable and a one-time setup command

Makefile
twunderlich-grapl: Ah, you meant the actual builder name, not the make command 🤦 . Fixing now

PR #1548, opened 2022-03-18
Resolve apt failures with a 'cloud-init status --wait', change to server-minimal us-east-2 AMI

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/880

- Resolve the issues in https://github.com/grapl-security/issue-tracker/issues/880 - there were startup commands happening I didn't know about that were conflicting with our `apt` commands.
- Change to server-minimal; results in slightly shorter build time (fewer things occurring at AMI startup)
- Change to us-east-2; the less hardcoded us-east-1 the better
- Document a CLI command to get latest AMI

thanks thomas for pointing out this possibility!

I'm reasonably convinced this resolves the nondeterminism. I ran it 10x in a row:
```
set -euo pipefail

for i in {1..10}
do
   echo "Attempt ${i}"

   make dist/firecracker_rootfs.tar.gz || \
    (echo "Attempt ${i} failed!" && exit 1)

   rm dist/firecracker_rootfs.tar.gz
done
```

I also tested and saw nondeterministic "sometimes the cloud-init status was already done, sometimes it was still doing things"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -644,6 +644,6 @@ dist/firecracker_kernel.tar.gz: firecracker/generate_firecracker_kernel.sh | dis
 FIRECRACKER_ROOTFS_FILES := $(shell find firecracker/packer -type f)
 dist/firecracker_rootfs.tar.gz: $(FIRECRACKER_ROOTFS_FILES) | dist
 	packer init -upgrade firecracker/packer/build-rootfs.pkr.hcl
-	packer build -on-error=ask \
+	packer build \
 	 	-var dist_folder="${GRAPL_ROOT}/dist" \
 		firecracker/packer/build-rootfs.pkr.hcl
--- a/Makefile
+++ b/Makefile
@@ -644,6 +644,6 @@ dist/firecracker_kernel.tar.gz: firecracker/generate_firecracker_kernel.sh | dis
 FIRECRACKER_ROOTFS_FILES := $(shell find firecracker/packer -type f)
 dist/firecracker_rootfs.tar.gz: $(FIRECRACKER_ROOTFS_FILES) | dist
 	packer init -upgrade firecracker/packer/build-rootfs.pkr.hcl
-	packer build \
+	packer build -on-error=ask \
 	 	-var dist_folder="${GRAPL_ROOT}/dist" \
 		firecracker/packer/build-rootfs.pkr.hcl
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -644,6 +644,6 @@ dist/firecracker_kernel.tar.gz: firecracker/generate_firecracker_kernel.sh | dis
 FIRECRACKER_ROOTFS_FILES := $(shell find firecracker/packer -type f)
 dist/firecracker_rootfs.tar.gz: $(FIRECRACKER_ROOTFS_FILES) | dist
 	packer init -upgrade firecracker/packer/build-rootfs.pkr.hcl
-	packer build \
+	packer build -on-error=ask \
```
christophermaier, line 647: How is this going to behave in CI if there's a failure? I don't think there's any way to get onto the builder to debug (at least, not how we're currently set up).

firecracker/packer/build-rootfs.pkr.hcl
```diff
@@ -16,7 +16,20 @@ variable "instance_type" {
 variable "region" {
   description = "In which region to build the rootfs. Must match the Base AMI's region."
   type        = string
-  default     = "us-east-1"
+  default     = "us-east-2"
```
christophermaier, line 19: Why the region switch?

firecracker/packer/build-rootfs.pkr.hcl
```diff
@@ -86,6 +97,12 @@ source "amazon-ebs" "grapl-build-rootfs" {
 build {
   sources = ["source.amazon-ebs.grapl-build-rootfs"]
 
+  # Prevents non-deterministic apt failures in install_dependencies.sh
+  # https://github.com/grapl-security/issue-tracker/issues/880
+  provisioner "shell" {
+    inline = ["cloud-init status --wait"]
+  }
```
christophermaier, line 104: Nice work tracking this down; nice clean solution, too 👍

christophermaier (approved):

Makefile
wimax-grapl: WOOPS THIS IS NOT INTENTIONAL

firecracker/packer/build-rootfs.pkr.hcl
wimax-grapl: As a general rule, the less hardcoded us-east-1s the better

PR #1549, opened 2022-03-18
Improve logging in Sysmon generator

inickles-grapl:

This improves logging in the Sysmon generator in a number of small ways:
- add new tracing instrumentation.
- provide additional logging.
- skip instrumenting `body` param for impl for PayloadDecoder. that was blowing up logs and wasn't helpful.
- remove redundant event filtering.

Inspecting logs from the Sysmon generator before and after these changes while debugging the generator.

### DIFF
--- a/src/rust/generators/sysmon-generator/src/serialization.rs
+++ b/src/rust/generators/sysmon-generator/src/serialization.rs
@@ -1,9 +1,6 @@
 use grapl_service::decoder::decompress::PayloadDecompressionError;
 use sqs_executor::event_decoder::PayloadDecoder;
-use sysmon_parser::{
-    system::EventId,
-    SysmonEvent,
-};
+use sysmon_parser::SysmonEvent;
 
 #[derive(Debug, Clone, Default)]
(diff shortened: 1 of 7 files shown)


src/rust/generators/sysmon-generator/src/serialization.rs
```diff
@@ -1,47 +1,34 @@
...
                         tracing::error!(
-                            message = "Unable to deserialize Sysmon event",
+                            message = "unable to deserialize Sysmon event",
                             %error,
                         );
                     }
                 }
 
-                result.ok()
+                result.ok().map(|event| event.into_owned())
```
colin-grapl, line 30: could just do `.map(Event::to_owned)` maybe but idc

colin-grapl (approved):

src/rust/generators/sysmon-generator/src/serialization.rs
inickles-grapl: Yeah, I'll make that change.

PR #1552, opened 2022-03-18
Fix 866 - Firecracker kernel .tar.gz was super nested

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/866

Previously if you untar'd the archive it'd be super duper nested in like 5 folders; this fixes it.

Ran the make for the kernel
ran tar -xvf
it extracted the kernel right into the directory :+1:


firecracker/generate_firecracker_kernel.sh
```diff
@@ -32,6 +32,16 @@ cd firecracker
 ########################################
 # Copy kernel into dist.
 ########################################
-readonly KERNEL_BIN_FILE="build/kernel/linux-${KERNEL_VERSION}/vmlinux-${KERNEL_VERSION}-x86_64.bin"
+readonly KERNEL_BIN_DIR="${BUILD_DIR}/firecracker//build/kernel/linux-${KERNEL_VERSION}/"
```
twunderlich-grapl, line 35: did you mean to have the double slash between firecracker and build?

twunderlich-grapl (approved):
One nit, otherwise approved

PR #1555, opened 2022-03-21
Upgrade to actix-web 4

inickles-grapl:

This upgrades to actix-web 4 for the grapl-web-ui, which allows us to get rid of some ugly hacks where we had to mix tokio runtimes.

This has a nice side benefit of reducing the number of duplicate dependencies in our Rust workspace. The number of dependencies goes from 677 to 652. Still a lot, but as d0nut would say "we take those".

There's a bunch of other cleanup I want to do, like clean and improve the errors, but I'll save for another PR.

CI

### DIFF
--- a/src/rust/grapl-web-ui/src/authn/user.rs
+++ b/src/rust/grapl-web-ui/src/authn/user.rs
@@ -9,7 +9,6 @@ use actix_web::{
     ResponseError,
 };
 use futures_util::future::Future;
-use tracing::Instrument;
 
 use crate::authn::{
     AuthDynamoClientError,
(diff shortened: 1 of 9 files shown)


colin-grapl (approved):

src/rust/grapl-web-ui/src/authn/user.rs
```diff
@@ -25,18 +24,16 @@ pub struct AuthenticatedUser {
...
+    #[error("unable to validate user session")]
+    SessionValidation(#[source] AuthDynamoClientError),
+    #[error("unable to get user data from validated session")]
+    UserNotFound(#[source] AuthDynamoClientError),
+    #[error("unable to access session storage")]
+    SessionStorage(#[source] actix_web::Error),
+    #[error("user is not authenticated")]
+    Unauthenticated,
+    #[error("unable to get database client from Actix app data")]
+    DynamoClientUnavailable,
```
christophermaier, line 36: Why are all these errors different if we're just updating Actix?

christophermaier (approved):

src/rust/grapl-web-ui/src/authn/user.rs
inickles-grapl: > Why are all these errors different if we're just updating Actix?

Primarily due to dropping `UserAuthenticationError::Opaque`, which was part of the ugly hack of mixing tokio runtimes (see variant comment). While changing that variant and the function using this error type, these other variants stood out as desperately needing improvement as well.

I plan on making similar improvements to all other error handling in the crate, but I'm saving that for another PR.

PR #1558, opened 2022-03-21
Generate manifest files for rootfs/kernel, recording "what's the version?"

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/881

This does a couple of things.
- Firecracker rootfs will now include a SHA-of-inputs in its version. 
- We'll use the version to replace the checksum check that was in there previously. That way we won't double-upload exactly-the-same rootfs. 
- You might say, "wait, why build and then dedupe post-build?" And you're not wrong. We can find a smarter way to organize these primitives after-the-fact. 
- *Most of this PR is scaffolding to allow artifacts to output a `.manifest` sidecar file that specifies metadata* - for now, just version.

```
BUILDKITE_JOB_ID=abc UPSTREAM_REGISTRY=grapl/wimax UPLOAD_TO_REGISTRY=grapl/wimax .buildkite/scripts/build_and_upload_firecracker_packages.sh
```


jgrillo-grapl (approved):

firecracker/kernel/build.sh
```diff
@@ -47,3 +47,15 @@ tar \
     --create \
     --gzip \
     "${KERNEL_BIN_FILE}"
+
+########################################
+# Write a manifest file containing the version.
+# The version should dedupe based on input file checksums.
+# (This may be replaced with Pants-based file diffs in the future.)
```
christophermaier, line 54: I'm working on this in another repository now, FYI.

.buildkite/scripts/lib/manifest.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
```
christophermaier, line 3: Might be good to write a paragraph here about what the manifest is, why we have it, what it's used for, etc.

.buildkite/scripts/lib/manifest.sh
```diff
@@ -0,0 +1,15 @@
...
+# Results in a single sha256sum of all files in directory
+sha256_of_dir() {
+    local -r dir_path="${1}"
+    find "${dir_path}" -type f -exec sha256sum {} \; | sha256sum | awk '{print $1;}'
```
christophermaier, line 7: I would sort the files before generating the SHA here; running this on my machine yielded output that was definitely _not_ sorted, and since that impacts the final SHA, we should eliminate that source of non-determinism.

```
find "${dir_path}" -type f | sort | xargs sha256sum | sha256sum | awk '{print $1;}'
```

It would also be nice to augment the comment with an example showing exactly what this is doing (you don't need to pick a directory with as many files as this; it's just the directory I was in).

```
# Generates a SHA256 checksum of the sorted output of the SHA256 
# checksums of all the files in this directory (recursively).
#
# $ find . -type f | sort | xargs sha256sum 
# 49a38fd2bd8123074210b621f35abbf8ffe715e5ed764f77fcf0edde70337c5f  ./BUILD
# 503575db67948f54c009463867063fa7cfce651ea2e00c39e81ee8cc4ae69201  ./build_and_upload_firecracker_packages.sh
# 1612b1614be1b0171fa93b89fa703018cf29dd6235be5fce1104836298461fe1  ./build_and_upload_images.sh
# 7d6a4097989400c4b0eb0e9ae0fd6870957e5a136980c6ae2e3d834b20fed103  ./ensure_regenerated_constraints.sh
# 21f84d8a2139892b9cdc755a8f613e85d7d8673b80146dd152fe5c33c2e2f875  ./extract_artifacts.sh
# 49a38fd2bd8123074210b621f35abbf8ffe715e5ed764f77fcf0edde70337c5f  ./grapl_testing_stack/BUILD
# a4431fa56de9407b71138f569e2a5580025e2f1d045a798da37c91cee6b3715e  ./grapl_testing_stack/destroy_stateful_resources.sh
# 0739bc94cee0f49a1a8e389fb2297e44941931ebea71a861ac94289afb44ee53  ./grapl_testing_stack/dump_artifacts.sh
# a96cad5feee9ab749810222daac6c4b34273c6a991477827a0b08e201d57ff73  ./grapl_testing_stack/run_parameterized_job.sh
# 503268de48c7d9ec410e0d98f7171ae4129ab700b1f015fdd5ddf37d8a073ae6  ./lib/artifacts.sh
# 84c15aa349803eb90a2cba09a1325116c68ec6b94397fca75bb69c6260cd1218  ./lib/artifacts_test.sh
# 73ce7319832aec36bec5d844e63a82ae0216e52b83e353f642009d8ec07559bf  ./lib/BUILD
# db37b18d5b9500699077df33b266504933f67131de2ff02dc2cc63e3bd3ea690  ./lib/cloudsmith.sh
# 421ca804d03b5c3ab6aefaa3c4705ec25a61a823fea84ed385d19101a28fb43e  ./lib/lambda.sh
# a9a0d602a7ae3abe48f3b235d1e084d9d08df645b0f088ef5fa270639071b86f  ./lib/lambda_test.sh
# 8364b239db29a18443546a0d1d3d013a32aec8257e64ec2ba511da70caa5b946  ./lib/manifest.sh
# 1c86ebdc5bd4b27d1a3c7d37ca8e40209eb481ead99028ff8712c4f55b4f3da7  ./lib/version.sh
# 777d89e927ae2987f7949df69720fd903523c37250d35eae33b9a3ecd89c598c  ./merge_artifact_files.sh
# 10e6b7d995b77d4b3f4db4066a385cc3b1addb9467b8310f3852bbff389770e5  ./upload_promotion_step_if_needed.sh
# 0f43a2b8e47f36e7218d487907fbb7bfece2b952b4e16fbf8602d1dc3820916f  ./validate_codeowners.sh
#
# Taking the SHA256 checksum of this output yields the final 
# checksum:
#
# $ find . -type f | sort | xargs sha256sum | sha256sum
# 5b611bf839fb19c58f57187053111ff82ebb72cb772a5342b569ec420d39e9e7  -
```

.buildkite/scripts/lib/manifest.sh
```diff
@@ -0,0 +1,15 @@
...
+}
+
+manifest_contents() {
+    local -r version="${1}"
+    jq --null-input \
+        --arg version "${version}" \
+        '{"version": $version}'
```
christophermaier, line 14: I trust you are anticipating additional content for this manifest in the near future? Otherwise, this `jq` invocation is _way_ overkill.

christophermaier: Also, calling this a "manifest" is a bit misleading, since that generally implies some kind of enumeration of contents.

The manifest could end up being something like a structured version of `find "${dir}" -type f | sort | xargs sha256sum`, and then the final checksum would just be a checksum of _that file_.

(But even then, that's just a listing of the _source_ files, and not a listing of the files in the final generated artifact (along with their individual checksums)... so it's more of an "input manifest".)

.buildkite/scripts/build_and_upload_firecracker_packages.sh
```diff
@@ -5,56 +5,32 @@ set -euo pipefail
...
-# This is the list of packages that actually have different shas
-new_packages=()
-
-# Returns a string like `2de4f0c...`
-sha256_of_file() {
-    sha256sum "${1}" | awk '{print $1;}'
+get_version_from_manifest() {
+    local -r artifact_path="${1}"
+    local -r manifest_path="${artifact_path}.manifest"
+    jq -r ".version" "${manifest_path}"
```
christophermaier, line 26: This whole function should be in the `manifest.sh` Bash library.

Ideally, nothing outside of that library file would need to be aware of the internal structure of the manifest file, making future evolution much easier.

firecracker/kernel/build.sh
```diff
@@ -45,3 +47,15 @@ tar \
...
+########################################
+source .buildkite/scripts/lib/manifest.sh
+INPUTS_SHA="$(sha256_of_dir firecracker/kernel)"
+readonly INPUTS_SHA_SHORT="${INPUTS_SHA:0:16}" # cloudsmith version field must be under 128 chars
+readonly MANIFEST_PATH="${DISTRIBUTION}.manifest"
+readonly VERSION="firecracker-${FIRECRACKER_RELEASE}-kernel-${KERNEL_VERSION}-input-sha256-${INPUTS_SHA_SHORT}"
```
christophermaier, line 60: 🤔 This version seems rather unwieldy, and includes extra information that is really just an "implementation detail" to get around the fact that we don't have a deterministic build.

The fact that we're also making a JSON object to just store this string seems overkill, and a bit limiting.

My suggestion would be to keep the version as it is currently (`firecracker-${FIRECRACKER_RELEASE}-kernel-${KERNEL_VERSION}`), and also keep the full-length SHA256 "input checksum". The JSON would look like this:

```json
{
    "version": "firecracker-XXX-kernel-YYY",
    "input-sha256": "abcdef01234656789..."
}
```
(Ideally, if this "manifest" is going to stay around long-term and also be uploaded to Cloudsmith as an artifact in and of itself, I'd break out the Firecracker version and the kernel version into their own dedicated fields, but that's a conversation for another time. It would also need a bit more though to harmonize with the usage for the rootfs image.)

Then, when you upload, you can have access to all the raw information and can do with it what you please. In particular, I think expressing this whole input checksum thing as a _tag_ would be a bit better and more flexible of a structure. You could just extract the checksum from the file and then add it as a tag on the artifact like `input-sha256:abcdef0123456799...`. Then, your query can include that tag as a criterion (as well as any additional tags we may see fit to add in the future). I'm not sure if there's a length limit on tags as there appears to be on versions, but at that point, you can truncate the checksum to whatever length you need to, but you're not truncating it for any other future uses (because the full-length checksum is still available in the file).

firecracker/rootfs/build.sh
```diff
@@ -0,0 +1,27 @@
...
+# (This may be replaced with Pants-based file diffs in the future.)
+########################################
+source .buildkite/scripts/lib/manifest.sh
+source .buildkite/scripts/lib/version.sh
+INPUTS_SHA="$(sha256_of_dir firecracker/rootfs)"
+readonly INPUTS_SHA_SHORT="${INPUTS_SHA:0:16}" # cloudsmith version field must be under 128 chars
+VERSION="$(timestamp_and_sha_version)-input-sha256-${INPUTS_SHA_SHORT}"
+readonly VERSION
+readonly MANIFEST_PATH="${GRAPL_ROOT}/dist/${IMAGE_NAME}.tar.gz.manifest"
+manifest_contents "${INPUTS_SHA}" "${VERSION}" > "${MANIFEST_PATH}"
```
christophermaier, line 27: `manifest_contents` only operates on the first argument, so `${VERSION}` here is being silently dropped.

firecracker/rootfs/build.sh
```diff
@@ -0,0 +1,27 @@
...
+########################################
+# Write a manifest file containing the version.
+# The version should dedupe based on input file checksums.
+# (This may be replaced with Pants-based file diffs in the future.)
+########################################
+source .buildkite/scripts/lib/manifest.sh
+source .buildkite/scripts/lib/version.sh
+INPUTS_SHA="$(sha256_of_dir firecracker/rootfs)"
+readonly INPUTS_SHA_SHORT="${INPUTS_SHA:0:16}" # cloudsmith version field must be under 128 chars
+VERSION="$(timestamp_and_sha_version)-input-sha256-${INPUTS_SHA_SHORT}"
```
christophermaier, line 24: Same comments re: a tag for `input-sha256` apply here.

firecracker/kernel/build.sh
wimax-grapl: Unfortunately you cannot have two raw packages of the same version but different tags. The tag is a nonstarter here.

https://grapl-internal.slack.com/archives/C0174A8QV2S/p1647882336138319

firecracker/kernel/build.sh
```diff
@@ -13,21 +13,23 @@ readonly REPOSITORY_ROOT
...
-########################################
-# Generate the kernel.
-# Based on https://github.com/firecracker-microvm/firecracker/blob/main/docs/rootfs-and-kernel-setup.md
-########################################
-./tools/devtool build_kernel --config "${KERNEL_CONFIG}" --nproc 8
+    ########################################
+    # Generate the kernel.
+    # Based on https://github.com/firecracker-microvm/firecracker/blob/main/docs/rootfs-and-kernel-setup.md
+    ########################################
+    ./tools/devtool build_kernel --config "${KERNEL_CONFIG}" --nproc 8
```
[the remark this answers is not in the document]
wimax-grapl: Mind if I do this in a followup PR? 🙏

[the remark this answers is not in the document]
christophermaier: Of course not 👍

.buildkite/scripts/lib/manifest.sh
wimax-grapl: done

.buildkite/scripts/lib/manifest.sh
wimax-grapl: Changed it to 'artifact-metadata.' And yes, I correctly anticipated additional content in this manifest!

firecracker/kernel/build.sh
```diff
@@ -13,35 +13,54 @@ readonly REPOSITORY_ROOT
...
+source .buildkite/scripts/lib/artifact_metadata.sh
+ARTIFACT_METADATA_PATH="$(artifact_metadata_path "${ARTIFACT_PATH}")"
+readonly ARTIFACT_METADATA_PATH
+
+source .buildkite/scripts/lib/version.sh
+# Note: in Cloudsmith, versions must be len <=128, hence we use a short git sha.
+VERSION="firecracker-${FIRECRACKER_RELEASE}-kernel-${KERNEL_VERSION}-$(sha_version)"
+readonly VERSION
+
+TAG="inputsha-$(sha256_of_dir firecracker/kernel)"
```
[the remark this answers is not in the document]
christophermaier: Nit: I think `input-sha256:${THE_SHA256}` would be a better format:

- it is a bit more readable (to me, at least: hyphens rather than all the words smooshed together)
- the `:` suggests a key-value pair, whereas the `-` doesn't
- "input-sha256" tells you exactly what type of checksum it is, whereas "input-sha" leaves you guessing.

.buildkite/scripts/build_and_upload_firecracker_packages.sh
```diff
@@ -58,24 +60,27 @@ for artifact_path in "${PACKAGES[@]}"; do
...
 
-    echo "Checking if a version ${version} exists upstream..."
-    if ! present_upstream "${artifact_name}" "${version}"; then
+    echo "Checking if a package with tag ${tag} exists upstream..."
+    if ! present_upstream "${artifact_name}" "${tag}"; then
         echo "Package not present upstream; will promote '${artifact_name}'"
         new_packages+=("${artifact_path}")
     else
-        echo "Package with this SHA exists upstream; no change needed"
+        echo "Package with this tag exists upstream; no change needed"
```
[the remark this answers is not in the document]
wimax-grapl: fair. can revert

.buildkite/scripts/build_and_upload_firecracker_packages.sh
```diff
@@ -61,10 +64,10 @@ for artifact_path in "${PACKAGES[@]}"; do
     artifact_name=$(basename "${artifact_path}")
     echo "--- :cloudsmith: Should we update '${artifact_name}' in '${UPSTREAM_REGISTRY}'?"
     version="$(get_version_from_artifact_metadata "${artifact_path}")"
-    tag="$(get_tag_from_artifact_metadata "${artifact_path}")"
+    input_sha256="$(get_input_sha256_from_artifact_metadata "${artifact_path}")"
 
-    echo "Checking if a package with tag ${tag} exists upstream..."
-    if ! present_upstream "${artifact_name}" "${tag}"; then
+    echo "Checking if a package generated with the same inputs exists upstream. SHA: ${input_sha256}"
+    if ! present_upstream "${artifact_name}" "${input_sha256}"; then
```
christophermaier, line 70: If I'm reading correctly, I think the tag you add and search for is _just_ the checksum, and not a labeled checksum like `input-sha256:<CHECKSUM>`. If that's correct, I really think we need to include an indicator of what the tag is; otherwise when we're (say) browsing through the contents of the repository, we're just going to have a context-free string of randomness. It'll be even worse if at some point in the future we tag with some other kind of checksum.

wimax-grapl:
wait thats a prefix

wimax-grapl:
(also for the record it didn't like me having a `:` in the tag)

christophermaier:
> (also for the record it didn't like me having a `:` in the tag)

:sob::sob::sob:

christophermaier (approved):
Looks good; thanks for bearing with me through the review :bow:

PR #1570, opened 2022-03-25
Update `buf` to version 1.2.1, introducing protobuf formatting

christophermaier:
`buf` 1.2.1 now has formatting abilities!

### DIFF
--- a/docker-compose.check.yml
+++ b/docker-compose.check.yml
@@ -62,13 +62,19 @@ services:
     image: bufbuild/buf:1.2.1
     command: ["format", "--verbose", "--write"]
     environment:
-      # buf tries to write to `~/.cache`, which turns out to be
-      # `/.cache` when running as a non-root user (as we need to do to
-      # write formatted files back out owned by the workstation
-      # user). We won't have permission to write to `/.cache`, so we
-      # need to change our `HOME` directory to somewhere we *do* have
-      # write permissions. This will result in a `.cache` directory in
-      # the repository root, but it will actually allow this to
+      # Since we will be writing files back to our workstation, they
+      # must be owned by our workstation user. This means we will run
+      # the container with `--user=$(id --user):$(id --group)`, which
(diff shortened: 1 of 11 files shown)


docker-compose.check.yml
```diff
@@ -46,6 +46,35 @@ services:
...
+    volumes:
+      - *read-only-workdir
+
+  # Use buf to rewrite our protobuf files with the proper formatting.
+  buf-format:
+    image: bufbuild/buf:1.2.1
+    command: ["format", "--verbose", "--write"]
+    environment:
+      # buf tries to write to `~/.cache`, which turns out to be
+      # `/.cache` when running as a non-root user (as we need to do to
```
jgrillo-grapl, line 66: "non-root user" or "root user"? If "non-root user" then why does it think `$HOME` is `/`?

jgrillo-grapl (approved):
LGTM

docker-compose.check.yml
christophermaier: Well, "it depends". In this particular case, we're entering the container with the user/group ID that we have on our workstations, which generally works out to `1000:1000`. The `bufbuild/buf` container doesn't actually _have_ a user with that ID, so we're effectively an anonymous user. I _think_ (but am having trouble finding supporting evidence now) that in such a scenario, `/` effectively becomes your home directory... but you won't be able to write to it.

christophermaier: I should probably update that comment, because this does sound confusing as it is written.

christophermaier: Well, I guess this is supporting evidence enough:

```
docker run -it --entrypoint=sh --user=1000:1000 bufbuild/buf:1.2.1 
/ $ whoami
whoami: unknown uid 1000
/ $ echo $HOME
/
/ $ 
```

jgrillo-grapl: interesting

PR #1571, opened 2022-03-25
Avoid glob import

colin-grapl:
None, this is a one line fix.

Avoids generating a glob import, which is unnecessary. Imports the types it needs explicitly instead.

I ran flake8 and mypy on the generated code.


jgrillo-grapl (approved):
is there a test somewhere that exercises any of this?

PR #1582, opened 2022-03-28
Local grapl consul dns

ghost:
https://github.com/grapl-security/issue-tracker/issues/887

This sets up consul dns on port 53 inside nomad containers. ie from within a container you can now dig redis.service.consul and get an answer.

Exec into a running nomad container
apt-get update && apt-get Install dnsutils
dig redis.service.consul. ANY
You'll get a reply

### DIFF
--- a/nomad/local/grapl-local-infra.nomad
+++ b/nomad/local/grapl-local-infra.nomad
@@ -428,6 +428,57 @@ job "grapl-local-infra" {
     }
   }
 
+  group "dnsmasq" {
+    network {
+      mode = "bridge"
+      port "dns" {
+        static = 53
+        to     = 53
+      }
+    }
+
+
+    task "dnsmasq" {
+      driver = "docker"
+
+      config {
+        image      = "andyshinn/dnsmasq"
+        force_pull = true
+        ports      = ["dns"]
+        args = [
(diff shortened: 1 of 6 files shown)


nomad/local/grapl-local-infra.nomad
```diff
@@ -428,6 +428,57 @@ job "grapl-local-infra" {
...
+        to     = 53
+      }
+    }
+
+
+    task "dnsmasq" {
+      driver = "docker"
+
+      config {
+        image      = "andyshinn/dnsmasq"
```
wimax-grapl, line 445: https://hub.docker.com/r/andyshinn/dnsmasq/ says we should use a different one

twunderlich-grapl: I've fixed this, thanks for pointing this out

nomad/local/nomad-agent-conf.nomad
```diff
@@ -15,6 +15,11 @@ plugin "docker" {
       # Required for the bind mount for docker.sock
       enabled = true
     }
+
+    # We need net_admin for dnsmasq to work. Everything else should be default
```
christophermaier, line 19: Out of curiosity, do you have a link to documentation or code that says what the default capabilities are?

christophermaier (approved):

PR #1578, opened 2022-03-28
Try switching the "build firecracker packages" to packer queue

wimax-grapl:
https://buildkite.com/grapl/grapl-merge/builds/573#a57b6130-9723-42ce-be57-a4cb8eb97432 failed due to
```
0.4s
--
  | ./firecracker/rootfs/build.sh
  | Installed plugin github.com/hashicorp/amazon v1.0.8 in "/var/lib/buildkite-agent/.config/packer/plugins/github.com/hashicorp/amazon/packer-plugin-amazon_v1.0.8_x5.0_linux_amd64"
  | Error: Datasource.Execute failed: Error querying AMI: UnauthorizedOperation: You are not authorized to perform this operation.
  | status code: 403, request id: f0695d2a-643f-4eba-9977-318074592a73
  |  
  | on firecracker/rootfs/build-rootfs.pkr.hcl line 80:
  | (source code not available)
```

Makes sense, the "docker" queue wasn't the right tool for this ("packer" has more AWS perms). 
I'm reasonably sure we do not need the "docker" queue for building the rootfs-inside-docker.

PR #1585, opened 2022-03-29
Split up 'build_and_upload_firecracker_packages' into two build steps and one upload step

wimax-grapl:
https://github.com/grapl-security/grapl/pull/1578 didn't work, so this is a more robust solution.

- Previously we were building and uploading our FIrecracker packages all in one Buildkite step.
- Turns out, that's really finicky, since our different agents have different keys granted to them.
- So now we split up the building and the uploading into different steps.

hah, good joke

### DIFF
--- a/.buildkite/pipeline.merge.yml
+++ b/.buildkite/pipeline.merge.yml
@@ -25,26 +25,52 @@ steps:
     artifact_paths:
       - "*.grapl-artifacts.json"
 
+  # Artifacts consumed by "Upload Firecracker Packages"
+  - label: ":firecracker: Build Kernel"
+    key: "build_firecracker_kernel"
+    command:
+      - "make dist/firecracker_kernel.tar.gz"
+    artifact_paths: &build_firecracker_kernel_artifacts
+      - "dist/firecracker_kernel.tar.gz"
+      - "dist/firecracker_kernel.tar.gz.artifact-metadata.json"
+    agents:
+      queue: "beefy" # TODO: or "docker"?
+
+  # Artifacts consumed by "Upload Firecracker Packages"
+  - label: ":firecracker: Build RootFS"
+    key: "build_firecracker_rootfs"
+    command:
+      - "make dist/firecracker_rootfs.tar.gz"
+    artifact_paths: &build_firecracker_rootfs_artifacts
+      - "dist/firecracker_rootfs.tar.gz"
+      - "dist/firecracker_rootfs.tar.gz.artifact-metadata.json"
+    agents:
+      queue: "packer"
+
   - label: ":firecracker: Upload Firecracker packages"
+    depends_on:
+      - build_firecracker_kernel
+      - build_firecracker_rootfs
     command:
       # Note: these packages are uploaded to the `raw`
       # repository. If they all get built successfully, then we'll
       # promote them as a group later.
       - ".buildkite/scripts/upload_firecracker_packages.sh"
     plugins:
+      - artifacts#v1.4.0:
+          download:
+            <<: *build_firecracker_kernel_artifacts
+            <<: *build_firecracker_rootfs_artifacts
       - grapl-security/vault-login#v0.1.2
       - grapl-security/vault-env#v0.1.0:
           secrets:
             # We need access to the CLOUDSMITH_API_KEY so we can upload
             # to Cloudsmith.
             - CLOUDSMITH_API_KEY
-            - grapl/TOOLCHAIN_AUTH_TOKEN
+            - grapl/TOOLCHAIN_AUTH_TOKEN # TODO: Is this needed? I don't think so
     agents:
-      # This builds both the kernel (in docker) and the rootfs (in packer).
-      # Of those two possible queues to use, 'packer' is the winner; the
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.merge.yml
```diff
@@ -25,28 +25,53 @@ steps:
...
+    steps:
+      # Artifacts consumed by "Upload Firecracker Packages"
+      - label: ":firecracker: Build Kernel"
+        command:
+          - "make dist/firecracker_kernel.tar.gz"
+        artifact_paths: &build_firecracker_kernel_artifacts
+          - "dist/firecracker_kernel.tar.gz"
+          - "dist/firecracker_kernel.tar.gz.artifact-metadata.json"
+        agents:
+          queue: "beefy" # TODO: or "docker"? nothing else uses beefy in merge
```
wimax-grapl, line 38: resolve TODO

.buildkite/pipeline.merge.yml
```diff
@@ -25,28 +25,53 @@ steps:
...
+              download:
+                <<: *build_firecracker_kernel_artifacts
+                <<: *build_firecracker_rootfs_artifacts
+          - grapl-security/vault-login#v0.1.2
+          - grapl-security/vault-env#v0.1.0:
+              secrets:
+                # We need access to the CLOUDSMITH_API_KEY so we can upload
+                # to Cloudsmith.
+                - CLOUDSMITH_API_KEY
+                - grapl/TOOLCHAIN_AUTH_TOKEN # TODO: Is this needed? I don't think so
```
wimax-grapl, line 69: resolve TODO

.buildkite/scripts/lib/artifact_metadata.sh
```diff
@@ -19,8 +19,8 @@ set -euo pipefail
 #
 # $ find . -type f | sort | xargs sha256sum
 # 123veryLongSha  ./BUILD
-# 456veryLongSha  ./build_and_upload_firecracker_packages.sh
-# 789veryLongSha  ./build_and_upload_images.sh
+# 456veryLongSha  ./upload_firecracker_packages.sh
+# 789veryLongSha  ./upload_images.sh
```
wimax-grapl, line 23: this was an unintentional replace, but since it's just an example it doesn't matter that much

.buildkite/scripts/upload_firecracker_packages.sh
```diff
@@ -61,9 +61,6 @@ readonly PACKAGES=(
 # This is the list of packages that actually have different shas
 new_packages=()
 
-echo "--- Building packages"
```
christophermaier, line 64: I'd really like to comment on the file as a whole, but this is what Github gives me....

At this point, this script isn't so much about uploading Firecracker packages as it is about uploading raw artifacts to Cloudsmith with some "input files checksum" logic to work around promotion issues. If we could somehow refine the interface for this script to better highlight that, it would be nice.

In particular, it's now a little odd that we have the Firecracker package paths hard-coded in this script, but also have them very explicitly mentioned in the pipeline file (my comments about those aside); I would expect to either have the paths in the pipeline file and then pass them into this script, or simply assume we're uploading all of `dist` in the pipeline file, and then hard-coding the file names here, rather than hard-coding them in both places, but without an obvious link between them.

The fact that the presence of the artifact metadata JSON files is implicitly assumed is also a little surprising, now that they're not actually generated in the course of running this script.

We could probably address all of this by passing in a path to a directory (e.g., "dist"), and maybe an extension (".tar.gz"); the implication being "we're going to upload all the .tar.gz files in dist to Cloudsmith; we're also assuming that each .tar.gz file also has a corresponding artifact metadata file". This could also be spelled out in a top-level comment in this script (that also needs to be updated anyway, since it talks about actually making the artifacts in this file).

I think that would really focus what the core work of this script is, and also make it easier to reuse for other "raw" packages in the future, should that be required.

.buildkite/pipeline.merge.yml
```diff
@@ -25,28 +25,53 @@ steps:
...
+      - wait
+
+      - label: ":firecracker: Upload Firecracker packages"
+        command:
+          # Note: these packages are uploaded to the `raw`
+          # repository. If they all get built successfully, then we'll
+          # promote them as a group later.
+          - ".buildkite/scripts/upload_firecracker_packages.sh"
+        plugins:
+          - artifacts#v1.4.0:
```
christophermaier, line 59: Current release of this plugin is [1.5.0](https://github.com/buildkite-plugins/artifacts-buildkite-plugin/releases/tag/v1.5.0), btw.

christophermaier (approved):
Looks good, with a few suggestions.

.buildkite/pipeline.merge.yml
wimax-grapl: yes, but I would rather keep it in sync with the rest of the file so it can be e.g. shepherd'ed

.buildkite/scripts/upload_firecracker_packages.sh
wimax-grapl: I agree with all of the above, and I'll try to address that tonight

PR #1586, opened 2022-03-29
Firecracker Rootfs asset: upload to S3, and pipe all the way through to plugin.nomad

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/742

Relates to https://github.com/grapl-security/issue-tracker/issues/857

Upload the Firecracker rootfs asset to S3 in both prod and local (this decision can be re-litigated later; see the above 857 issue).
This is the same thing we currently do for the kernel.

~then~ we pipe the s3url all the way from:
Pulumi -> grapl-core.nomad -> plugin-registry -> plugin.nomad

make test-integration
the plugin task group still fails because it tries to download the artifact for plugin, but that's something I'll solve in a later PR.

### DIFF
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -1196,16 +1196,16 @@ job "grapl-core" {
       }
 
       env {
-        AWS_REGION                       = var.aws_region
-        NOMAD_SERVICE_ADDRESS            = "${attr.unique.network.ip-address}:4646"
-        PLUGIN_REGISTRY_BIND_ADDRESS     = "0.0.0.0:${NOMAD_PORT_plugin-registry-port}"
-        PLUGIN_REGISTRY_DB_HOSTNAME      = var.plugin_registry_db_hostname
-        PLUGIN_REGISTRY_DB_PASSWORD      = var.plugin_registry_db_password
-        PLUGIN_REGISTRY_DB_PORT          = var.plugin_registry_db_port
-        PLUGIN_REGISTRY_DB_USERNAME      = var.plugin_registry_db_username
-        PLUGIN_BOOTSTRAP_CONTAINER_IMAGE = var.container_images["plugin-bootstrap"]
-        KERNEL_ARTIFACT_URL              = var.plugin_registry_kernel_artifact_url
-        ROOTFS_ARTIFACT_URL              = var.plugin_registry_rootfs_artifact_url
+        AWS_REGION                          = var.aws_region
+        NOMAD_SERVICE_ADDRESS               = "${attr.unique.network.ip-address}:4646"
+        PLUGIN_REGISTRY_BIND_ADDRESS        = "0.0.0.0:${NOMAD_PORT_plugin-registry-port}"
+        PLUGIN_REGISTRY_DB_HOSTNAME         = var.plugin_registry_db_hostname
+        PLUGIN_REGISTRY_DB_PASSWORD         = var.plugin_registry_db_password
+        PLUGIN_REGISTRY_DB_PORT             = var.plugin_registry_db_port
+        PLUGIN_REGISTRY_DB_USERNAME         = var.plugin_registry_db_username
+        PLUGIN_BOOTSTRAP_CONTAINER_IMAGE    = var.container_images["plugin-bootstrap"]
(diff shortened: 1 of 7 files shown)


nomad/grapl-core.nomad
```diff
@@ -1200,6 +1205,7 @@ job "grapl-core" {
         PLUGIN_REGISTRY_DB_USERNAME      = var.plugin_registry_db_username
         PLUGIN_BOOTSTRAP_CONTAINER_IMAGE = var.container_images["plugin-bootstrap"]
         KERNEL_ARTIFACT_URL              = var.plugin_registry_kernel_artifact_url
+        ROOTFS_ARTIFACT_URL              = var.plugin_registry_rootfs_artifact_url
```
wimax-grapl, line 1208: I took an executive decision to stop prepending PLUGIN_BOOTSTRAP_ at this level; feel free to say 'i don't like that'. 
But especially with our structopts it feels really weird in plugin-registry code to reference `config.plugin_registry_kernel_artifact_url` when you already _know_ you're in plugin_registry land.

src/rust/plugin-registry/src/server/deploy_plugin.rs
```diff
@@ -38,24 +38,15 @@ pub async fn deploy_plugin(
     let job_name = "grapl-plugin"; // Matches what's in `plugin.nomad`
     let job = {
         let job_file_hcl = static_files::PLUGIN_JOB;
-        // TODO: Deprecate this in next PR and replace with passed-in env variable
-        // ~ wimax Feb 2022
```
wimax-grapl, line 42: did it!!!

src/rust/plugin-registry/src/server/service.rs
```diff
@@ -129,9 +129,10 @@ pub struct PluginRegistryServiceConfig {
     pub plugin_bootstrap_container_image: String,
     #[structopt(env)]
     pub plugin_execution_container_image: String,
-    // TODO in my followup PR ~ wimax Feb 2022
-    // Leaving this as a TODO because it requires a larger refactor
-    // kernel_artifact_url: String,
```
wimax-grapl, line 134: that larger refactor was https://github.com/grapl-security/grapl/pull/1583

nomad/grapl-core.nomad
christophermaier: I think there are two related but distinct concerns here.

When you're in the code of the plugin registry service, I agree that constantly typing things like `config.plugin_registry_kernel_artifact_url` is a downer. You're really _doubly_ namespaced, in a way: you're in the plugin registry crate, and you're also dealing with a type named `PluginRegistryServiceConfig`. So yeah, just call that `config.kernel_artifact_url` and be done with it; no need to add another layer of namespacing.

Environment variables are slightly different, though, since they really don't _have_ any kind of native namespacing method, apart from naming conventions. It can be useful to have all the environment variables for configuring a service to be clearly named for that service. It may just be a holdover from pre-cloud, pre-container days when multiple services could be on the same machine, and you wanted to ensure that values intended for one service didn't accidentally make their way into another service (just which service is supposed to get `DB_USERNAME`, for instance?). Be that as it may, I (personally) still find it useful from an organizational / documentation perspective to name environment variables in this way.

So, my two cents is: keep the `PLUGIN_REGISTRY_` prefix on the environment variables, but ditch it in the Rust configuration code.

christophermaier (approved):

pulumi/grapl/Pulumi.local-grapl.yaml
```diff
@@ -29,6 +29,7 @@ config:
   grapl:cloudsmith-repository-name: grapl/testing
   grapl:artifacts:
     "firecracker_kernel.tar.gz": "firecracker-v1.0.0-kernel-4.14.174"
+    "firecracker_rootfs.tar.gz": "20220329205538-5ba93c6"
```
christophermaier, line 32: Do we have a way to keep this up-to-date, or is this just manual at this point?

christophermaier: Yeah, I was just asking for general knowledge purposes; I'm not suggesting any additional work for this PR.

PR #1595, opened 2022-04-05
Remove ETH0 env var that's no longer used

ghost:
[grapl-security/issue-tracker#887](https://github.com/grapl-security/issue-tracker/issues/887)

This removes the ETH0 env var since we're now using consul dns so we don't need to pass an IP address around.

make up


christophermaier (approved):
🔥 :fire: :fire:

PR #1597, opened 2022-04-06
Fix dgraph healthchecks with multiple dgraph nodes

ghost:
N/A this is a bug fix

This fixes health check failures when we set either dgraph_replicas or dgraph_shards to more than 1. This is due to hardcoded ports in the health checks.
We're not currently using multiple nodes, but this is a step in making it possible.

change dgraph_replicas or dgraph_shards to more than 1. dgraph-alpha-1 and dgraph-zero-1 will not fail due to health checks,

### DIFF
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -66,7 +66,10 @@ variable "analyzer_executor_queue" {
 variable "dgraph_replicas" {
   type    = number
   default = 1
-  # This value must be odd. Otherwise dgraph_zero will exit
+  validation {
+    condition     = var.dgraph_replicas % 2 == 1
+    error_message = "This value must be odd. Otherwise dgraph_zero will exit."
+  }
 }
 
 variable "dgraph_shards" {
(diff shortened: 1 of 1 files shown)


nomad/grapl-core.nomad
```diff
@@ -66,6 +66,7 @@ variable "analyzer_executor_queue" {
 variable "dgraph_replicas" {
   type    = number
   default = 1
+  # This value must be odd. Otherwise dgraph_zero will exit
```
wimax-grapl, line 69: not a must do but just saying its possible
```
validation {
    condition     = var.dgraph_replicas % 2 == 1
    error_message = "This value must be odd. Otherwise dgraph_zero will exit"
}
```

wimax-grapl (approved):

nomad/grapl-core.nomad
christophermaier: oooooh

christophermaier (approved):

nomad/grapl-core.nomad
twunderlich-grapl: I knew that terraform supported this but I couldn't find any mention of it in the nomad docs. I've now tested it and it does work. 
I may try and put up a PR to the nomad docs for this

PR #1598, opened 2022-04-06
Add clean command for dgraph images and volumes

ghost:
As we found out today, our clean command prior to this PR didn't affect anything dgraph-related. This lets us clean dgraph docker images and volumes.

In the future we should add labels to volumes

Ran the make changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,10 @@ NONROOT_DOCKER_COMPOSE_CHECK := ${DOCKER_COMPOSE_CHECK} --user=${COMPOSE_USER}
 #
 # This is set in docker-bake.hcl
 DOCKER_FILTER_LABEL := org.opencontainers.image.vendor="Grapl, Inc."
+# We pull some vendor containers directly
+DOCKER_DGRAPH_FILTER_LABEL := maintainer="Dgraph Labs <contact@dgraph.io>"
+# We don't set labels on most volumes. This is one of the few that we do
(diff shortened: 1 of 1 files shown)


colin-grapl (approved):

Makefile
```diff
@@ -565,6 +571,18 @@ clean-docker-images: ## Remove all Grapl images
 		--quiet \
 	| xargs --no-run-if-empty docker rmi --force
 
+clean-docker-volumes: ## Remove all Grapl volumes
```
christophermaier, line 574: This isn't _all_ our volumes (it doesn't touch the ones for the build containers, for instance), just the ones generated in the course of running `docker-compose` (judging by the label), right? If so, it might be good to reflect that in the target name and the comment (as well as the comment on `DOCKER_VOLUME_FILTER_LABEL`, too).

Makefile
```diff
@@ -565,6 +571,18 @@ clean-docker-images: ## Remove all Grapl images
...
+	docker volume prune \
+		--filter=label=$(DOCKER_VOLUME_FILTER_LABEL) \
+		--force
+
+clean-docker-dgraph: ## Remove dgraph images, containers and volumes
```
christophermaier, line 579: This command only looks like it removes images, not containers or volumes.

Makefile
twunderlich-grapl: I've updated the comments, however since only the dgraph_export volume is labeled at the moment, I'm pondering whether it makes sense to drop the filter and just blow away all volumes. Thoughts?

PR #1600, opened 2022-04-06
Nomad Local Infra - deploy grapl-local-infra and observability simultaneously - saves 18s for make up 

wimax-grapl:
This is just a dev loop niceness upgrade - instead of deploying these two jobs serially, we deploy them in parallel and then monitor their success serially. 

After 3 runs, averages of `make down && time make start-nomad-detach`
Before: 1m1s
After: 43s

### DIFF
--- a/nomad/local/nomad_run_local_infra.sh
+++ b/nomad/local/nomad_run_local_infra.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 THIS_DIR=$(dirname "${BASH_SOURCE[0]}")
 GRAPL_ROOT="$(git rev-parse --show-toplevel)"
-NOMAD_FILE="${GRAPL_ROOT}/nomad/local/grapl-local-infra.nomad"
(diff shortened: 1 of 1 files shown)


nomad/local/nomad_run_local_infra.sh
```diff
@@ -20,33 +20,38 @@ echo "Deploying Nomad local infrastructure"
...
 # fyi: Exit code 1 means "Allocations created or destroyed" and is what we want
 while [[ $(attemptPlan) -ne 1 ]]; do
     echo "Waiting to be able to do a nomad-job-run"
     sleep 1
 done
 
-nomad job run -verbose "${NOMAD_VARS[@]}" "${NOMAD_FILE}"
+# Kick off all your nomad jobs in parallel
+nomad_evals=()
+nomad_evals+=("$(nomad_run "${NOMAD_VARS[@]}" "${LOCAL_INFRA_NOMAD_FILE}")")
```
twunderlich-grapl, line 45: Oh this is such a nicer pattern. Thanks for doing this

PR #1602, opened 2022-04-06
Remove a bunch of JS deps

wimax-grapl:
- Remove a bunch of JS deps
- Also move a bunch of them to devDependencies


wimax-grapl:
update: this is so much worse than I thought lol, nevermind

PR #1604, opened 2022-04-07
Switch from React-Force-Graph to React-Force-Graph-2D. Report the source-map-explorer results to CI. 

wimax-grapl:
None

- Instead of the full react-force-graph which contains like WebGL WebVR lots of 3D useless stuff - just include the 2D version. Built JS goes from 15MB to 5MB.
- Use `source-map-explorer` in CI to generate a report of what's using how much space. It'll be uploaded as an artifact.

A note: We just upload it as an artifact in e2e/integration-test CI; there's no better spot for it currently. 

Tested locally with make build-code; let's see how it behaves in CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,8 +38,6 @@ src/rust/grapl-web-ui/frontend/*
 nomad/local/mount-grapl-root-as-volume.nomad
 
 # Distribution / packaging
-/src/js/engagement_view/build/
-/src/js/engagement_view/source_map_explorer_result.html
 *.zip
 /dist/
 
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,8 @@ src/rust/grapl-web-ui/frontend/*
 nomad/local/mount-grapl-root-as-volume.nomad
 
 # Distribution / packaging
-build/
+/src/js/engagement_view/build/
+/src/js/engagement_view/source_map_explorer_result.html
 *.zip
 /dist/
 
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -38,7 +38,8 @@ src/rust/grapl-web-ui/frontend/*
 nomad/local/mount-grapl-root-as-volume.nomad
 
 # Distribution / packaging
-build/
+/src/js/engagement_view/build/
+/src/js/engagement_view/source_map_explorer_result.html
```
christophermaier, line 42: Might be better to put these into `src/js/engagement_view/.gitignore` for now, per the comment at the top of that file.

Makefile
```diff
@@ -156,9 +156,12 @@ build-e2e-pex-files:
 build-engagement-view: ## Build website assets to include in grapl-web-ui
 	@echo "--- Building the engagement view"
 	$(ENGAGEMENT_VIEW_MAKE) build-code
+	TARGET_FRONTEND_DIR="${PWD}/src/rust/grapl-web-ui/frontend/"
```
christophermaier, line 159: The `${PWD}` stuff isn't really needed; `src/rust/grapl-web-ui/frontend` is sufficient.

Also, the fact that there's a trailing `/` here means that the command on the next line is actually 
```shell
rm -rf ${PWD}/src/rust/grapl-web-ui/frontend//*
```
(for this reason, it's generally better to _not_ include trailing slashes in directory path variables)

christophermaier (commented):
Could you file an issue to revisit the placement of this `source-map-explorer` run in CI, and tag it with the `internal-engineering` label? It's fine for now, but we may want to rework this in the future.

andrea-grapl (approved):

christophermaier (approved):

PR #1606, opened 2022-04-07
Remove dnsmasq from local Grapl Nomad infra

inickles-grapl:

This removes dnsmasq from local Grapl infra defined in Nomad. There are a few reasons to do this:
1. It's not currently being used. All services are using "localhost" and exposed ports, we aren't using service discovery locally.
2. I don't love that it's giving CAP_NET_ADMIN to containers made from an image from a seemingly random person.
3. It's causing issues for non-ChromeOS/Crostini development environments.

Initially tested locally, will rely on I tests.

### DIFF
--- a/etc/chromeos/systemd-resolve-dns.sh
+++ b/etc/chromeos/systemd-resolve-dns.sh
@@ -2,8 +2,6 @@
 
 set -euo pipefail
 
-# WARNING: This is not idempotent (yet). Only run this once
-
 # check systemd version
 SYSTEMD_VERSION=$(systemctl --version | head --lines 1 | awk '{ print $2}')
 if [[ $SYSTEMD_VERSION -lt 246 ]]; then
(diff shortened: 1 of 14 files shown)


twunderlich-grapl:
@inickles-grapl please hold off on merging this in. Because of docker stripping out 127.0.0.1 I've got to do more finagling to get this working from the host. The good news is that it probably means that the aws work I've been doing is probably broken on the same problem, so getting this working should fix both

inickles-grapl:
Latest ChromeOS update introduces systemd v247, which makes this whole thing *so* much easier. We'll just need to drop a small config in `/etc/systemd/resolved.conf.d/consul.conf`.

https://learn.hashicorp.com/tutorials/consul/dns-forwarding#systemd-resolved-setup

twunderlich-grapl (approved):
@inickles-grapl,
Can you take a look when you have the chance?
 It turns out that systemd-resolve didn't end up being simpler than the dnsmasq setup. The main issue is that containers need a way to reach out to the host dns, so they need a private IP address for the dns server.

etc/chromeos/systemd-resolve-dns.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo '`sudo bash /opt/google/cros-containers/bin/upgrade_container DEBIAN_BULLSEYE`'
+fi
+
+# Since systemd-resolved listens on 127.0.0.53, we're going to also have it listen on the docker0 bridge.
+# This will usually resolve to 172.17.0.1
+DOCKER0_BRIDGE=$(docker network inspect bridge --format='{{(index .IPAM.Config 0).Gateway}}')
+echo "DNSStubListenerExtra=$DOCKER0_BRIDGE" | sudo tee --append /etc/systemd/resolved.conf
+
+sudo systemctl enable systemd-resolved
+sudo systemctl restart systemd-resolved
```
inickles-grapl, line 22: This seems unnecessary because it's done below.

etc/chromeos/systemd-resolve-dns.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo "Systemd version must be 246+ in order to enable consul dns."
+    echo "Please update your chromeos container to Debian Bullseye by running"
+    # shellcheck disable=SC2016
+    echo '`sudo bash /opt/google/cros-containers/bin/upgrade_container DEBIAN_BULLSEYE`'
+fi
+
+# Since systemd-resolved listens on 127.0.0.53, we're going to also have it listen on the docker0 bridge.
+# This will usually resolve to 172.17.0.1
+DOCKER0_BRIDGE=$(docker network inspect bridge --format='{{(index .IPAM.Config 0).Gateway}}')
+echo "DNSStubListenerExtra=$DOCKER0_BRIDGE" | sudo tee --append /etc/systemd/resolved.conf
```
inickles-grapl, line 19: This isn't necessary, right? We just need `/etc/systemd/resolved.conf.d/consul.conf`.

Messing with the main config, and even having `/etc/systemd/resolved.conf.d/dns_servers.conf`, seems unnecessary and error prone.

inickles-grapl (commented):
@ThomasWunderlich I don't think we need stub listener, I think we just need `/etc/systemd/resolved.conf.d/consul.conf`? That's basically what I have on my other system.

Without this extra stub listener, it seems like we wouldn't need:
- LOCAL_HOST_IP
- to touch /etc/systemd/resolved.conf
- nomad var for dns server.

Is the stub needed for something?

twunderlich-grapl:
@inickles-grapl,
Every time I remove the dns stanzas inside the nomad files, consul dns from inside the containers stops working for me. I'm not sure if that means that    I've got a misconfiguration on my computer or if its failing silently for you. I'll ping you on slack to follow-up

inickles-grapl:
> @inickles-grapl, Every time I remove the dns stanzas inside the nomad files, consul dns from inside the containers stops working for me. I'm not sure if that means that I've got a misconfiguration on my computer or if its failing silently for you. I'll ping you on slack to follow-up

@ThomasWunderlich Okay yeah, looks like we still need dns stanzas and the extra stub listener. I cleaned up the setup script to make it no longer change /etc/resolv.conf, make it idempotent and keep the systemd-resolved configs to just consul.conf. By forwarding all requests on our extra stub to Consul it looks like the `/etc/systemd/resolved.conf.d/dns_servers.conf` config is redundant because Consul is already configured to do similar via the `recursors` (https://github.com/grapl-security/grapl/blob/main/nomad/local/consul-agent-conf.hcl#L22). Lemme know what you think.

twunderlich-grapl (approved):
@inickles-grapl,
Yeah I think your changes make a lot of sense. Let's merge this in

inickles-grapl:
This PR makes new tradeoffs we aren't ready for. We're closing this and making new PRs to improve the situation on main.

PR #1609, opened 2022-04-07
Firecracker CNI

ghost:
https://github.com/grapl-security/issue-tracker/issues/890

This adds CNI configs and the tc-redirect-tap CNI plugin dependency. Sadly this doesn't appear to be usable on ChromeOS due to missing kernel modules, but this will still be useful when running this on AWS or other distros.

Both bash scripts were run successfully

### DIFF
--- a/etc/chromeos/lib/installs.sh
+++ b/etc/chromeos/lib/installs.sh
@@ -281,6 +281,48 @@ download_and_install_tarball() {
     rm -Rf "${temp_dir}"
 }
 
+# Download and install a zip.
+#
+# Assumptions:
+# - URL is HTTPS
+# - URL is for a zip file
+# - The zip has all its contents at the root of the
+#   archive. Everything in the archive will be moved as-is into the
+#   destination directory.
+
+download_and_install_zip() {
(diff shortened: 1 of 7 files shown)


etc/chromeos/lib/installs.sh
```diff
@@ -322,3 +368,20 @@ install_sqlx_prepare_deps() {
     cargo install sqlx-cli --no-default-features --features postgres,rustls
     sudo apt install --yes netcat # used for `nc`
 }
+
+install_tc_redirect_tap() {
```
twunderlich-grapl, line 372: Since this requires golang, I'm not making this mandatory. 
I'm split on whether it makes sense to auto-install golang or just package this up somewhere for downloading in the future

wimax-grapl (approved):

etc/chromeos/lib/installs.sh
christophermaier: Packaging it and downloading from, say, Cloudsmith probably makes the most sense long-term.

Short-term, you could compile it in a Go container and just spit the compiled binary out.

PR #1610, opened 2022-04-08
etc/chromeos/lib/installs.sh - stop reinstalling existing stuff unless FORCE_REINSTALL=1 

wimax-grapl:
On origin/main, when this script had previously been already run, it took 8m36s *and* some confirmation prompts (for the timing I did it *as soon* as the prompt showed up on screen) to rerun

Now, when there's a previous run, it takes 35s and is zero touch.

I did not hit _every_ dependency, just the ones that take long (especially the cargo install compilations)

### DIFF
--- a/etc/chromeos/lib/installs.sh
+++ b/etc/chromeos/lib/installs.sh
@@ -27,7 +27,7 @@ should_force_reinstall() {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


christophermaier (approved):
Nice stuff

PR #1611, opened 2022-04-08
Make etc/chromeos nvm install work in gitpod

wimax-grapl:
slash, any other distro that doesn't have XDG_CONFIG_HOME set up. The existing code assumes the NVM dir will be ~/.config/nvm; we now *declare* that to be so.

Compare with https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh


twunderlich-grapl (approved):

christophermaier (approved):
FUN TIMES

PR #1612, opened 2022-04-11
resolve FE performance issues by properly using ctx.save(), ctx.resto…

andrea-grapl:
(https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/888)

Appropriately use ctx.save() and ctx.restore(). Without proper use, graph rendering is extremely slow.

Manually.

`KEEP_TEST_ENV=1 make test-e2e`

on `localhost:3000` interact with graph, there shouldn't be any lags, and the UI shouldn't freeze.

### DIFF
--- a/src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
+++ b/src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
@@ -1,28 +1,26 @@
 import React, {
     useEffect,
     useState,
-    useMemo,
     useCallback,
     useRef,
 } from "react";
-import ForceGraph2D from "react-force-graph-2d";
-import { nodeFillColor, riskOutline } from "./graphVizualization/nodeColoring";
+import {ForceGraph2D} from "react-force-graph";
+import {nodeFillColor, riskOutline} from "./graphVizualization/nodeColoring";
 import {
-    calcLinkParticleWidth,
     calcLinkColor,
     calcLinkDirectionalArrowRelPos,
 } from "./graphVizualization/linkCalcs";
-import { nodeSize } from "./graphVizualization/nodeCalcs";
-import { getLinkLabel } from "./graphLayout/labels";
-import { updateGraph } from "./graphUpdates/updateGraph";
-import { Link, VizNode, VizGraph } from "../../types/CustomTypes";
+import {nodeSize} from "./graphVizualization/nodeCalcs";
+import {getLinkLabel} from "./graphLayout/labels";
+import {updateGraph} from "./graphUpdates/updateGraph";
+import {Link, VizNode, VizGraph} from "../../types/CustomTypes";
 import {
     GraphState,
     GraphDisplayState,
     GraphDisplayProps,
 } from "../../types/GraphDisplayTypes";
 
-import { colors } from "./graphVizualization/graphColors";
+import {colors} from "./graphVizualization/graphColors";
 
 type ClickedNodeState = VizNode | null;
 
@@ -30,7 +28,7 @@ const defaultGraphDisplayState = (
     lensName: string | null
 ): GraphDisplayState => {
     return {
-        graphData: { nodes: [], links: [], index: {} },
+        graphData: {nodes: [], links: [], index: {}},
         curLensName: lensName,
     };
 };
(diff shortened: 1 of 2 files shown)


src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -39,36 +32,42 @@ const defaultClickedState = (): ClickedNodeState => {
...
+    state: any,
+    setState: any
+) {
+    if (lensName) {
+        await updateGraph(lensName, state as GraphState, setState); // state is safe cast, check that lens name is not null
+
+        console.log(
+            "setState",
+            await updateGraph(lensName, state as GraphState, setState)
+        );
```
colin-grapl, line 46: This updates the graph twice in a row. If you want to log the return value of updateGraph do `let merged = await updateGraph` and log merged.

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -39,36 +32,42 @@ const defaultClickedState = (): ClickedNodeState => {
...
     useEffect(() => {
-        // Set the initial state immediately
-        updateGraphAndSetState();
-
-        // refresh every 10 seconds
-        const interval = setInterval(updateGraphAndSetState, 10000);
+        updateGraphAndSetState(lensName, state, setState);
+        // Set the initial state immediately refresh every 5 seconds
+        const interval = setInterval(() => {
+            updateGraphAndSetState(lensName, state, setState);
```
colin-grapl, line 65: Can we do a then + catch and log a debug/error?

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -39,36 +32,42 @@ const defaultClickedState = (): ClickedNodeState => {
...
+        }, 5000);
```
colin-grapl, line 66: Can we move this to 1 second?

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -39,36 +32,42 @@ const defaultClickedState = (): ClickedNodeState => {
...
     const [state, setState] = useState(defaultGraphDisplayState(lensName));
-
-    async function updateGraphAndSetState() {
-        if (lensName) {
-            await updateGraph(lensName, state as GraphState, setState); // state is safe cast, check that lens name is not null
-        }
-    }
+    const [clickedNode, setClickedNode] = useState(defaultClickedState());
+    const [highlightNodes, setHighlightNodes] = useState(new Set());
+    const [highlightLinks, setHighlightLinks] = useState(new Set());
```
wimax-grapl, line 55: Do we have types for a "Node" or a "Link"? If so: 

`useState(new Set<Node>());` 

and such

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -122,33 +121,46 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
...
+    let clickedNodeKey = null;
+    if (clickedNode !== null) {
+        clickedNodeKey = clickedNode.id;
+    }
+
+    let hoverNodeKey = null;
+    if (hoverNode !== null) {
+        hoverNodeKey = (hoverNode as any).id;
+    }
+
```
colin-grapl, line 133: Add a comment here, something like "We only want to rerender when the id of a node changes, but we don't want to update based on any of its other attributes"

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -39,36 +32,42 @@ const defaultClickedState = (): ClickedNodeState => {
...
-
-    async function updateGraphAndSetState() {
-        if (lensName) {
-            await updateGraph(lensName, state as GraphState, setState); // state is safe cast, check that lens name is not null
-        }
-    }
+    const [clickedNode, setClickedNode] = useState(defaultClickedState());
+    const [highlightNodes, setHighlightNodes] = useState(new Set());
+    const [highlightLinks, setHighlightLinks] = useState(new Set());
+    const [hoverNode, setHoverNode] = useState(null);
```
wimax-grapl, line 56: i also worry here that hoverNode, setHoverNode are untyped since we don't know what the `null` could be. I tihnk you can do like

```
hoverNodeBackingModel: GoodType | null = null; 
const [hoverNode, setHoverNode] = useState(hoverNodeBackingModel);
```

colin-grapl (changes requested):

colin-grapl (approved):
Approving but please make those changes.

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -122,33 +121,46 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
         [setHoverNode, updateHighlight, highlightLinks, highlightNodes]
     );
 
+    let clickedNodeKey = null;
+    if (clickedNode !== null) {
+        clickedNodeKey = clickedNode.id;
+    }
```
wimax-grapl, line 127: I think this is the perfect use case for the Elvis Operator

```
let clickedNodeKey = clickedNode?.id;
```

same with below

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -122,33 +121,46 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
...
+
+    let hoverNodeKey = null;
+    if (hoverNode !== null) {
+        hoverNodeKey = (hoverNode as any).id;
```
wimax-grapl, line 131: yeah, this cast to `as any` here confirms my concerns about the typing above

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -230,36 +244,42 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
...
+            }}
             onNodeDragEnd={(node) => {
                 node.fx = node.x;
                 node.fy = node.y;
             }}
             linkColor={(link) =>
                 highlightLinks.has(link)
                     ? colors.highlightLink
-                    : calcLinkColor(link as Link, data as VizGraph)
+                    : // : "#555"
```
wimax-grapl, line 274: remove comment

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -230,36 +244,42 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
...
+                      calcLinkColor(link as Link, data as VizGraph)
```
wimax-grapl, line 275: does this imply that the above `link: any` line 197 should be a `link: Link` ? I could be wrong

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
andrea-grapl: This was for testing, sorry, removing

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
andrea-grapl: in this case, the null in useState is the initial value for the graph, the hoverNodes are typed elsewhere, so this should be okay, but thak you for being so thoughtful here.

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
andrea-grapl: I'll resolve this in a separate PR - we'll have to rework a couple of the types, making an issue, thanks for bringing this to my attention!

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
andrea-grapl: Will resolve in a separate PR - we need a few new types to distinguish types for the Graph and types that come back as raw data. These changes are more than just a few lines and don't exactly fit into the scope of this PR, but I will address this. Thanks!

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
andrea-grapl: Skipping Elvis for now - because we have people who don't know TS well sometimes working in our TS codebase, I want the code to be as clear as possible even though Elvis is very very cool.

PR #1615, opened 2022-04-12
Store DGraph data in Docker volumes, instead of leaving in container fs.

inickles-grapl:

https://github.com/grapl-security/issue-tracker/issues/900

This has Nomad use a Docker volume for storing Dgraph data, allowing it to persist across container resets, which can happen from time to time. This prevents data loss observed during local grapl development. By decoupling Dgraph storage from the execution, I no longer notice when Nomad jobs reset.

Note to reviewers: I'm looking for input on two things:
1. Is there a better way to define these volumes mounts once and refer to them in each task stanza?
2. I'm just piggy-backing off of the auto-label used by the Makefile, which should probably be given a better name (key) and the comments should be updated.

I tested similar changes in my `inickles/dropper` branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,10 +59,16 @@ NONROOT_DOCKER_COMPOSE_CHECK := ${DOCKER_COMPOSE_CHECK} --user=${COMPOSE_USER}
 DOCKER_FILTER_LABEL := org.opencontainers.image.vendor="Grapl, Inc."
 # We pull some vendor containers directly
 DOCKER_DGRAPH_FILTER_LABEL := maintainer="Dgraph Labs <contact@dgraph.io>"
-# Currently we don't label volumes explicitly. The only labeled volume is the
-# grapl-dgraph-data which is labeled automatically by docker-compose
-# TODO label volumes explicitly and then update this filter
-DOCKER_VOLUME_FILTER_LABEL := com.docker.compose.project="grapl"
+# This filter should differentiate between data volumes, such as those used for
+# DGraph (grapl-data-dgraph), and those used for caching builds
+# (grapl-engagement-view-yarn).
+#
+# Multiple filter arguments can be supplied here.
+#
+# We may want to stick to a naming convention where data volume should just
+# prefix their name to match the filter here, but that doesn't necessarily need
+# to be a strict requirement.
+DOCKER_DATA_VOLUME_FILTERS := --filter=name=grapl-data
 
 # Run a Pants goal across all Python files
 PANTS_PYTHON_FILTER := ./pants filter --target-type=python_sources,python_tests :: | xargs ./pants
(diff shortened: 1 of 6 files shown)


wimax-grapl:
> Is there a better way to define these volumes mounts once and refer to them in each task stanza?
Unless I'm missing something big, I have yet to find a way to not repeat **_stanzas_** in HCL2. However, you could do something like the following:

```
locals {
  dgraph_volume_args = {
    target = "/dgraph"
    source = "grapl-dgraph"
    # ...
  }
}
 
...later...


        mount {
          type     = "volume"
          target   = local.dgraph_volume_args["target"]
          source   = local.dgraph_volume_args["source"]
``` 
that is to say, you can deduplicate the values of the fields, but not the `volume { }` stanza itself.

I'm not necessarily sure it's worth it, but I don't hate it.
---

> I'm just piggy-backing off of the auto-label used by the Makefile, which should probably be given a better name (key) and the comments should be updated.
yeah, i'm down with you changing that around if you'd like

inickles-grapl:
Damn, looks like we can't have dots in some names, like for docker volume labels, per HCL 2. Unless I'm missing something, that's a little unfortunate, but something we can still work with.

twunderlich-grapl:
@inickles-grapl ,
Regarding how to DRY this up, Max's solution of using local variables is a good one. We could also switch to using either levant or nomad-pack (in alpha/beta) in order to use go-templates with nomad files, but I was hoping to wait until nomad-pack was stable to do that. Long-term that's the right approach

inickles-grapl:
> @inickles-grapl , Regarding how to DRY this up, Max's solution of using local variables is a good one. We could also switch to using either levant or nomad-pack (in alpha/beta) in order to use go-templates with nomad files, but I was hoping to wait until nomad-pack was stable to do that. Long-term that's the right approach

I like the idea of using something like that, but personally I think I'd prefer to have that be a different PR and keep with Max's solution for this one. How does that sound?

twunderlich-grapl (approved):

inickles-grapl:
@twunderlich-grapl ty for the approval 🙏 

@wimax-grapl I'm interested in your thoughts on the third commit I added.

wimax-grapl:
Perhaps there's something creative we can do with
https://www.pulumi.com/registry/packages/docker/api-docs/volume/

where we also set the volume as a stateful-resource-to-be-deleted. We'd have to set up that volume on every nomad-agent host, though, which gets.... complicated.... i think....?

inickles-grapl:
> Perhaps there's something creative we can do with https://www.pulumi.com/registry/packages/docker/api-docs/volume/
> 
> where we also set the volume as a stateful-resource-to-be-deleted. We'd have to set up that volume on every nomad-agent host, though, which gets.... complicated.... i think....?

Couldn't we just use that to create/define the Docker volume before Nomad tries to use it?

wimax-grapl:
Yes, -but- the problem is that - unlike on our machines, where the Nomad agent is the same computer as the one executing Pulumi - here, the pulumi-runner is in Buildkite and needs to inform every Nomad Agent machine to create (or destroy) that volume

inickles-grapl:
> Yes, -but- the problem is that - unlike on our machines, where the Nomad agent is the same computer as the one executing Pulumi - here, the pulumi-runner is in Buildkite and needs to inform every Nomad Agent machine to create (or destroy) that volume

Gotcha, makes sense.

PR #1618, opened 2022-04-13
Label dgraph_export volume

ghost:
Since docker bake doesn't support creating volumes at the moment, we
are labeling the volume inside of docker-compose. This will enable
better filtering in Make commands and/or Nomad

This adds basic labels to the dgraph_export volume. This will make it easier to filter and work with in the future.

Remove the dgraph volume, then recreate the volume via make up

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,8 +4,16 @@
 # 84xx - debugger ports (see vsc_debugger.py)
 
 version: "3.8"
+# docker-bake doesn't support volume creation, so we define and label them here
 volumes:
   dgraph_export:
+    # Labels are only applied to newly created volumes. Existing volumes will NOT have labels updated
+    labels:
+      "org.opencontainers.image.title": "dgraph_export"
+      "org.opencontainers.image.vendor": "Grapl, Inc."
+      # This label is for backwards compatibility
+      "com.docker.compose.project": "grapl"
+      maintainer: "Dgraph Labs <contact@dgraph.io>"
 
 services:
   pulumi:


wimax-grapl (approved):

docker-compose.yml
```diff
@@ -4,8 +4,16 @@
 # 84xx - debugger ports (see vsc_debugger.py)
 
 version: "3.8"
+# docker-bake doesn't support volume creation, so we define and label them here
 volumes:
   dgraph_export:
+    # Labels are only applied to newly created volumes. Existing volumes will NOT have labels updated
+    labels:
+      "org.opencontainers.image.title": "dgraph_export"
+      "org.opencontainers.image.vendor": "Grapl, Inc."
```
inickles-grapl, line 13: Note these are "image" label keys, prob not what we want for volumes.

docker-compose.yml
```diff
@@ -4,8 +4,16 @@
...
+      # This label is for backwards compatibility
+      "com.docker.compose.project": "grapl"
```
christophermaier, line 15: Backwards compatibility with what?

docker-compose.yml
```diff
@@ -4,8 +4,16 @@
 # 84xx - debugger ports (see vsc_debugger.py)
 
 version: "3.8"
+# docker-bake doesn't support volume creation, so we define and label them here
```
christophermaier, line 7: I don't think this comment is really necessary. `docker buildx bake` is about building images; volumes don't apply to that usecase.

docker-compose.yml
```diff
@@ -4,8 +4,16 @@
...
 volumes:
   dgraph_export:
+    # Labels are only applied to newly created volumes. Existing volumes will NOT have labels updated
+    labels:
+      "org.opencontainers.image.title": "dgraph_export"
+      "org.opencontainers.image.vendor": "Grapl, Inc."
+      # This label is for backwards compatibility
+      "com.docker.compose.project": "grapl"
+      maintainer: "Dgraph Labs <contact@dgraph.io>"
```
christophermaier, line 16: Why is this here? Dgraph doesn't maintain our volumes 🙃

PR #1621, opened 2022-04-13
Run an initial updateTableStateWithLenses upon loading engagements page instead of waiting 5s

wimax-grapl:
None

I was tooling around the new UI and re-encountered a tihng that's bothered me for a while. I finally decided to dive into React and learn how to fix it.

This method can also be applied to src/js/engagement_view/src/components/engagementView/EngagementView.tsx
which currently has a `renderedOnce` check

...but that I'm saving for a different PR.

checked the timings for initial graphql call in F12 chrome tools, way better with this change.

### DIFF
--- a/src/js/engagement_view/src/components/engagementView/sidebar/tables/toggleLensTable.tsx
+++ b/src/js/engagement_view/src/components/engagementView/sidebar/tables/toggleLensTable.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from "react";
+import { useEffectUponMount } from "lib/custom_react_hooks";
 
 import Button from "@material-ui/core/Button";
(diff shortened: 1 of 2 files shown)


src/js/engagement_view/src/components/engagementView/sidebar/tables/toggleLensTable.tsx
```diff
@@ -48,29 +49,27 @@ export function ToggleLensTable({ setLens }: ToggleLensTableProps) {
         setPageState(0);
     };
 
+    async function updateTableStateWithLenses() {
```
wimax-grapl, line 52: This function is exactly the same as the anonymous setInterval function it replaces (but also ported to await())

colin-grapl (approved):

PR #1627, opened 2022-04-14
Refactor a rudimentary library for shell colors out of rust/run_in_container.sh into its own library + use it in devbox

wimax-grapl:
This is strictly a readability thing. 

image

### DIFF
--- a/src/sh/BUILD
+++ b/src/sh/BUILD
@@ -1,0 +1,1 @@
+shell_sources()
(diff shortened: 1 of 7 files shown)


colin-grapl (approved):

src/sh/BUILD
```diff
@@ -0,0 +1 @@
+shell_sources()
```
wimax-grapl, line 1: notable: I just added `src/sh` for the first time

the rust dockerfile context starts in `src/` so it seemed a reasonable place to put it

christophermaier (approved):
👍

PR #1628, opened 2022-04-14
Add @dataclass to our Query and View types in analyzerlib

wimax-grapl:
Gives it automatic __repr__ and __eq__, among other things

note, i haven't actually tested the results yet, so, waiting on https://grapl-internal.slack.com/archives/C0174A8QV2S/p1649967932058359?thread_ts=1649886853.645549&cid=C0174A8QV2S

### DIFF
--- a/src/rust/grapl-graphql-codegen/src/node_type.rs
+++ b/src/rust/grapl-graphql-codegen/src/node_type.rs
@@ -166,12 +166,12 @@ impl NodeType {
             view_name=?view_name,
         );
 
-        let supertype =
+        let superclasses =
             format!("grapl_analyzerlib.nodes.entity.EntityQuery['{view_name}', '{query_name}']");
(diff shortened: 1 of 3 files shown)


src/rust/grapl-graphql-codegen/src/node_type.rs
```diff
@@ -157,26 +157,21 @@ impl NodeType {
...
         tracing::trace!(
             message="Generating Python Queryable",
             node_type=?self.type_name,
-            query_name=?q,
-            view_name=?v,
+            query_name=?query_name,
+            view_name=?view_name,
         );
 
+        let supertype =
```
wimax-grapl, line 169: in retrospect i guess the right term is superclass

jgrillo-grapl (approved):
This seems sensible at face value

PR #1629, opened 2022-04-16
Add a step to `grapl-provision` that will nuke dgraph data

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/904
Relates to https://github.com/grapl-security/grapl/pull/1615

Add a step to `grapl-provision` that will RPC to dgraph and tell it to drop all data.
With the addition of persistent volumes, we need to handle this case somehow.

Currently, it will -always- do that, but it's hooked up to an env{} flag for easy orchestration down the line.

I will try to play with these changes on top of the dgraph PR and see how it behaves.

### DIFF
--- a/src/python/Dockerfile
+++ b/src/python/Dockerfile
@@ -6,6 +6,7 @@ RUN apt-get update \
     && apt-get -y install --no-install-recommends \
         bash=5.1-2+b3 \
         libstdc++6=10.2.1-6 \
+        curl=7.74.0-1.3+deb11u1 \
     && rm -rf /var/lib/apt/lists/* \
     && adduser \
         --disabled-password \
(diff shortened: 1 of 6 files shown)


wimax-grapl:
Note that this would be awful to have in a real production setting, but I think that's something we should handle when do the migration to Scylla

src/python/Dockerfile
```diff
@@ -33,7 +34,7 @@ CMD ["./engagement-creator.pex"]
 ################################################################################
 FROM grapl-python-base AS provisioner-deploy
 COPY ./dist/provisioner.pex .
-CMD ["./provisioner.pex"]
+CMD [":"]
```
wimax-grapl, line 37: this is the bash noop command

jgrillo-grapl (approved):

PR #1631, opened 2022-04-17
Upgrade prettier + run the four incantations in parallel (13 seconds -> 7 seconds)

wimax-grapl:
1. Upgrade the version of Prettier and the Node docker image it runs on

2. Run the four `prettier` calls in parallel; join on them at the end.
The output _is_ uglier, for the record.

The 'good' successful case looks like
```
docker-compose --file=docker-compose.check.yml run --rm --user=1000:1000 prettier-lint
Creating grapl_prettier-lint_run ... done
Checking formatting...
Checking formatting...
Checking formatting...
All matched files use Prettier code style!
../../src/js/engagement_view/src/components/graphDisplay/graphLayout/vizGraphFromLensScope.tsx../../src/python/grapl-template-generator/grapl_template_generator/grapl-templates/rust-grpc-sAll matched files use Prettier code style!
../../src/rust/nomad-client-gen/docs/ScalingPolicyListStub.mdChecking formatting...
All matched files use Prettier code style!
All matched files use Prettier code style!
```

the error case looks like

image

where yes devbox/README.md is the one i screwed up


wimax-grapl:
i think i meant to say invocations but i've been playing too much elden ring

etc/formatter/format.sh
```diff
@@ -54,39 +54,48 @@ else
...
 # Slightly different config for yaml
-echo "--- Prettier YAML"
 prettier \
     --config prettierrc-yaml.toml \
     ${prettier_arg} \
     ${repo_root}/**/*.yml \
     ${repo_root}/**/*.yaml \
     ${repo_root}/.buildkite/**/*.yml \
-    ${repo_root}/.github/**/*.yml
+    ${repo_root}/.github/**/*.yml &
```
christophermaier, line 77: Does Prettier ignore dot files in its globs? I'm just wondering why the Buildkite and Github files have to be called out specifically here.

etc/formatter/format.sh
```diff
@@ -54,39 +54,48 @@ else
...
-echo "--- Prettier Markdown"
 prettier \
     ${prettier_arg} \
     --prose-wrap always \
     --print-width 80 \
-    ${repo_root}"/{,!(**/(target|*venv)/**)}**/*.md"
+    ${repo_root}"/{,!(**/(target|*venv)/**)}**/*.md" &
+PIDS+=("$!")
 
 # No config for markdown
```
christophermaier, line 88: Or TOML 😁

christophermaier (commented):
I'm not a huge fan of how the output of the different jobs is interleaved (as well as somewhat more opaque) 😦 

One thing I had been wanting to try was breaking up the invocation by file type, rather than having one call do 4 different things. If you want to format Markdown files, you don't want to have to wait for Javascript files to be formatted, for instance. I'm curious if something like that might help address the speed issue (which seems to be the main goal of this PR), while still preserving sane output?

etc/formatter/format.sh
wimax-grapl: I believe so, yes.

colin-grapl (approved):

PR #1636, opened 2022-04-19
remove superflous env var PIPELINE_INGRESS_BIND_ADDRESS

jgrillo-grapl:

NA

This PR removes a superfluous environment variable `PIPELINE_INGRESS_BIND_ADDRESS` which was obfuscating how the service actually works. In particular, it was not clear from looking at the service code that it was binding to a port supplied by nomad (see e.g. https://www.nomadproject.io/docs/integrations/consul-connect)

Similarly, it was unclear in the test code that the client _should not_ use the bind address to attempt to connect to the server. Removing this unnecessary environment variable makes the intentions clear.

Automated tests.

### DIFF
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -1255,7 +1255,6 @@ job "grapl-core" {
       env {
         AWS_REGION                                       = var.aws_region
         NOMAD_SERVICE_ADDRESS                            = "${attr.unique.network.ip-address}:4646"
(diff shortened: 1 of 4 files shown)


twunderlich-grapl (approved):

colin-grapl:
>  it was not clear from looking at the service code that it was binding to a port supplied by nomad 

I'm not sure why it should be clear in service code where the port is coming from.

>  it was unclear in the test code that the client should not use the bind address

I think the name `BIND_ADDRESS` is pretty clear tbh.

PR #1643, opened 2022-04-21
e2e test for consul dns resolving

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/887

This adds a test to confirm whether consul dns is working from inside containers and fixes the dns within containers so it can actually use consul dns.

make e2e-test


jgrillo-grapl (approved):
LGTM

jgrillo-grapl (approved):

inickles-grapl (approved):
Looks good, just one minor comment that's not important:

Can't say I'm a fan of the name `LOCAL_HOST_IP`, because despite the description where defined in the Makefile, it sounds like the value should be loopback IP, because `localhost` usually points to loopback iface. Although it _is_ a local host's IP, I'm thinking a name referring more to what it's used _for_ would be more clear, like `CONSUL_DNS_IP`. Not sure if we'd use `LOCAL_HOST_IP` for anything else, which might be why we'd wanna keep the name generic. It's fine IMO, feel free to leave as is.

PR #1645, opened 2022-04-21
comment out rust-integration-tests-new

jgrillo-grapl:
NA

This is an attempt to get the artifact into Cloudsmith. The hope is that we can
unblock the Pulumi run s.t. the image upload script will succeed.

Automated tests

### DIFF
--- a/nomad/integration-tests-new.nomad
+++ b/nomad/integration-tests-new.nomad
@@ -94,37 +94,39 @@ job "integration-tests-new" {
       }
     }
 
-    task "rust-integration-tests-new" {
-      driver = "docker"
+    # commented out temporarily to try to get an artifact into cloudsmith
+    #
+    # task "rust-integration-tests-new" {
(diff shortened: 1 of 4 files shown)


nomad/integration-tests-new.nomad
```diff
@@ -94,37 +94,39 @@ job "integration-tests-new" {
       }
     }
```
wimax-grapl, line 96: Instead of commenting this out, I'd recommend commenting out the Pulumi nomad-job-deployment stuff:
`integration_tests_new_job_vars`, `integration_tests_new`

The reason:
This will fail. The job still accepts a `container_images` variable, but you're skipping that in the `integration_tests_new_job_vars`.

wimax-grapl (changes requested):
Seems fine, but i'm unsure how nomad will feel about a

jgrillo-grapl:
Closing as this won't fix the problem.

PR #1646, opened 2022-04-21
add rust-integration-tests-new to cloudsmith-images

jgrillo-grapl:
NA

Adds `"rust-integration-tests-new"` to the `"cloudsmith-images"` group s.t. the testing image gets uploaded to Cloudsmith.

Automated tests

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -231,19 +231,12 @@ group "integration-tests" {
   ]
 }
 
-group "integration-tests-new" {
-  # NOTE: Please keep this list sorted in alphabetical order
-  targets = [
-    "rust-integration-tests-new"
-  ]
-}
-
 group "all-tests" {
   # NOTE: Please keep this list sorted in alphabetical order
   targets = [
     "e2e-tests",
     "integration-tests",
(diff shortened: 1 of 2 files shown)


wimax-grapl (approved):

docker-bake.hcl
```diff
@@ -162,7 +162,8 @@ group "cloudsmith-images" {
   # NOTE: Please keep this list sorted in alphabetical order
   targets = [
     "e2e-tests",
-    "grapl-services"
+    "grapl-services",
+    "rust-integration-tests-new"
```
christophermaier, line 166: It would probably be better to put the group `integration-tests-new` in here instead of the target `rust-integration-tests-new`.

(Or just remove the `integration-tests-new` group, if it's not serving a clear purpose right now. `integration-tests` is a group because we have Rust and Python, and `e2e-tests` is a bare target because there's only one.)

christophermaier (approved):
It'll work, but I think we can clean it up just a tad.

docker-bake.hcl
jgrillo-grapl: fixed

PR #1648, opened 2022-04-22
wire up the new integration test creds

jgrillo-grapl:
 NA

This PR updates the infrastructure configurations to account for the changes in https://github.com/grapl-security/confluent-cloud-infrastructure/pull/3.

The upshot of this is now we have Kafka credentials specifically for the integration test consumers, and we use them instead of trying to piggy-back on e.g. the pipeline-ingress credentials.

Automated tests

### DIFF
--- a/nomad/integration-tests-new.nomad
+++ b/nomad/integration-tests-new.nomad
@@ -36,19 +36,19 @@ variable "pipeline_ingress_healthcheck_polling_interval_ms" {
   description = "The amount of time to wait between each healthcheck execution."
 }
 
-variable "pipeline_ingress_kafka_consumer_group_name" {
+variable "integration_tests_kafka_consumer_group_name" {
   type        = string
-  description = "The name of the consumer group the pipeline-ingress consumers will join."
+  description = "The name of the consumer group the integration test consumers will join."
 }
 
-variable "pipeline_ingress_kafka_sasl_username" {
+variable "integration_tests_kafka_sasl_username" {
   type        = string
-  description = "The Confluent Cloud API key to configure pipeline-ingress producers and consumers with."
+  description = "The Confluent Cloud API key to configure integration test consumers with."
 }
 
-variable "pipeline_ingress_kafka_sasl_password" {
+variable "integration_tests_kafka_sasl_password" {
   type        = string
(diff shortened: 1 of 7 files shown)


pulumi/grapl/__main__.py
```diff
@@ -196,14 +196,19 @@ def main() -> None:
...
+    integration_tests_kafka_credentials = kafka.service_credentials("integration-tests")
+    pulumi.export(
+        "integration-tests-kafka-sasl-username",
+        integration_tests_kafka_credentials.apply(lambda c: c.api_key),
+    )
     pulumi.export(
-        "pipeline-ingress-kafka-consumer-group-name",
-        pipeline_ingress_kafka_consumer_group_name,
+        "integration-tests-kafka-sasl-password",
+        integration_tests_kafka_credentials.apply(lambda c: c.api_secret),
```
wimax-grapl, line 211: Purely hypothetically, couldn't these be declared in the `integration_tests/__main__.py`? That way you could skip the whole export-import song and dance, plus it's better namespacing.

wimax-grapl: Not saying you gotta do this, but just raising it as a possibility.

nomad/integration-tests-new.nomad
```diff
@@ -36,19 +36,19 @@ variable "pipeline_ingress_healthcheck_polling_interval_ms" {
   description = "The amount of time to wait between each healthcheck execution."
 }
 
-variable "pipeline_ingress_kafka_consumer_group_name" {
+variable "integration_tests_kafka_consumer_group_name" {
```
wimax-grapl, line 39: I'd argue you don't need to necessarily add these integration_tests_ prefixes given the filename, but not that important. I'm fine with the env vars having it, though.

wimax-grapl (approved):

pulumi/grapl/__main__.py
jgrillo-grapl: Yeah, I think they can. It'll take constructing the appropriate `Kafka()` but it seems doable. I'll give that a shot.

jgrillo-grapl: fixed

PR #1657, opened 2022-04-23
Upgrade a number of Cargo.toml dependencies, and un-patch-version-ify a few

wimax-grapl:
Cleaning stuff up. This isn't a full workspace-hack or anything but brings a number of our dependencies mostly-in-sync.


src/rust/rust-proto-new/Cargo.toml
```diff
@@ -12,17 +12,16 @@ futures = "0.3"
 prost = "0.10"
 prost-types = "0.10"
 thiserror = "1.0"
-tokio = { version = "1.14", features = ["rt", "time"] }
+tokio = { version = "1.17", features = ["rt", "time"] }
```
colin-grapl, line 15: We probably want to use the same tokio features everywhere

colin-grapl (approved):

src/rust/rust-proto-new/Cargo.toml
jgrillo-grapl: why?

PR #1663, opened 2022-04-26
Button bug

andrea-grapl:
https://github.com/grapl-security/issue-tracker/issues/903

Remove directional arrows & link particles - distracting 
Lens Names are all capped, should be case sensitive - Better UX
Highlight Current Page on Nav Bar - Better UX
Ability to copy and paste lens names "click here to copy lens name" feature - Better UX

Manually - `make test-e2e`, click sidebar, lens names, copy lens names, etc.


src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -187,29 +184,29 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
...
+            // ctx.save();
             ctx.beginPath();
             ctx.arc(node.x, node.y, NODE_R * 1.4, 0, 2 * Math.PI, false);
             ctx.fillStyle =
                 node === hoverNode
                     ? colors.hoverNodeFill
                     : riskOutline(node.risk_score);
             ctx.fill();
-            ctx.restore();
+            // ctx.restore();
```
colin-grapl, line 198: let's remove these commented lines

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -187,29 +184,29 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
...
 
             // Node Fill Styling
-            ctx.save();
+            // ctx.save();
```
colin-grapl, line 201: delete

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -218,7 +215,7 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
                 98,
                 NODE_R / ctx.measureText(label).width
             );
-            ctx.save();
+            // ctx.save();
```
colin-grapl, line 218: delete

src/js/engagement_view/src/types/LensAndNodeTableTypes.tsx
```diff
@@ -6,6 +6,8 @@ export type SelectLensProps = {
     uid: number;
     lens_type: string;
     setLens: (lens: string) => void;
+    selectedId: number;
+    setLensTableState: any;
```
colin-grapl, line 10: let's type this instead of any

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/lensTable.tsx
```diff
@@ -15,6 +15,8 @@ import { PaginationState } from "types/CustomTypes";
 
 export const lensTable = (
     state: PaginationState,
+    selectedIdState: number,
+    setSelectedIdState: (selectedIdState: number) => void,
```
wimax-grapl, line 19: preface for the rest of this PR: I am not a react expert

I guess I'm confused why there's additional state outside the variable named `state`
could we rename that to `paginationState`?

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/lensTable.tsx
```diff
@@ -71,6 +73,8 @@ export const lensTable = (
                                     lens_type={lens.lens_type}
                                     score={lens.score}
                                     setLens={setLens}
+                                    selectedId={selectedIdState}
+                                    setLensTableState={setSelectedIdState}
```
wimax-grapl, line 77: there's an incongruency in these args and their values, could we maybe sync them up a bit more if it makes sense to do so?

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/lensTableStyles.tsx
```diff
@@ -56,7 +56,12 @@ export const useStyles = makeStyles((theme) =>
             width: "95%",
         },
         tableRow: {
-            background: theme.palette.background.default,
+            "&.Mui-selected": {
+                backgroundColor: "#2B3648",
+                "& > .MuiTableCell-root": {
+                    color: "#1976D2",
+                },
+            },
```
wimax-grapl, line 64: It's a bit odd to me that these are the only bare hex-colors in this file. Is there any way we can associate them with the Theme palette? Would that make sense to do or are they unrelated?

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/selectLens.tsx
```diff
@@ -11,21 +10,19 @@ import { SelectLensProps } from "types/LensAndNodeTableTypes";
...
-                    // className={classes.lensName}
-                    className={btnClass ? "btnClass clicked" : "btnClass"}
-                    onClick={() => {
-                        props.setLens(props.lens);
-                        btnClass ? setBtnClass(false) : setBtnClass(true);
-                    }}
-                >
-                    {props.lens_type + " :\t\t" + props.lens}
-                </Button>
+                {props.lens_type + " :\t\t" + props.lens}
```
wimax-grapl, line 25: could you leave a comment on what does :\t\t do

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/lensTable.tsx
wimax-grapl, line 77: could we call this setLensSelectedId or something?

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -187,29 +184,25 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
...
-            ctx.save();
             ctx.beginPath();
             ctx.arc(node.x, node.y, NODE_R * 1.4, 0, 2 * Math.PI, false);
             ctx.fillStyle =
                 node === hoverNode
                     ? colors.hoverNodeFill
                     : riskOutline(node.risk_score);
             ctx.fill();
-            ctx.restore();
+            // ctx.restore();
```
wimax-grapl, line 196: I see this as 'resolved' but it's still commented?

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -249,6 +241,10 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
         const LABEL_NODE_MARGIN = 12;
         const start = link.source;
         const end = link.target;
+
+        link.fx = link.x;
+        link.fy = link.y;
+
```
wimax-grapl, line 247: # try this change after you do everything else
and i'm not opposed if you do it in a followup PR, as long as you do _do_ it 
---

We can actually get correct types on `constlinkStyling` fwiw
https://github.com/vasturiano/react-force-graph/blob/master/src/packages/react-force-graph-2d/index.d.ts#L71
```
it's a CanvasCustomRenderFn<LinkObject>;
which boils down to a 
(link: LinkObject, canvasContext: CanvasRenderingContext2D) => {
```

you'll have to import LinkObject type from react-force-graph-2d
and you'll have to import CanvasRenderingContext2D from some presupplied web api typescript gives you
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D

---
that said, here I see you doing like `link.x, link.y` whereas LinkObject in the linked source code says it only has properties `.source` and `.target`, so, I'm a tad confused how that all works...

src/js/engagement_view/src/components/reusableComponents/collapsibleDrawer.tsx
```diff
@@ -107,40 +116,56 @@ export default function CollapsibleNavDrawer() {
                 </DrawerHeader>
                 <Divider />
                 <List className={classes.drawer}>
-                    <ListItem button key="Analyzers">
+                    <ListItemButton
+                        key="Analyzers"
+                        selected={selectedIndex === 0}
+                        onClick={(event) => handleListItemClick(event, 0)}
+                    >
```
wimax-grapl, line 123: I won't stop the PR on this, but this section definitely has a code smell. Consider doing something like the following:
```
interface ListItem {
  label: String,
  link: String,
}

const items: ListItem[] = [
  { label: "Analyzers", link: "/analyzers" },
  { label: "Engagements", link: "/engagements" },
];

items.forEach((value, index) => {
	// render each item as a ListItemButton
});

```

src/js/engagement_view/src/components/reusableComponents/drawer.tsx
```diff
@@ -34,6 +35,21 @@ const DrawerHeader = styled("div")(({ theme }) => ({
 export const NavigationDrawer = () => {
     const classes = useStyles();
 
+    const location = window.location.href.split("/").slice().pop();
+
+    let index: number = 0;
+
+    location === "analyzers" ? (index = 0) : (index = 2);
+
```
wimax-grapl, line 43: Okay, this I'd block the PR on. It's basically re-stating the "analyzers is 0" mapping in collapsibleDrawer.tsx, but obviously the code here is wrong (it's not always 2 if it's not analyzers)

wimax-grapl: While I don't love using the window.location as a tracker of state, you could use it and the aforementioned 
```

const items: ListItem[] = [
  { label: "Analyzers", link: "/analyzers" },
  { label: "Engagements", link: "/engagements" },
];
```
to accurately figure out which index it currently is.

src/js/engagement_view/src/components/reusableComponents/drawer.tsx
```diff
@@ -56,40 +72,55 @@ export const NavigationDrawer = () => {
             </DrawerHeader>
 
             <List className={classes.drawer}>
```
wimax-grapl, line 74: More proof of repeated code-smell - this entire section is repeated across files. Maybe this can be extracted out to its own component?

wimax-grapl (changes requested):

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/lensTable.tsx
andrea-grapl: Sorry, this is unclear - this state refers to the selected lens, not selected pagination. We have two separate states here, one for pagination, one for current selected lens. Will rename the variables to be more clear.

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
andrea-grapl: This is dead code that shouldn't be in here, sorry - was testing something and forgot to remove this.You're right with the  source and target.

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/lensTableStyles.tsx
andrea-grapl: Moved to theme, now referenced in pallete

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/selectLens.tsx
andrea-grapl: Dead and unnecessary code - removed.

src/js/engagement_view/src/components/reusableComponents/collapsibleDrawer.tsx
andrea-grapl: Used a new feature to set active links in upgrade to react router dependency.

src/js/engagement_view/src/components/reusableComponents/drawer.tsx
andrea-grapl: Also resolved with react router update!

src/js/engagement_view/src/components/reusableComponents/drawer.tsx
andrea-grapl: Set active link with upgrade to react router!

src/js/engagement_view/Makefile
```diff
@@ -108,7 +108,7 @@ shell: ## Drop into a shell inside the build container
 
 .PHONY: start
 start: extra_docker_run_args := --interactive --tty --network host
-start: build-image
+start: build-code
```
wimax-grapl, line 111: This suggests that you need to `yarn build` (make a release build) before you `yarn start` (which is the debugging one that auto-updates I believe), which seems incorrect.

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/lensTable.tsx
```diff
@@ -71,6 +73,8 @@ export const lensTable = (
                                     lens_type={lens.lens_type}
                                     score={lens.score}
                                     setLens={setLens}
+                                    selectedId={selectedIdState}
+                                    setSelectedLens={setSelectedIdState}
```
wimax-grapl, line 77: There's a mismatch in these state and setter names.

IMO this should look like
```
                                    selectedLensId={selectedLensIdState}
                                    setSelectedLensId={setSelectedLensIdState}
```

a `number` named `id` tells me nothing about how that variable/state is used; I almost always try to label my IDs (lens_id, page_id, user_id, etc).

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/selectLens.tsx
```diff
@@ -11,21 +10,19 @@ import { SelectLensProps } from "types/LensAndNodeTableTypes";
...
-    const [btnClass, setBtnClass] = useState(false);
-
     return (
-        <TableRow className={classes.tableRow} key={props.uid}>
+        <TableRow
+            hover
+            key={props.uid}
+            onClick={() => {
+                props.setSelectedLens(props.uid);
+                props.setLens(props.lens);
```
wimax-grapl, line 19: more proof that, yeah, setSelectedLens should be named setSelectedLensId or setSelectedLensUid or something - the fact that `setSelectedLens` and `setLens` take two different types isn't great.

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
```diff
@@ -218,7 +210,6 @@ const GraphDisplay = ({ lensName, setCurNode }: GraphDisplayProps) => {
                 98,
                 NODE_R / ctx.measureText(label).width
             );
```
wimax-grapl, line 212: and just to make sure, all this removal of ctx is expected?

src/js/engagement_view/src/components/reusableComponents/collapsibleDrawer.tsx
```diff
@@ -101,59 +91,17 @@ export default function CollapsibleNavDrawer() {
                             )}
                         </IconButton>
                         <Icon>
-                            <img className={classes.logoImage} src={Img} />
+                            <img
+                                className={classes.logoImage}
+                                src={Img}
+                                alt="Grapl Logo"
+                            />
```
wimax-grapl, line 98: Looks like we have a number of these - perhaps we should make this its own `<GraplLogo />` component?

image

src/js/engagement_view/src/components/reusableComponents/drawerList.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <>
+            <List className={classes.drawer}>
+                <ListItemButton>
+                    <NavLink
+                        to="/analyzers"
+                        className={classes.navLink}
+                        style={({ isActive }) => ({
+                            color: isActive ? "#1A76D2" : "#C6D1E7",
+                        })}
+                    >
```
wimax-grapl, line 29: It looks like react-router offers a more robust way to do this:

https://v5.reactrouter.com/web/api/NavLink/activeclassname-string

by default, when a NavLink becomes active, its `class` becomes `className + ".active"`

so I think you can just do this in the styles file

```
        navLink: {
            textDecoration: "none",
            color: theme.palette.info.contrastText,
            "& .active": {
              color: theActiveColorGoesHere
            }
        },
```

I _believe_ the & is how you do `.navlink.active` in this CSS library, but I'm not entirely sure.

src/js/engagement_view/src/theme.tsx
```diff
@@ -30,6 +31,18 @@ const theme = createTheme({
             contrastText: "#FFFFFF",
         },
     },
+    components: {
+        MuiListItemButton: {
+            styleOverrides: {
+                root: {
+                    "&.Mui-selected": {
```
wimax-grapl, line 38: yeah, this is the same `& .class` thing I mentioned above, for the record

src/js/engagement_view/src/components/graphDisplay/GraphDisplay.tsx
andrea-grapl: Yes

src/js/engagement_view/Makefile
andrea-grapl: Leaving per cm

src/js/engagement_view/src/components/reusableComponents/collapsibleDrawer.tsx
andrea-grapl: Done. Good call.

src/js/engagement_view/src/components/reusableComponents/drawerList.tsx
andrea-grapl: I think the above method was right for v5 of react router, but according to the linked docs for v6 (which we just started using) I think the docs want us to style active links with functions. Happy to change this, but according to the following paragraph, I think we're supposed to be using inline styles, but could totally be reading this wrong. Happy to change if necessary. 

```
In React Router v6, activeStyle will be removed and you should use the function style to apply inline styles to either active or inactive NavLink components.
```

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/selectLens.tsx
andrea-grapl: done

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/lensTable.tsx
andrea-grapl: done

src/js/engagement_view/src/components/engagementView/sidebar/tables/lensTable/lensTable.tsx
andrea-grapl: resolved in different comment

wimax-grapl (approved):
Approved, and I won't stop the PR if you don't include this, but worth noting that the `.active` suggestion still exists.

per slack:
```
actually, after looking at the v6 docs, it looks like my suggestion still stands
https://reactrouter.com/docs/en/v6/api#navlink
By default, an active class is added to a <NavLink> component when it is active. This provides the same simple styling mechanism for most users who are upgrading from v5. One difference as of v6.0.0-beta.3 is that activeClassName and activeStyle have been removed from NavLinkProps.
basically v5 gave you too many options to customize stuff so they ripped it out in v6

```

PR #1665, opened 2022-04-26
Add protocol::Status

colin-grapl:
910

Adds a straightforward implementation of a Status, modeled after the tonic::Status but significantly stripped down (no arbitrary metadata support, no rust error support).

I kept it small so we can add changes incrementally but otherwise get moving with this.

The code is basically copy pasted from tonic. I don't know what tests I'd write that aren't basically tautological. If you want to add tests though go nuts.


src/rust/rust-proto-new/src/protocol/status.rs
```diff
@@ -0,0 +1,363 @@
+use tonic::{Code as GrpcCode, Status as GrpcStatus};
+
+
+/// Status codes used by [`Status`].
+///
+/// These variants closely mirror the [gRPC status codes].
+///
+/// [gRPC status codes]: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md#status-codes-and-their-use-in-grpc
```
jgrillo-grapl, line 8: Do we actually need to support all these codes? For example, where will we use a `DeadlineExceeded` or a `FailedPrecondition`? Let's start out supporting only the error states we intend to actually use.

src/rust/rust-proto-new/src/protocol/status.rs
```diff
@@ -0,0 +1,363 @@
...
+            GrpcCode::Unavailable => Code::Unavailable,
+            GrpcCode::DataLoss => Code::DataLoss,
+            GrpcCode::Unauthenticated => Code::Unauthenticated,
+        }
+    }
+}
+
+pub struct Status {
+    /// The gRPC status code, found in the `grpc-status` header.
+    code: Code,
```
jgrillo-grapl, line 159: Why is a `Status` composed of a `code` and a `message`? Does this give us some advantage over an enumeration? It feels to me like an `enum Status` would be more ergonomic to callers.

src/rust/rust-proto-new/src/protocol/status.rs
```diff
@@ -0,0 +1,363 @@
...
+        GrpcStatus::new(status.code.into(), status.message)
+    }
+}
+
+impl From<GrpcStatus> for Status {
+    fn from(status: GrpcStatus) -> Self {
+        Status::new(status.code().into(), status.message())
+    }
+}
+
```
jgrillo-grapl, line 347: lots of extra newlines

jgrillo-grapl (changes requested):
I have some questions, best summarized by asking why should gRPC's abstraction necessarily be our abstraction?

src/rust/rust-proto-new/src/protocol/status.rs
colin-grapl: Structs are much easier to extend. If I want to add a field to an enum I have to add that field to every single variant. A struct wrapping an enum is a fairly common pattern for this reason.

src/rust/rust-proto-new/src/protocol/status.rs
colin-grapl: The only two I can imagine removing are DeadlineExceeded and, perhaps, DataLoss. FailedPrecondition is indeed something I could imagine returning - for example, if a service tried to access a table in a database but the table had not been created, just as a somewhat contrived example.

So I could remove DeadlineExceeded and DataLoss?

colin-grapl:
> I have some questions, best summarized by asking why should gRPC's abstraction necessarily be our abstraction?

Just seems like it's fairly comprehensive and it makes this straightforward to apply. We could make up our own, but I don't see much reason to do so. gRPC's abstraction is a lot smaller and easier to extend than, say, HTTPs, so yeah I don't see a reason to use something else.

src/rust/rust-proto-new/src/protocol/status.rs
jgrillo-grapl: Do we expect we'll need to add more fields?

src/rust/rust-proto-new/src/protocol/status.rs
jgrillo-grapl: Well, I think instead it would be best to let requirements drive the design here. Which status variants do we currently have a demonstrated (e.g. in a service) need to support? Let's implement _those_ variants (and only those invariants) until such time as we require others.

colin-grapl: I've removed a few of the variants. What's left, based on a quick grep and analysis of our current use of 'internal' where we should be using other codes, as well as accepted RFCs, should be just what we need now or in the immediate future.

src/rust/rust-proto-new/src/protocol/status.rs
colin-grapl: Possibly. The tonic::Status includes the source rust error, which lets it impl Error, as well as as an http metadata map.

jgrillo-grapl (approved):

PR #1666, opened 2022-04-27
Workaround for hwloc-induced scylladb crash on ChromeOS

ghost:
Closes https://github.com/grapl-security/issue-tracker/issues/912

Currently there's a chrome bug https://bugs.chromium.org/p/chromium/issues/detail?id=1304418
that causes the crostini vm to send invalid hardware information. This
bug causes scylla to crash on startup with a SIGFPE https://github.com/scylladb/scylla/issues/10439.
Since hwloc has already issues a patch to deal with this, we're
replacing the dynamically linked lib scylladb uses with the patched version

### DIFF
--- a/scylladb/Dockerfile
+++ b/scylladb/Dockerfile
@@ -1,9 +1,10 @@
 FROM debian:bullseye-slim as build
 
 ENV HWLOC_VERSION="2.7.1"
+# At least one of the install-recommends is necessary in order for hwloc to compile. 
+# hadolint ignore=DL3015
 RUN apt-get update \
-    && apt-get -y install \
-        curl=7.74.0-1.3+deb11u1 \
+    && apt-get -y install curl=7.74.0-1.3+deb11u1 \
     libssl-dev=1.1.1n-0+deb11u1 \
     ca-certificates=20210119  \
     gcc=4:10.2.1-1 \
@@ -14,10 +15,9 @@ curl --proto "=https" \
     --location \
     --output /tmp/hwloc-${HWLOC_VERSION}.tar.gz \
 "https://download.open-mpi.org/release/hwloc/v2.7/hwloc-${HWLOC_VERSION}.tar.gz" && \
-cd /tmp && \
-tar -vxzf hwloc-${HWLOC_VERSION}.tar.gz && \
-cd hwloc-${HWLOC_VERSION} && \
-./configure && \
+tar -vxzf /tmp/hwloc-${HWLOC_VERSION}.tar.gz
+WORKDIR /tmp/hwloc-${HWLOC_VERSION}
+RUN ./configure && \
 make && \
 make install
 
--- a/scylladb/Dockerfile
+++ b/scylladb/Dockerfile
@@ -1,4 +1,4 @@
-FROM scylladb/scylla
+FROM scylladb/scylla:4.6.3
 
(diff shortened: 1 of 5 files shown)


scylladb/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM scylladb/scylla
+
+# This downloads a prebuild static lib for hwloc 2.7.1 which includes a chromeos fix, and replaces the dynamically
+# linked version that scylla uses. See the README for more info
+
+RUN curl --proto "=https" \
+    --tlsv1.2 \
+    --location \
+    --output /usr/local/lib/libhwloc.so.15.5.3 \
+    "https://dl.cloudsmith.io/public/grapl/thomas/raw/versions/15.5.3/libhwloc.so.15.5.3" && \
```
christophermaier, line 10: We can (and probably should) put this in one of our "real" repositories (like `releases`), rather than your personal one.

scylladb/README.md
```diff
@@ -0,0 +1,35 @@
...
+    --tlsv1.2 \
+    --location \
+    --output /tmp/hwloc-${HWLOC_VERSION}.tar.gz \
+"https://download.open-mpi.org/release/hwloc/v2.7/hwloc-${HWLOC_VERSION}.tar.gz"
+cd /tmp
+tar -vxzf hwloc-${HWLOC_VERSION}.tar.gz
+cd hwloc-${HWLOC_VERSION}
+./configure
+make
+make install
```
christophermaier, line 29: I was going to suggest creating an actual script for this, so it would be executable. Another possibility would be to fold this into the Dockerfile itself (e.g., in a different target). I'm not sure how long it takes to build, but if it's not too long, it might be good to remove some additional moving parts (including the need to store anything in Cloudsmith, not that that's a horrible thing to do, of course).

christophermaier (approved):
Couple suggestions, but 👍 overall! Nice solution!

scylladb/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM scylladb/scylla
+
+# This downloads a prebuild static lib for hwloc 2.7.1 which includes a chromeos fix, and replaces the dynamically
+# linked version that scylla uses. See the README for more info
+
```
colin-grapl, line 5: Can we add a comment here to the Scylla issue tracker / our internal tracker here

colin-grapl: Nevermind, I see it below!

scylladb/README.md
```diff
@@ -0,0 +1,35 @@
+# Summary
+
+This directory includes a dockerfile to create a patched version of scylladb so
+it will run in chromeos. Without it, scylladb
+[crashes with a SIGFPE error from hwloc](https://github.com/scylladb/scylla/issues/10439).
+This is due to a
+[chromeos bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1304418).
+Fortunately, hwloc has released a
+[patch](https://github.com/open-mpi/hwloc/commit/33b555b5a1c8339daeb4215bad57430b57a6b33f)
+in 2.7.1.
```
colin-grapl, line 10: Hell yeah docs

colin-grapl (approved):

scylladb/README.md
colin-grapl: Moving into the Dockerfile as a first stage makes sense to me

twunderlich-grapl: I've moved it into the Dockerfile

PR #1669, opened 2022-04-28
Proof-of-concept http (not grpc) grapl plugin to experiment with consul proxy stuff

wimax-grapl:
`test_deploy_plugin` will now deploy a plugin that just runs `python3 -m http.server`; this will in the _very_ near future be replaced with a grpc server implementing the plugin sdk, but in the meantime, this really helps me validate that the proxy stuff is working correctly.

---

To use it:

# Start environment and deploy plugin
KEEP_TEST_ENV=1 make test-integration-new

# Get a proxy into the plugin service
PLUGIN_SERVICE="plugin-5fe1d97a-3d86-6e04-0193-8b121dba1a26"
consul connect proxy -service respresenting-devbox-1 -upstream ${PLUGIN_SERVICE}:9999

# In another tab, curl against the proxy
```
$ curl http://localhost:9999
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
<title>Directory listing for /</title>
...
```

### DIFF
--- a/src/rust/plugin-registry/tests/small_test_binary.sh
+++ b/src/rust/plugin-registry/tests/small_test_binary.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+echo "Hello, I am small_test_binary.sh"
+echo "Starting a simple server on ${BIND_PORT}..."
+apt-get update
+apt-get install --yes python3
+python3 -m http.server "${BIND_PORT}"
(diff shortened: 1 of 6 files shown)


src/rust/plugin-registry/tests/small_test_binary.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+echo "Hello, I am small_test_binary.sh"
+echo "Starting a simple server on ${BIND_PORT}..."
+apt-get update 
+apt-get install --yes python3
```
wimax-grapl, line 5: it's weird but this is a short term hack so whatever

colin-grapl (approved):

PR #1674, opened 2022-05-02
Hax: Default to Google DNS if CONSUL_DNS_IP doesn't exist

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/924

The CONSUL_DNS_IP env variable doesn't exist in Buildkite, and doesn't
really make sense at the moment in AWS. Since this is going away
somewhat shortly once the fix is merged into Nomad, this is a temporary
fix.

Commented out the CONSUL_DNS_IP in docker-compose and ran make up without problems.


christophermaier (approved):

christophermaier:
Could we add an issue to track this particular hack, just to keep track of it?

twunderlich-grapl:
Added. https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/925 is the ticket for fixing/cleaning this up

PR #1677, opened 2022-05-02
Generator executor

wimax-grapl:
Satisfies the first 2 checkboxes of 
https://github.com/grapl-security/issue-tracker/issues/733
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/733

My next steps are to do a couple of the blockers on this. 
(If it seems like I've been doing things out-of-order: I have been!) 

Implement the Graph Generator Executor as described in the Plugins RFC.
https://github.com/grapl-security/grapl-rfcs/blob/main/text/0000-plugins.md#-executor-apis-graph-generator-executor-analyzer-executor

Still has quite a bit TODO.

So, here's the thing - this service is basically a one-file glue service. I'm not convinced it's worth unit testing, though I absolutely am obviously convinced it's worth integration testing.

However! 

As we've seen from previously introducing new services, you generally want to introduce the new docker image *before* you integrate it into Pulumi and our whole stack.

### DIFF
--- a/src/rust/Dockerfile
+++ b/src/rust/Dockerfile
@@ -425,6 +425,15 @@ FROM rust-dist AS osquery-generator-deploy
 COPY --from=build /outputs/osquery-generator /
 ENTRYPOINT ["/osquery-generator"]
 
+##### generator-executor
+FROM rust-dist AS generator-executor-deploy
+
+# Hadolint v2.8.0 appears to be mistaken, this is a false positive
+# https://github.com/hadolint/hadolint/wiki/DL3022
+# hadolint ignore=DL3022
+COPY --from=build /outputs/generator-executor /
+ENTRYPOINT ["/generator-executor"]
+
 ##### web-ui
 FROM rust-dist AS grapl-web-ui-deploy
 
--- a/src/rust/plugin-executor/Cargo.toml
+++ b/src/rust/plugin-executor/Cargo.toml
@@ -22,7 +22,7 @@ plugin-work-queue = { path = "../plugin-work-queue" }
 rust-proto = { path = "../rust-proto" }
 structopt = { version = "0.3", default-features = false }
 tokio = { version = "1.17", features = ["macros", "rt", "rt-multi-thread"] }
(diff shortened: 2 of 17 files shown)


src/rust/plugin-executor/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[dependencies]
+async-trait = "0.1.51"
+generator-sdk = { path = "../plugin-sdk/generator-sdk", features = ["client"] }
+plugin-work-queue = { path = "../plugin-work-queue" }
+rust-proto = { path = "../rust-proto" }
+structopt = { version = "0.3", default-features = false }
+tokio = { version = "1.17", features = ["macros", "rt", "rt-multi-thread"] }
+tonic = "0.6" # TODO move to 0.7 once we upgrade plugin-work-queue to -new
```
wimax-grapl, line 25: or just, yknow, remove it

jgrillo-grapl (commented):
Would be 👍 👍 to port this to `rust-proto-new`

wimax-grapl:
> Would be 👍 👍 to port this to `rust-proto-new`

Yep, but I'm gonna do that as a separate PR. Logic first, then upgrade.

src/rust/Dockerfile
```diff
@@ -425,6 +425,15 @@ FROM rust-dist AS osquery-generator-deploy
 COPY --from=build /outputs/osquery-generator /
 ENTRYPOINT ["/osquery-generator"]
 
+##### generator-executor
+FROM rust-dist AS generator-executor-deploy
```
jgrillo-grapl, line 429: should this be "plugin-executor-deploy"?

jgrillo-grapl (approved):

src/rust/Dockerfile
wimax-grapl: No, although this is up for reexamination later. If you look at Cargo.toml this package outputs 2 bins. The hope is to share logic between the two in this one package's lib.

PR #1679, opened 2022-05-02
Set option to make logs readable in CI

ghost:
This makes the logs readable in CI, so when there are issues you can actually go back and see where in the logs something failed. Right now the logs are cut off so its hard to tell what's being pushed or tagged

Ran the --progress plain flag manually in my terminal


wimax-grapl (approved):

jgrillo-grapl (approved):
👏

PR #1680, opened 2022-05-03
Try fixing MANIFEST_UNKNOWN error by splitting build and push

ghost:
This splits the building images and the pushing images into two separate commands. This could potentially fix the issue _if_ the issue stems from the manifest being missing locally. However if the issue is from cloudsmith, then this just adds extra time.

### DIFF
--- a/.buildkite/scripts/build_and_upload_images.sh
+++ b/.buildkite/scripts/build_and_upload_images.sh
@@ -36,6 +36,8 @@ echo "--- Building all ${IMAGE_TAG} images"
 # all the buildx file introspection (and thus build-target awareness)
 # localised here.
 make build-image-prerequisites
+
+docker buildx bake --file="${BUILDX_BAKE_FILE}" --progress "plain" "${BUILDX_TARGET}"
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/build_and_upload_images.sh
```diff
@@ -36,6 +36,8 @@ echo "--- Building all ${IMAGE_TAG} images"
 # all the buildx file introspection (and thus build-target awareness)
 # localised here.
 make build-image-prerequisites
+
+docker buildx bake --file="${BUILDX_BAKE_FILE}" --progress "plain" "${BUILDX_TARGET}"
 docker buildx bake --file="${BUILDX_BAKE_FILE}" --progress "plain" --push "${BUILDX_TARGET}"
```
christophermaier, line 41: I'm not sure if this is going to actually do what you want. `--push` is just a shorthand for "the output of this build is a registry image at XXX registry".

True, the first `bake` invocation will just build the images locally, and the second one will essentially be a no-op due to everything being cached. I don't think that building is the issue, though.

twunderlich-grapl:
Closing as this failed, which makes it more likely to be a cloudsmith error

PR #1686, opened 2022-05-03
Refactor frontend artifacts so they're dumped to dist/ not src/rust/grapl-web-ui

wimax-grapl:
Currently we are shoving `engagement-view` artifacts in `src/rust` when we don't really need to.

Right now we have the following:
```.PHONY: build-engagement-view
build-engagement-view: ## Build website assets to include in grapl-web-ui
	[...] 
	# copy the built engagement view artifacts to src/rust/grapl-web-ui/frontend```

This cache-busts the Rust build `FROM rust:1-slim-bullseye AS base` because of
`COPY rust rust`

That `frontend/` artifact is not _actually_ used in `rust-base` build at all; it's only used in
```##### web-ui
FROM rust-dist AS grapl-web-ui-deploy

COPY --from=build /outputs/grapl-web-ui /
COPY rust/grapl-web-ui/frontend /frontend
ENTRYPOINT ["/grapl-web-ui"]```

*SOLUTION*
We shove `/frontend` into `dist/` 
We use `dist/` as a named secondary context

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,9 +30,6 @@ node_modules/
 # corepack installs the node version in .node. Kinda annoying but makes sense since its not yarn-specific
 .node/
 
-# Web UI frontend artifacts
-src/rust/grapl-web-ui/frontend/*
-
 # Nomad artifacts
 # This is a generated config file for local grapl
 nomad/local/mount-grapl-root-as-volume.nomad
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -412,7 +412,7 @@ target "sysmon-generator" {
 target "_python-base" {
   contexts = {
     dist-ctx = "dist"
-    etc-ctx = "etc"
+    etc-ctx  = "etc"
   }
   dockerfile = "src/python/Dockerfile"
   labels     = oci_labels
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: docker-bake.hcl [hunk 1]
```diff
@@ -269,7 +269,13 @@ group "all" {
...
 # this target.
 target "_rust-base" {
-  context    = "src"
+  context = "src"
+
+  # Additional named contexts: 
+  # https://www.docker.com/blog/dockerfiles-now-support-multiple-build-contexts/
+  contexts = {
+    dist-ctx = "dist"
+  }
```
wimax-grapl, line 278: tbh I like that it's named something slightly different than `dist` so that one can ctrl-f backtrack to figure out "what the hell is this `--from` about" but i'm not married to it

.gitignore
```diff
@@ -30,9 +30,6 @@ node_modules/
 # corepack installs the node version in .node. Kinda annoying but makes sense since its not yarn-specific
 .node/
 
-# Web UI frontend artifacts
-src/rust/grapl-web-ui/frontend/*
-
```
wimax-grapl, line 35: dist/ is entirely ignored

christophermaier (approved):
👍

wimax-grapl:
https://github.com/hadolint/hadolint/issues/830

PR #1694, opened 2022-05-05
Port plugin_work_queue and its dependencies to rust-proto-new

wimax-grapl:
Fulfill a promise I made on https://github.com/grapl-security/grapl/pull/1677#issuecomment-1115443765

- port plugin_work_queue protos out of rust-proto into rust-proto-new
- upfix plugin_work_queue service to use rust-proto-new
- upfix plugin-executor to use rust-proto-new
- delete old rust-proto version of this

Added SerDe tests for the models.
Skipped the mocking-out-backing-api step for now since plugin-work-queue already has robust integration testing that exercises the real thing.


jgrillo-grapl (approved):
LGTM

PR #1699, opened 2022-05-07
'shuttle-rust-to-dist': a way to pull binary artifacts out of `rust-build` into `dist/`.

wimax-grapl:
So, we actually _already had this functionality_ for plugin-bootstrap-init, but I wanted to genericize the approach. 
I realized we'd need this because I'll soon be shuttling a Generator plugin binary out into `dist/`.

There's no real reason to separate this into separate images (i.e. one for the generator binary, one for the plugin-bootstrap-init binary).
- All of the binaries are built together all at once
- Nothing directly consumes the image, it's just an implementation detail of pulling stuff out into dist/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -659,11 +659,11 @@ dist/firecracker_kernel.tar.gz: firecracker/kernel/build.sh | dist
 # NOTE: While this target is PHONY, it *does* represent a real directory in
 # dist/
 .PHONY: dist/plugin-bootstrap-init
-dist/plugin-bootstrap-init: _shuttle-rust-to-dist  ## Build the Plugin Bootstrap Init (+ associated files) and copy it to dist/
+dist/plugin-bootstrap-init: _export-rust-build-artifacts-to-dist  ## Build the Plugin Bootstrap Init (+ associated files) and copy it to dist/
 
-.PHONY: _shuttle-rust-to-dist
-_shuttle-rust-to-dist: | dist  ## Copy all specified Rust binary artifacts to dist/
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -659,8 +659,11 @@ dist/firecracker_kernel.tar.gz: firecracker/kernel/build.sh | dist
 # NOTE: While this target is PHONY, it *does* represent a real directory in
 # dist/
 .PHONY: dist/plugin-bootstrap-init
-dist/plugin-bootstrap-init: | dist  ## Build the Plugin Bootstrap Init (+ associated files) and copy it to dist/
-	$(DOCKER_BUILDX_BAKE_HCL) plugin-bootstrap-init
+dist/plugin-bootstrap-init: _shuttle-rust-to-dist  ## Build the Plugin Bootstrap Init (+ associated files) and copy it to dist/
+
+.PHONY: _shuttle-rust-to-dist
+_shuttle-rust-to-dist: | dist  ## Copy all specified Rust binary artifacts to dist/
```
jgrillo-grapl, line 665: Can we call this something a little more descriptive, like maybe `extract-binary-artifacts`? I can't tell from the name what this does.

jgrillo-grapl (approved):
LGTM formally, I don't get the naming but that's not a blocker.

Makefile
wimax-grapl: Yeah, I'm cool with that. There are some artifacts that aren't strictly binaries tho 😱 but i'll play around with it

PR #1701, opened 2022-05-09
Rust-Proto-New-ify plugin_sdk/generators

wimax-grapl:
Arguably related to https://github.com/grapl-security/issue-tracker/issues/752 but tbh that was already satisfied.

Port the Graph Generators SDK types to rust-proto-new.

SerDe types are proptest'd. 
The server/client types are currently untested, but will be with a parallel PR I'm doing (I actually cut this code out of that PR).

### DIFF
--- a/src/rust/rust-proto-new/src/graplinc/grapl/api/plugin_sdk/generators/v1beta1_server.rs
+++ b/src/rust/rust-proto-new/src/graplinc/grapl/api/plugin_sdk/generators/v1beta1_server.rs
@@ -1,0 +1,166 @@
+use std::{
+    marker::PhantomData,
+    time::Duration,
+};
+
+use futures::{
+    channel::oneshot::{
+        self,
+        Receiver,
+        Sender,
+    },
+    Future,
+    FutureExt,
+};
+use thiserror::Error;
+use tokio::net::TcpListener;
+use tokio_stream::wrappers::TcpListenerStream;
+use tonic::{
+    transport::{
+        NamedService,
+        Server,
+    },
+    Request,
+    Response,
+};
+
+use crate::{
+    execute_rpc,
+    graplinc::grapl::api::plugin_sdk::generators::v1beta1 as native,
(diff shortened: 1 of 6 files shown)


src/rust/rust-proto-new/src/graplinc/grapl/api/plugin_sdk/generators/v1beta1_server.rs
```diff
@@ -0,0 +1,166 @@
...
+        request: Request<proto::RunGeneratorRequest>,
+    ) -> Result<Response<proto::RunGeneratorResponse>, tonic::Status> {
+        execute_rpc!(self, request, run_generator)
+    }
+}
+
+/**
+ * !!!!! IMPORTANT !!!!!
+ * This is almost entirely cargo-culted from PipelineIngressServer.
+ * Lots of opportunities to deduplicate and simplify.
```
jgrillo-grapl, line 86: Are there still opportunities to deduplicate and simplify? I think it would be a good thing to write those up as github issues. It's not clear to me what future changes this comment suggests.

jgrillo-grapl (approved):
LGTM

src/rust/rust-proto-new/src/graplinc/grapl/api/plugin_sdk/generators/v1beta1_server.rs
wimax-grapl: I mean, I think we're one or two things away from being able to stop copy-pasting the <T,H,F> stuff - like we are on the verge of writing that once and injecting any Api trait into it - but my rust type knowledge just isn't quite there.

wimax-grapl: In fact, it looks like we may no longer need the PhantomData<F>?

wimax-grapl: but yeah i'll file an issue

PR #1706, opened 2022-05-09
Set up Vault and use Vault-generated certs

ghost:
https://github.com/grapl-security/issue-tracker/issues/747

This sets up vault, and then sets consul to use vault-generated certificates.

This currently works without policies since we're using -dev for consul and vault. For prod we'll need to set up policies first.

Unfortunately there's some extra junk around polling/sleeping since this change does make it longer for all the services to boot up

e2e and integration tests. make up also works


nomad/local/vault-agent-conf.hcl
```diff
@@ -0,0 +1,18 @@
+log_level = "Info"
+log_requests_level = "info"
+ui = true
+
+api_addr = "http://127.0.0.1:8200"
+
+#service_registration "consul" {
```
twunderlich-grapl, line 7: ugh, so there's a circular dependency here. We need vault to configure consul, but that means we can't use the service_registration stanza here since that expects consul to be ready immediately 💩 
I'm looking for cli commands to do this now

nomad/local/start_detach.sh
```diff
@@ -54,31 +55,61 @@ ensure_valid_env() {
...
         -config="${THIS_DIR}/nomad-agent-conf.nomad" \
         -dev-connect > "${NOMAD_LOGS_DEST}" &
     local -r nomad_agent_pid="$!"
+
+    # Wait for vault to finish booting up
+    sleep 10
+
+    configure_vault
+    create_dynamic_consul_config
+
```
wimax-grapl, line 112: Is there a more robust way to tell if Vault is ready?

wimax-grapl (approved):

nomad/local/start_detach.sh
```diff
@@ -65,24 +74,82 @@ create_dynamic_consul_config() {
...
+    address = "http://127.0.0.1:8200"
+    token = "$VAULT_TOKEN"
+    root_pki_path = "connect-root"
+    intermediate_pki_path = "connect-intermediate"
+  }
+}
 EOF
 }
 
+clear_hashicorp_log_files() {
```
twunderlich-grapl, line 90: Guess who was trying to debug based on logs from some other build? 😠

etc/ci_scripts/dump_artifacts/nomad_artifacts.py
```diff
@@ -53,9 +54,15 @@ def dump_all(artifacts_dir: Path, dump_agent_logs: bool) -> None:
...
-    shutil.copy2(consul_agent_log_path, artifacts_dir)
+def _dump_hashicorp_agent_logs(artifacts_dir: Path) -> None:
+    hashicorp_log_paths = [
+        nomad_agent_log_path,
+        consul_agent_log_path,
+        vault_agent_log_path,
+    ]
+    for log_path in hashicorp_log_paths:
+        if os.path.exists(log_path):
+            shutil.copy2(log_path, artifacts_dir)
```
twunderlich-grapl, line 65: If something fails prior to nomad starting, you better believe I want the logs

christophermaier (approved):

nomad/local/start_detach.sh
```diff
@@ -65,24 +74,83 @@ create_dynamic_consul_config() {
...
+        timeout --foreground "${wait_secs}" bash -c -- "$(
+            cat << EOF
+                # We need to allow non-zero exit codes for vault status
+                set +e
+                # General rule: Variable defined in this EOF? Use \$
+                wait_attempt=1
+                # vault status returns an exit code of 0 for unsealed (ie ready), 1 for error and 2 for sealed
+                # Since we only want to capture the exit code, we need to drop all output from the command
+                vault_exit_code=\$(vault status &>/dev/null; echo $?)
+                while [[  \${vault_exit_code} != 0 ]]; do
```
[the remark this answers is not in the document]
twunderlich-grapl: Thanks! I was doing all that extra work to avoid the script exiting, but after running set +x that didn't matter. Fixed

PR #1708, opened 2022-05-11
DONTREVIEWYET: Change PluginRegistry::create_plugin to a streaming RPC

wimax-grapl:
none

create_plugin needs to be able to transfer up to a 250MB binary as the plugin.
This isn't really well-supported by any of our infrastructure;
- protobuf supports many "small" (in the eye of the beholder) messages on the wire, but doesn't play well with large ones.
- the envoy proxy times out during a single long request

I started toying with this and it turns out Tonic makes this very, very easy.

CI right here right now

### DIFF
--- a/src/rust/rust-proto-new/src/graplinc/grapl/api/plugin_registry/v1beta1.rs
+++ b/src/rust/rust-proto-new/src/graplinc/grapl/api/plugin_registry/v1beta1.rs
@@ -228,7 +228,7 @@ impl CreatePluginRequest {
         // `move` so we can capture base_proto_request
         let chunked_requests = plugin_artifact_byte_chunks.map(move |byte_chunk| {
             let mut chunk_proto_request = base_proto_request.clone();
-            chunk_proto_request.plugin_artifact = byte_chunk.to_owned();
+            chunk_proto_request.plugin_artifact = byte_chunk;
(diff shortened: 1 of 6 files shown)


src/rust/rust-proto-new/src/graplinc/grapl/api/plugin_registry/v1beta1.rs
```diff
@@ -173,6 +176,66 @@ impl ProtobufSerializable for CreatePluginRequest {
...
+    ) -> impl tonic::IntoStreamingRequest<Message = proto::CreatePluginRequest> {
+        // Each request in the stream will have the same non-artifact fields.
+        // Why? Simplicity of implementation.
+        let base_proto_request = proto::CreatePluginRequest::from(Self {
+            tenant_id: self.tenant_id,
+            display_name: self.display_name,
+            plugin_type: self.plugin_type,
+            plugin_artifact: vec![],
+        });
+        // Split the artifact into 4MB chunks. 4MB chosen arbitrarily.
```
christophermaier, line 226: Minor nitpick: since the `chunk_size` is passed as a parameter, it doesn't make sense to speak about it being a particular size (i.e., 4 MB) in this internal comment.

PR #1711, opened 2022-05-11
devbox-sync-back: claw back specified files from devbox

wimax-grapl:
My laptop's `format-prettier` stopped working, so I cooked up a script that will let me:
- format-prettier on my devbox
- rsync a specified set of files back from my devbox to my host OS

the rsync command is heavily based on / shared with the one in devbox-sync.sh but there's enough disparity to separate them.


inickles-grapl (approved):
lgtm

PR #1712, opened 2022-05-12
[sysmon-parser] Add support for alternate format of empty Data elements and enhanced error reporting

inickles-grapl:

The sysmon-parser only expected empty Data elements to use the form of start-tag immediately followed by an end-tag (`<Data Name="RuleName"></Data>`), as this is the only format we've seen (up to this point) Sysmon use. However, we've seen a pretty-print operation change to use the alternate format for this: empty-element tag (`<Data Name="RuleName" />`), which does not parse. The situation was made worse by incorrect error messages due the parser being out of sync with the expected format.

This PR improves flexibility by handling both formats for empty Data elements.

Additionally, this adds explicit checks for the format it expects and provides better errors when what's expected is not found.

I tested the parser using the original input that surfaced the issues ( with the empty-element tag). I also wrote unit tests for this alternate format, as well as tests for error cases.


src/rust/sysmon-parser/src/util.rs
```diff
@@ -184,52 +184,156 @@ pub(crate) fn unescape_xml<'a, 'b: 'a>(span: &'a StrSpan<'b>) -> Result<Cow<'b,
...
+///
+/// To be called after the tokenizer reaches the start of a `Data` element.
+pub(crate) fn get_data_name_value<'a>(
+    tokenizer: &mut xmlparser::Tokenizer<'a>,
+    position: usize,
+) -> Result<(&'a str, Option<StrSpan<'a>>)> {
+    let name = match tokenizer.next() {
+        Some(Ok(xmlparser::Token::Attribute { local, value, .. })) if local.as_str() == "Name" => {
+            value.as_str()
+        }
```
wimax-grapl, line 198: TIL match statements can have if statements

src/rust/sysmon-parser/src/util.rs
```diff
@@ -184,52 +184,156 @@ pub(crate) fn unescape_xml<'a, 'b: 'a>(span: &'a StrSpan<'b>) -> Result<Cow<'b,
...
+            text
         }
-        match result {
-            Some(r) => r,
-            None => continue,
+        _ => {
+            return Err(Error::ParseEventDataXML {
+                message: "unexpected XML token".to_string(),
+                position,
+            })
```
wimax-grapl, line 264: (FWIW it looks like all these messages are always &'static strs; you could change the error signature and save an alloc; but NBD)

src/rust/sysmon-parser/src/util.rs
```diff
@@ -184,52 +184,156 @@ pub(crate) fn unescape_xml<'a, 'b: 'a>(span: &'a StrSpan<'b>) -> Result<Cow<'b,
...
-            match token? {
-                Token::Attribute { local, value, .. } => match local.as_str() {
-                    "Name" => result = Some(value.as_str()),
-                    _ => {}
-                },
-                Token::ElementEnd { .. } => break,
-                _ => {}
-            }
+/// Can be used to retrieve XML text from an element only text is expected next from the tokenizer,
+/// no other elements or attributes.
```
wimax-grapl, line 272: stops making grammatical sense around 'only'

src/rust/sysmon-parser/src/util.rs
```diff
@@ -184,52 +184,156 @@ pub(crate) fn unescape_xml<'a, 'b: 'a>(span: &'a StrSpan<'b>) -> Result<Cow<'b,
...
-pub(crate) use next_text_str_span;
+pub(crate) use unwrap_or_continue;
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use super::{
+        get_data_name_value,
+        *,
+    };
```
wimax-grapl, line 336: 🤔

wimax-grapl (approved):
Approved with some minor comments

src/rust/sysmon-parser/src/util.rs
inickles-grapl: 🤔

src/rust/sysmon-parser/src/util.rs
inickles-grapl: "match guards" 👍

src/rust/sysmon-parser/src/util.rs
inickles-grapl: Hmmm yeah. I definitely would if this were the happy path, I think I just might for this as well.

src/rust/sysmon-parser/src/util.rs
inickles-grapl: blargh

src/rust/sysmon-parser/src/util.rs
inickles-grapl: Okay yeah EZ

src/rust/sysmon-parser/src/util/eventdata_iterator.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+impl<'a, 'input: 'a> Iterator for EventDataIterator<'a, 'input> {
+    type Item = Result<(&'input str, xmlparser::StrSpan<'input>)>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // If the last item was an error short-circuit and fuse
+        if self.previous_error {
+            return None;
+        }
```
wimax-grapl, line 83: dont love this hack but can't tihnk of anything better :)

wimax-grapl (approved):

PR #1714, opened 2022-05-12
Optimize scylla nomad setup

ghost:
https://github.com/grapl-security/issue-tracker/issues/936

- Switch scylladb to single-node mode
- Set up health check for scylla
- Set up a data volume
- Remove static ports that aren't necessary 

make up

### DIFF
--- a/nomad/local/grapl-local-infra.nomad
+++ b/nomad/local/grapl-local-infra.nomad
@@ -555,11 +555,34 @@ job "grapl-local-infra" {
 
       config {
         image = "scylladb-ext:${var.image_tag}"
+        args = [
+          # Set up scylla in single-node mode instead of in overprovisioned mode, ie DON'T use all available cpu/memory
+          "--smp", "1"
+        ]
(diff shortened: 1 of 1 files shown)


nomad/local/grapl-local-infra.nomad
```diff
@@ -555,11 +554,52 @@ job "grapl-local-infra" {
...
+          name = "nodestatus_check"
+          # We use bin/bash so we can pipe to grep
+          command  = "bin/bash"
+          args     = ["nodetool", "status", "|", "grep", "'UN'"]
+          interval = "30s"
+          timeout  = "10s"
+
+          check_restart {
+            # Wait up to 4 minutes for scylla to be ready, ie this is a readiness check
+            grace           = "4m"
```
colin-grapl, line 596: 4m seems high, would suggest 1m

twunderlich-grapl: Yep, this is a holdover from prior to switching to single node mode

jgrillo-grapl (approved):
👍

PR #1715, opened 2022-05-12
Change all usage of structopt to clapv3

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/868

Change from structopt to clapv3. Structopt is in maintenance mode and suggests people use this instead.

One major note: adding `long` along with `env` tells clap to stop doing some unnecessary-for-us command line specific validation, which we just don't care about. 
Including `long` is not necessary in much of our code, _but_ is a pattern I want to cargo-cult to future consumers (think of it as a good habit).

CI


jgrillo-grapl (approved):
LGTM pending successful CI build

PR #1724, opened 2022-05-18
Add correct perms to plugin_registry bucket; clean up naming around it in Pulumi, Nomad, rust

wimax-grapl:
> [4:06](https://grapl-internal.slack.com/archives/C017F10727P/p1652904407194439)
```
    plugins_bucket = Bucket("plugins-bucket", sse=True)
    pulumi.export("plugins-bucket", plugins_bucket.bucket)

    plugin_buckets = [
        analyzers_bucket,
        model_plugins_bucket,
    ]
the variable plugins_bucket is not in plugin_buckets
```
> [cm] [4:07 PM](https://grapl-internal.slack.com/archives/C017F10727P/p1652904470386639)
> That's a thorny bit of code from a naming perspective... plugins_bucket, model_plugins_bucket, plugin_buckets

### DIFF
--- a/pulumi/grapl/__main__.py
+++ b/pulumi/grapl/__main__.py
@@ -189,9 +189,9 @@ def main() -> None:
     engagement_creator_queue.subscribe_to_emitter(analyzer_matched_emitter)
 
     analyzers_bucket = Bucket("analyzers-bucket", sse=True)
+    pulumi.export("analyzers-bucket", analyzers_bucket.bucket)
     model_plugins_bucket = Bucket("model-plugins-bucket", sse=False)
     plugin_registry_bucket = Bucket("plugin-registry-bucket", sse=True)
-    pulumi.export("analyzers-bucket", analyzers_bucket.bucket)
 
     all_plugin_buckets = [
         analyzers_bucket,
--- a/pulumi/grapl/__main__.py
+++ b/pulumi/grapl/__main__.py
@@ -189,9 +189,15 @@ def main() -> None:
     engagement_creator_queue.subscribe_to_emitter(analyzer_matched_emitter)
 
     analyzers_bucket = Bucket("analyzers-bucket", sse=True)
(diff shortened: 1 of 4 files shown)


pulumi/grapl/__main__.py
```diff
@@ -189,27 +189,25 @@ def main() -> None:
     engagement_creator_queue.subscribe_to_emitter(analyzer_matched_emitter)
 
     analyzers_bucket = Bucket("analyzers-bucket", sse=True)
-    pulumi.export("analyzers-bucket", analyzers_bucket.bucket)
     model_plugins_bucket = Bucket("model-plugins-bucket", sse=False)
-    pulumi.export("model-plugins-bucket", model_plugins_bucket.bucket)
```
twunderlich-grapl, line 194: Did you mean to delete this export?

christophermaier, line 194: Is this bucket not exported anymore because it's "old Grapl"?

pulumi/grapl/__main__.py
```diff
@@ -189,27 +189,25 @@ def main() -> None:
...
+    plugin_registry_bucket = Bucket("plugin-registry-bucket", sse=True)
```
christophermaier, line 193: Are we not exporting this bucket?

pulumi/grapl/__main__.py
```diff
@@ -189,27 +189,25 @@ def main() -> None:
...
+    pulumi.export("analyzers-bucket", analyzers_bucket.bucket)
```
christophermaier, line 194: If we're continuing to export this bucket, it would be better to move the export closer to where we created the bucket, like it was before.

christophermaier (approved):

pulumi/grapl/__main__.py
wimax-grapl: Upon searching, these exported buckets are never used; somebody was just cargo-culting it. 

If we don't use the export, we shouldn't export it.

PR #1725, opened 2022-05-19
Make 'make test-with-env' not print out a bunch of useless commands; make 'dump_artifacts' fail gracefully

wimax-grapl:
1. test-with-env printed out a bunch of useless commands; `.SILENT` quiets them.

2. `dump_artifacts` would print out a stack trace when grapl-core/ didn't exist for analysis, gross!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -331,7 +331,7 @@ test-e2e: ## Build and run e2e tests
 # Think of it as a Context Manager that:
 # - Before test-time, brings up a `make up`
 # - After test-time, tears it all down and dumps artifacts.
-.SILENT:
+.SILENT: test-with-env
 .PHONY: test-with-env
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -331,6 +331,7 @@ test-e2e: ## Build and run e2e tests
 # Think of it as a Context Manager that:
 # - Before test-time, brings up a `make up`
 # - After test-time, tears it all down and dumps artifacts.
+.SILENT:
```
christophermaier, line 334: You need to add the name of the target you want to silence as a prerequisite, as with `.PHONY`. As written, I believe this silences _everything_.

christophermaier (approved):
👍 once the `.SILENT` thing is fixed

PR #1726, opened 2022-05-19
Streaming CreatePlugin

wimax-grapl:
Satisfies https://github.com/grapl-security/issue-tracker/issues/937
Seemingly resolves root cause of https://github.com/grapl-security/grapl/pull/1716

We now stream 1MB chunks of the plugin artifact into S3, and hold up to 8 1MB messages at a time on the server side.


src/rust/plugin-registry/tests/test_deploy_plugin.rs
```diff
@@ -12,32 +15,38 @@ use rust_proto_new::graplinc::grapl::api::plugin_registry::v1beta1::{
...
 
-// Temporarily skipping this test due to CI failures that I suspect are due to
-// the lack of streaming. Tackling this is my next task, but I want to unblock
-// other developers for the time being.
-// https://github.com/grapl-security/issue-tracker/issues/937
-#[ignore]
+async fn get_client() -> Result<PluginRegistryServiceClient, PluginRegistryServiceClientError> {
+    // For some reason, I'm seeing nondeterministic failures when initializing
+    // a client.
+    // Suspicions:
```
wimax-grapl, line 25: This predates the streaming btw

colin-grapl (changes requested):

src/proto/graplinc/grapl/api/plugin_registry/v1beta1/plugin_registry.proto
```diff
@@ -29,19 +29,24 @@ message Plugin {
   bytes plugin_binary = 4;
 }
 
-// A request to create a plugin entry
 message CreatePluginRequest {
-  // the actual plugin code
-  bytes plugin_artifact = 1;
+  oneof inner {
+    CreatePluginRequestMetadata metadata = 1;
+    // the actual plugin binary
```
colin-grapl, line 35: I think this comment is out of date.

src/rust/grapl-utils/src/future_ext/mod.rs
```diff
@@ -25,3 +25,32 @@ pub trait GraplFutureExt: Future {
...
+
+#[async_trait::async_trait]
+pub trait GraplAsyncRetry<Fut, T, E>: Fn() -> Fut
+where
+    Fut: Future<Output = Result<T, E>> + Send,
+    E: std::error::Error + Send + Sync + 'static,
+{
+    /// Helper method to auto-retry an async () -> Result.
+    /// If it fails after `num_retries + 1` times, return the Err.
+    async fn retry(&self, num_retries: u8) -> Result<T, E> {
```
colin-grapl, line 37: fwiw elsewhere we use `futures-retry` for this sort of thing (graph-merger), but nbd

src/rust/plugin-registry/src/error.rs
```diff
@@ -65,7 +65,9 @@ impl From<PluginRegistryServiceError> for Status {
             Error::NomadJobAllocationError => {
                 Status::internal("Unable to allocate Nomad job - it may be out of resources.")
             }
-            Error::ArtifactTooLargeError(msg) => Status::invalid_argument(msg),
+            Error::StreamInputError(_) => {
```
colin-grapl, line 68: Seems like we could format the value in the enum into the message? dunno if it would help

src/rust/plugin-registry/src/server/service.rs
```diff
@@ -109,16 +119,26 @@ pub struct PluginRegistry {
...
-        } else {
-            Ok(())
+fn io_input_error(input: impl std::fmt::Display) -> std::io::Error {
+    std::io::Error::new(std::io::ErrorKind::InvalidInput, input.to_string())
+}
+
+/// Keep adding the size of an incoming chunk to an AtomicUsize.
+/// Start throwing errors when it exceeds limit.
+async fn accumulate_stream_size_to_limit(
+    stream_length: Arc<AtomicUsize>,
```
colin-grapl, line 129: I suspect you can just take `stream_length: &AtomicUsize`, and then call `.as_ref()` on the Arc. NBD but saves an atomic increment. I haven't seen the calling clone yet so I could be wrong.

src/rust/plugin-registry/src/server/service.rs
```diff
@@ -109,16 +119,26 @@ pub struct PluginRegistry {
...
-            )))
-        } else {
-            Ok(())
+fn io_input_error(input: impl std::fmt::Display) -> std::io::Error {
+    std::io::Error::new(std::io::ErrorKind::InvalidInput, input.to_string())
+}
+
+/// Keep adding the size of an incoming chunk to an AtomicUsize.
+/// Start throwing errors when it exceeds limit.
+async fn accumulate_stream_size_to_limit(
```
colin-grapl, line 128: This function doesn't look like it calls `.await` anywhere internally, so I don't think it needs to be async.

src/rust/plugin-registry/src/server/service.rs
```diff
@@ -129,32 +149,76 @@ impl PluginRegistryApi for PluginRegistry {
...
         let s3_key = generate_artifact_s3_key(plugin_type, &tenant_id, &plugin_id);
 
+        // Convert the incoming request stream of CreatePluginRequest::Chunk
+        // into a stream of Bytes to be sent to Rusoto S3.put_object.
+        let body_stream = request.then(|result| async {
+            match result {
+                CreatePluginRequest::Chunk(c) => Ok(Bytes::from(c)),
+                _ => Err(io_input_error("Expected request 1..N to be Chunk")),
+            }
+        });
```
colin-grapl, line 181: I think ideally we'd do this via S3 multipart uploads, that way we can drop the data from memory, but I would prefer that we do it this way for now and just add a ticket for that later.

src/rust/plugin-registry/src/server/service.rs
colin-grapl: Based on the caller it might end up being a pain actually, so nvm, not worth it.

src/rust/plugin-registry/src/error.rs
wimax-grapl: This is the over-the-wire message, which we usually want to anonymize. In this case it might make sense

src/rust/plugin-registry/src/server/service.rs
```diff
@@ -129,32 +149,76 @@ impl PluginRegistryApi for PluginRegistry {
...
+                .into_inner();
+            let total_duration = std::time::SystemTime::now()
+                .duration_since(start_time)
+                .unwrap_or_default();
+
+            tracing::info!(
+                message = "CreatePlugin benchmark",
+                display_name = ?display_name,
+                duration_millis = ?total_duration.as_millis(),
+                size_mb = stream_length / (1024 * 1024)
```
colin-grapl, line 218: For a debug log I think it'd make more sense to just put `stream_length` in there directly rather than divide

src/rust/plugin-registry/src/server/service.rs
wimax-grapl: It does! In fact, I'm pretty sure the rust linter doesn't allow you to declare an un-consumed future.

However, the mechanism is unconventional.

```
        let body_stream = body_stream.then(move |result| {
            accumulate_stream_size_to_limit(stream_length_handle.clone(), limit_bytes, result)
        });
```

this replaces a `.then(move |result| async move { })` 
and `then` takes an F: [FnMut](https://doc.rust-lang.org/nightly/core/ops/function/trait.FnMut.html)(<St as [Stream](https://docs.rs/futures/latest/futures/stream/trait.Stream.html)>::[Item](https://docs.rs/futures/latest/futures/stream/trait.Stream.html#associatedtype.Item)) -> Fut

src/rust/plugin-registry/src/server/service.rs
wimax-grapl: Rusoto automatically does that for us! You pass it a Stream.

wimax-grapl: wow jk i think i must have made that up

src/rust/grapl-utils/src/future_ext/mod.rs
wimax-grapl: done

src/rust/plugin-registry/src/server/service.rs
wimax-grapl: i changed it to use multipart

src/rust/plugin-registry/src/server/service.rs
wimax-grapl: done

src/rust/rust-proto-new/src/graplinc/grapl/api/plugin_registry/v1beta1_server.rs
```diff
@@ -94,9 +96,43 @@ where
...
+                    tx.send(req)
+                        .await
+                        .map_err(|e| Status::internal(e.to_string()))?;
+                }
+                Ok(())
+            } as Result<(), Status>)
+        };
+
+        let api_handler_thread = async move {
+            ({ self.api_server.create_plugin(rx).await.map_err(Into::into) }
```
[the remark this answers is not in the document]
wimax-grapl: the trait bound `protocol::status::Status: From<<T as PluginRegistryApi>::Error>` is not satisfied
required because of the requirements on the impl of `FromResidual<Result<Infallible, <T as PluginRegistryApi>::Error>>` for

src/rust/rust-proto-new/src/graplinc/grapl/api/plugin_registry/v1beta1_client.rs
```diff
@@ -32,21 +40,39 @@ impl PluginRegistryServiceClient {
...
+    /// Simplified wrapper function for create_plugin
+    pub async fn create_plugin(
+        &mut self,
+        metadata: native::CreatePluginRequestMetadata,
+        plugin_artifact: impl Sized + Iterator<Item = u8> + Send + 'static,
+    ) -> Result<native::CreatePluginResponse, PluginRegistryServiceClientError> {
+        // Split the artifact up into 5MB chunks
+        let plugin_chunks = plugin_artifact.chunks_owned(1024 * 1024 * 5);
+        // Send the metadata first followed by N chunks
+        let request = futures::stream::iter(vec![native::CreatePluginRequest::Metadata(metadata)])
```
colin-grapl, line 69: There's an 'iter::Once' for this sort of thing I think

colin-grapl (approved):

PR #1728, opened 2022-05-19
Fix bug where we weren't rebuilding rust-proto-new when protos were changed

colin-grapl:
Experimental change to the rust-proto-new build script to have it rebuild more aggressively.


wimax-grapl (changes requested):
as discussed in slack, the problem here is that the fs walk is not recursive

wimax-grapl (approved):

PR #1731, opened 2022-05-20
Ensure Consul agent is up before Nomad starts

wimax-grapl:
[wimax](https://app.slack.com/team/U0175EMAA5U)  [2:07 PM](https://grapl-internal.slack.com/archives/C02JCBKS97V/p1653070056555209)

I'm seeing CI having trouble with nomad_local_infrastructure. Not 100% sure why, but I just found the following in the logs:
```
    2022-05-20T15:44:20Z: Task Group "plugin-work-queue-db" (failed to place 1 allocation):
      * Constraint "${attr.consul.version} semver >= 1.7.0": 1 nodes excluded by filter
    2022-05-20T15:44:20Z: Task Group "organization-management-db" (failed to place 1 allocation):
      * Constraint "${attr.consul.version} semver >= 1.7.0": 1 nodes excluded by filter
etc etc
```

[wimax](https://app.slack.com/team/U0175EMAA5U)  [2:45 PM](https://grapl-internal.slack.com/archives/C02JCBKS97V/p1653072305270219)
i can repro locally after updating to nomad 1.3.0
it appears to be nondeterministic
as such, my suspicion is that the nomad agent expects the consul agent to be fully up-and-ready before you deploy anything to it, and we don't have any explicit check for that
[2:45](https://grapl-internal.slack.com/archives/C02JCBKS97V/p1653072346751579)
and this expectation is new as of nomad 1.3 (edited)

Relates to https://github.com/hashicorp/nomad/issues/12958

### DIFF
--- a/nomad/local/start_detach.sh
+++ b/nomad/local/start_detach.sh
@@ -86,7 +86,31 @@ start_nomad_detach() {
         -dev > "${CONSUL_LOGS_DEST}" &
     local -r consul_agent_pid="$!"
 
-    # Wait a short period of time before attempting to deploy infrastructure
+    # Ensure Consul agent is up and running
+    (
+        readonly wait_secs=45
+        # shellcheck disable=SC2016
+        timeout --foreground "${wait_secs}" bash -c -- "$(
+            cat << EOF
+                # General rule: Variable defined in this EOF? Use \$
+                set -euo pipefail
+                wait_attempt=1
+                while [[ -z \$(consul info 2>&1 | grep "leader = true") ]]; do
+                    if ! ps -p "${consul_agent_pid}" > /dev/null; then
+                        echo "Consul Agent unexpectedly exited?"
+                        exit 42
+                    fi
+
+                    echo "Waiting for consul-agent [\${wait_attempt}/${wait_secs}]"
+                    sleep 1
+                    ((wait_attempt=wait_attempt+1))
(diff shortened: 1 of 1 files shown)


nomad/local/start_detach.sh
```diff
@@ -73,22 +73,48 @@ start_nomad_detach() {
...
+                # General rule: Variable defined in this EOF? Use \$
+                set -euo pipefail
+                wait_attempt=1
+                while [[ -z \$(consul info 2>&1 | grep "leader = true") ]]; do
+                    if ! ps -p "${consul_agent_pid}" > /dev/null; then
+                        echo "Consul Agent unexpectedly exited?"
+                        exit 42
+                    fi
+
+                    echo "Waiting for consul-agent [\${wait_attempt}/${wait_secs}]"
```
christophermaier, line 100: This is a definite nitpick, but this formulation conflates the number of attempts with how long all of them take, which is a little confusing, just from a pure code reading standpoint.

`wait_secs` could be renamed to something like `total_attempts` to address this.

christophermaier (approved):

nomad/local/start_detach.sh
wimax-grapl: I'm going to commit it as-is to unblock people asap but fix this in an unrelated followup PR

PR #1737, opened 2022-05-21
Rust integration tests now do not capture stdout even if they pass

wimax-grapl:
Currently, integration tests will just swallow all their output if they succeed.

This verbosity can help with debugging.

Relates to https://github.com/clap-rs/clap/issues/3741


wimax-grapl:
Actually, gonna put this on ice for a bit, but it'll eventually come back

PR #1738, opened 2022-05-23
Skip consul dns test

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/944

Skip the consul dns test.
There's a fix for this behavior in an upcoming Nomad release, but this
is currently causing e2e tests to fail in AWS. Since this is coming
soon, we should skip the test for now

Automated tests

### DIFF
--- a/src/python/e2e-test-runner/e2e_test_runner/test_consul_dns.py
+++ b/src/python/e2e-test-runner/e2e_test_runner/test_consul_dns.py
@@ -3,7 +3,11 @@ Smoke test to ensure that consul dns is available from within the service mesh
 """
 import socket
 
+import pytest
 
+
+# TODO re-enable this after the fix to Nomad DNS is released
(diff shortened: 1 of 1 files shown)


src/python/e2e-test-runner/e2e_test_runner/test_consul_dns.py
```diff
@@ -3,7 +3,11 @@
 """
 import socket
 
+import pytest
 
+
+# TODO re-enable this after the fix to Nomad DNS is released
```
christophermaier, line 9: A link to an issue, or indication of expected Nomad version would be useful here.

christophermaier (approved):

src/python/e2e-test-runner/e2e_test_runner/test_consul_dns.py
twunderlich-grapl: Added. The fix is in main, just waiting for the next version release of Nomad

PR #1741, opened 2022-05-24
Update internal error codes

ghost:
We've been using error code 42 for all internal errors. Unfortunately,
that makes tracking down what the actual error is more difficult.

N/A


christophermaier (approved):
Any thoughts on adding some kind of central "catalog" of exit codes, just to make it easier to see which codes have been used already, and which ones are free?

twunderlich-grapl:
> Any thoughts on adding some kind of central "catalog" of exit codes, just to make it easier to see which codes have been used already, and which ones are free?

I was actually in the process of putting that into a markdown file when you mentioend that.

colin-grapl (approved):

PR #1700, opened 2022-05-08
Add skeleton for uid-allocator

colin-grapl:
### Which issue does this PR correspond to?
913

### What changes does this PR make to Grapl? Why?
Service skeleton for the uid-allocator
as defined in https://github.com/grapl-security/grapl-rfcs/blob/main/text/0004-graph.md

### How were these changes tested?
This is just the basic skeleton of the allocator.

PR #1744, opened 2022-05-30
Uid allocator base

colin-grapl:
913

There was a previous PR that went stale. https://github.com/grapl-security/grapl/pull/1700

Adds the base for the uid-allocator service. This is just the boilerplate stuff like getting a service definition there, but contains no business logic - that will come in a future PR.

Existing tests, that's it. Again, this PR is only meant to stage the next PR.


pulumi/grapl/__main__.py
```diff
@@ -368,11 +369,21 @@ def main() -> None:
             port=5532,
         )
 
+        uid_allocator_db = LocalPostgresInstance(
+            name="uid-allocator-db",
+            port=5532,
+        )
```
wimax-grapl, line 375: Okay, lot to unpack here.

These port numbers should be unique, but it's the same one as plugin-work-queue.

So - what's actually happening at runtime is, we're using the same 1 postgres instance for both of these different services.

Personally? For local development? I think that's completely fine and okay, and that we should have 1 postgres instance for everything in local (and N postgres instances for everything in prod).

pulumi/grapl/__main__.py
```diff
@@ -544,6 +567,17 @@ def main() -> None:
...
+        pulumi.export("uid-allocator-db-port", uid_allocator_postgres.port().apply(str))
+        pulumi.export(
+            "uid-allocator-db-username",
+            uid_allocator_postgres.username(),
+        )
+        pulumi.export(
+            "uid-allocator-db-password",
+            uid_allocator_postgres.password(),
+        )
+
```
wimax-grapl, line 580: You don't need to export these unless they're used in integration tests (they're not)

pulumi/grapl/__main__.py
```diff
@@ -368,11 +369,21 @@ def main() -> None:
...
         pulumi.export("plugin-work-queue-db-hostname", plugin_work_queue_db.hostname)
         pulumi.export("plugin-work-queue-db-port", str(plugin_work_queue_db.port))
         pulumi.export("plugin-work-queue-db-username", plugin_work_queue_db.username)
         pulumi.export("plugin-work-queue-db-password", plugin_work_queue_db.password)
 
+        pulumi.export("uid-allocator-db-hostname", uid_allocator_db.hostname)
+        pulumi.export("uid-allocator-db-port", str(uid_allocator_db.port))
+        pulumi.export("uid-allocator-db-username", uid_allocator_db.username)
+        pulumi.export("uid-allocator-db-password", uid_allocator_db.password)
+
```
wimax-grapl, line 386: You don't need to export these unless they're used in integration tests (they're not)

wimax-grapl (changes requested):

pulumi/grapl/__main__.py
colin-grapl: They will be though post base.

wimax-grapl (approved):

PR #1747, opened 2022-06-01
Strip debuginfo in default local-grapl rust builds.

wimax-grapl:
Introduced a new Cargo build profile, `dev-local-grapl`, that's basically the same as debug but without debuginfo.
- Symbols are still there.
- But you basically can't use `gdb` on it.
You can get debuginfo back by running with RUST_BUILD=debug.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # required for our local usage of Nomad, because Nomad won't resolve a
 # `latest` tag from the host machine.)
 IMAGE_TAG ?= dev
-RUST_BUILD ?= debug
+RUST_BUILD ?= dev-local-grapl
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -13,7 +13,7 @@
 # required for our local usage of Nomad, because Nomad won't resolve a
 # `latest` tag from the host machine.)
 IMAGE_TAG ?= dev
-RUST_BUILD ?= debug
+RUST_BUILD ?= dev-local-grapl
```
wimax-grapl, line 16: why do we define this default in 3 separate places lol

jgrillo-grapl (approved):

PR #1750, opened 2022-06-03
Simplify postgres database stuff by passing it around Pulumi/Nomad as a structured Object

wimax-grapl:
Where once we had 4 Nomad vars per database, now we have 1.

That also means we can `pulumi.export` just 1 thing.


jgrillo-grapl (approved):

christophermaier (approved):
Nice 👍

PR #1754, opened 2022-06-05
Generator dispatcher - reads from Kafka raw_logs topic, writes to Plugin Work Queue

wimax-grapl:
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/714
https://github.com/grapl-security/issue-tracker/issues/714

Adds a new service that reads from raw_logs and writes them as Plugin Work Queue jobs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -461,7 +461,7 @@ _up:
 		--file docker-compose.yml \
 		up --force-recreate --always-recreate-deps --renew-anon-volumes \
 		--exit-code-from pulumi \
-		pulumi
+		pulumi && echo "✅ Pulumi deployed successfully ✅"
 
 .SILENT: down
 .PHONY: down
--- a/Makefile
+++ b/Makefile
@@ -471,8 +471,8 @@ down: ## docker compose down - both stops and removes the containers
 	# about. This must be the network that is used in Localstack's
 	# LAMBDA_DOCKER_NETWORK environment variable.
 	$(MAKE) stop-nomad-detach
-	docker-compose $(EVERY_COMPOSE_FILE) down --timeout=0
-	@docker-compose $(EVERY_COMPOSE_FILE) --project-name $(COMPOSE_PROJECT_INTEGRATION_TESTS) down --timeout=0
+	docker compose $(EVERY_COMPOSE_FILE) down --timeout=0
+	@docker compose $(EVERY_COMPOSE_FILE) --project-name $(COMPOSE_PROJECT_INTEGRATION_TESTS) down --timeout=0
 
 .PHONY: stop
 stop: ## docker compose stop - stops (but preserves) the containers
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -623,6 +623,45 @@ job "grapl-core" {
   ## Begin actual Grapl core services ##
   #######################################
 
+  group "generator-dispatcher" {
+    network {
+      mode = "bridge"
+      dns {
+        servers = local.dns_servers
+      }
+    }
+
+    task "generator-dispatcher" {
+      driver = "docker"
+
+      config {
+        image = var.container_images["generator-dispatcher"]
+      }
+
+      env {
+        # Upstreams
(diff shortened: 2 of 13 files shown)


Makefile
```diff
@@ -461,7 +461,7 @@ _up:
 		--file docker-compose.yml \
 		up --force-recreate --always-recreate-deps --renew-anon-volumes \
 		--exit-code-from pulumi \
-		pulumi
+		pulumi && echo "✅ Pulumi deployed successfully ✅"
```
wimax-grapl, line 464: I added this because I've been constantly annoyed by the big scary red text "ERROR EXIT CODE 0"  here 
image

Makefile
```diff
@@ -471,8 +471,8 @@ down: ## docker compose down - both stops and removes the containers
 	# about. This must be the network that is used in Localstack's
 	# LAMBDA_DOCKER_NETWORK environment variable.
 	$(MAKE) stop-nomad-detach
-	docker-compose $(EVERY_COMPOSE_FILE) down --timeout=0
-	@docker-compose $(EVERY_COMPOSE_FILE) --project-name $(COMPOSE_PROJECT_INTEGRATION_TESTS) down --timeout=0
+	docker compose $(EVERY_COMPOSE_FILE) down --timeout=0
+	@docker compose $(EVERY_COMPOSE_FILE) --project-name $(COMPOSE_PROJECT_INTEGRATION_TESTS) down --timeout=0
```
wimax-grapl, line 476: seemed to be a missed rebase thing

src/rust/generator-dispatcher/src/config.rs
```diff
@@ -0,0 +1,39 @@
...
+use rust_proto_new::SerDe;
+
+#[derive(clap::Parser, Debug)]
+pub struct GeneratorDispatcherConfig {
+    #[clap(flatten)]
+    pub kafka_config: KafkaConsumerConfig,
+}
+
+#[derive(clap::Parser, Debug)]
+pub struct KafkaConsumerConfig {
```
wimax-grapl, line 14: The idea I had here was to eventually share this struct services - I don't currently see any services that consume from more than one consumer group, hence the somewhat generic env variable names

nomad/grapl-core.nomad
```diff
@@ -153,6 +153,21 @@ variable "plugin_registry_bucket_name" {
...
+variable "generator_dispatcher_kafka_sasl_password" {
+  type        = string
+  description = "blah blah im refactoring these next PR"
+}
+
+variable "generator_dispatcher_kafka_consumer_group" {
+  type        = string
+  description = "blah blah im refactoring these next PR"
+}
+
```
wimax-grapl, line 170: these will be refactored into an Object

src/rust/Dockerfile
```diff
@@ -363,3 +363,9 @@ FROM rust-dist AS uid-allocator-deploy
 
 COPY --from=build /outputs/uid-allocator /
 ENTRYPOINT ["/uid-allocator"]
+
```
jgrillo-grapl, line 366: will also need to add `"generator-dispatcher/new_integration_tests"` to L205

src/rust/generator-dispatcher/src/config.rs
jgrillo-grapl: Yeah this can probably go into the `kafka` library, and we should refactor s.t. the constructors take this type of config object instead of a bunch of strings. I don't personally care whether that happens now or later.

jgrillo-grapl (approved):
Would be great to have an integration test which fires a payload at pipeline-ingress and makes sure the service makes the expected gRPC call. See e.g. https://github.com/grapl-security/grapl/blob/internal-368-pipeline-kafka-migration/src/rust/generators/sysmon-generator/tests/integration_test.rs

wimax-grapl:
I will do the integration test before any further feature work in a separate PR. Ordering:
1. land this
2. refactor kafka stuff because it's been bothering me
3. PR an integration test

PR #1757, opened 2022-06-07
Refactor all Kafka producers/consumers to use a standardized set of Clap-parsed env vars

wimax-grapl:
None

Refactor all Kafka producers/consumers to use a standardized set of Clap-parsed env vars.
They're encapsulated in KafkaProducerConfig and KafkaConsumerConfig.
Call ::parse on these to retrieve from env.
This was originally introduced with Graph Dispatcher; I've pulled it out to the Kafka library. 

make test-integration-new


src/rust/kafka/src/lib.rs
```diff
@@ -290,6 +324,18 @@ where
         })
     }
 
+    pub fn new_from_config(
```
jgrillo-grapl, line 324: same

jgrillo-grapl (approved):
Looks good, I think we should break the existing constructor and replace it with this new one. Not going to block on that, though.

src/rust/generators/sysmon-generator/src/main.rs
```diff
@@ -79,7 +79,7 @@ async fn handler() -> Result<(), SysmonGeneratorError> {
 
     // TODO: also construct a stream processor for retries
 
-    let stream_processor = StreamProcessor::new_from_config(
+    let stream_processor = StreamProcessor::new(
         consumer_config,
         consumer_topic,
```
jgrillo-grapl, line 84: The topics should probably go in the config object also, sorry I missed this before.

colin-grapl: Agreed. NBD / can be done later, but if we place the topic itno the config we could also change this `new_from_config` into a TryFrom<Config>

christophermaier (approved):
Nice overall cleanup 👍 

I have some minor quibbles about naming, etc., but no blockers.

src/rust/generators/sysmon-generator/tests/integration_test.rs
```diff
@@ -3,8 +3,10 @@
 use std::time::Duration;
 
 use bytes::Bytes;
+use clap::Parser;
 use futures::StreamExt;
 use kafka::{
+    config::KafkaConsumerConfig,
```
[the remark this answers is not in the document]
wimax-grapl: yeah, this is mostly a legacy from when it lived in a different crate

jgrillo-grapl (approved):
LGTM

PR #1760, opened 2022-06-08
Prevent a race condition in generator-dispatcher integration tests.

wimax-grapl:

None

Introduce a test utils extension to PsqlQueue so integration tests can
    make assertions against not-publicly-exposed data/APIs. Prevent a
    race condition between our generator-dispatcher integration tests and
    our pipeline services.

```
/// This module is meant to be used by integration tests in other crates.
///
/// ## Example use case
/// So I've introduced a new service 'generator-dispatcher' which:
/// - pops logs off raw-logs
/// - sticks them in plugin-work-queue
///
/// The reasonable integration test for this would, then, be:
/// - insert a log into raw-logs
/// - read jobs in plugin-work-queue to ensure a matching one lives in there
///
/// The problem is this:
/// - We have two concurrent and competing binaries that are calling get_execute_generator:
///   this integration test, and generator-executor.
/// - Generator-executor could potentially mark the job as 'acknowledged', hiding
///   the job from the integration test - it's a race condition.
///
/// As such, I'm exposing a test-only utilities suite that will let a
/// downstream integration test inspect jobs that are already acknowledged, too.
```

its all tests baby

### DIFF
--- a/src/rust/plugin-work-queue/src/psql_queue.rs
+++ b/src/rust/plugin-work-queue/src/psql_queue.rs
@@ -1,9 +1,15 @@
+use grapl_utils::future_ext::GraplFutureExt;
 use sqlx::{
     Pool,
     Postgres,
 };
 use tracing::instrument;
 use uuid::Uuid;
+
+use crate::{
+    server::PluginWorkQueueInitError,
+    PluginWorkQueueDbConfig,
+};
 
 #[derive(Clone, Debug, Eq, PartialEq, sqlx::Type)]
 #[sqlx(type_name = "status", rename_all = "lowercase")]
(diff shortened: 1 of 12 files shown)


src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -262,38 +286,3 @@ impl PsqlQueue {
         Ok(())
     }
 }
-
```
wimax-grapl, line 265: these were no longer consumed

christophermaier (approved):
Big fan of test-only extension traits 👍

jgrillo-grapl (approved):

PR #1765, opened 2022-06-09
Introduce a new AssociatedNamedService for rust clients. Add a health check to PluginRegistryServiceClient

wimax-grapl:
none

So it turns out my understanding of "what does Nomad do with grpc health checks?" is completely incorrect. I thought Nomad would prevent marking grapl-core as done/ready, if services were not healthy. Then, that would prevent integration services from coming up.

Apparently not. Runtime health checks are required.

This change makes it a bit more repeatable to say "hey, wait for a certain client to be Ready".

CI right here right now


src/rust/rust-proto-new/src/protocol/service_client.rs
```diff
@@ -0,0 +1,5 @@
+/// Equivalent to the NamedService trait in tonic for server constructs.
+/// Pass this NAME to e.g. a healthcheck client.
+pub trait AssociatedNamedService {
```
jgrillo-grapl, line 3: Could we instead call this `ServiceName`? `AssociatedNamedService` is a mouthful, and I'm confused about the meaning of the `Associated` part.

jgrillo-grapl (approved):
One naming nit, otherwise LGTM

wimax-grapl:
so tonic has a NamedService that applies to a Server/Service construct

so this is the associated NamedService for a client

originally i was trying to associate the client to its server but decided, blah

PR #1767, opened 2022-06-14
Add helper for SSMing to an instance type

ghost:
https://github.com/grapl-security/issue-tracker/issues/955

This adds a helper script to make it easier to ssh into an instance via ssm.

Previously I was either logging into the AWS console and looking up an instance id manually or running one of the existing ssm scripts to get the instance id and then logging out.

Run ./bin/aws/ssm_ssh.sh "Nomad Agent" to log into an instance.

### DIFF
--- a/bin/aws/lib/ssm_tools.sh
+++ b/bin/aws/lib/ssm_tools.sh
@@ -51,7 +51,7 @@ EOF
         --parameters "${SSM_PARAMETERS}"
 }
 
-ssm_ssh() {
+ssm() {
     readonly SERVER_TYPE=$1
 
     if [ -z "${AWS_PROFILE}" ]; then
--- a/bin/aws/lib/ssm_tools.sh
+++ b/bin/aws/lib/ssm_tools.sh
@@ -1,7 +1,18 @@
 #!/usr/bin/env bash
 
+_get_server_instance() {
+    # shellcheck disable=SC2140
+    SERVER_INSTANCE_ID=$(
+        aws ec2 describe-instances \
+            --filters Name=tag:Name,Values="${SERVER_TYPE}" Name="instance-state-name",Values="running" \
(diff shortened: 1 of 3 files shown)


wimax-grapl (approved):
but besides the naming thing, looks good

bin/aws/lib/ssm_tools.sh
```diff
@@ -1,7 +1,18 @@
...
 
+_get_server_instance() {
+    # shellcheck disable=SC2140
+    SERVER_INSTANCE_ID=$(
+        aws ec2 describe-instances \
+            --filters Name=tag:Name,Values="${SERVER_TYPE}" Name="instance-state-name",Values="running" \
+            --query="Reservations[0].Instances[0].InstanceId" \
+            --output=text
+    )
+    echo "${SERVER_INSTANCE_ID}"
```
wimax-grapl, line 11: we could just remove the echo and remove the subshell - i tihnk it'll output the same thing - but i'm also not opposed to the over-verbosity of this

bin/aws/lib/ssm_tools.sh
```diff
@@ -45,3 +50,21 @@ EOF
         --document-name AWS-StartPortForwardingSession \
         --parameters "${SSM_PARAMETERS}"
 }
+
+ssm_ssh() {
```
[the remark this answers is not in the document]
twunderlich-grapl: While I suspect that the SSM Manager is basically creating a ssh session internally, I do agree that its important to clarify. I've renamed the script and the function

bin/aws/lib/ssm_tools.sh
twunderlich-grapl: I'd prefer the verbosity of this for now so that its clear what's going on.

PR #1769, opened 2022-06-15
Fix pulumi errors in release pipeline

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/956

1.Fixes a deprecation warning in the pulumi RDS resource (name argument has been renamed to db_name)
2. Removes a duplicate call to the _fix_pulumi_preview method
3. Updates the _fix_pulumi_preview to only replace pulumi.Output[str] with a fake string instead of just pulumi.Output. This fixes an issue where Nomad variables with complex types (objects, and lists) would instead get a string. Understandably, this caused Nomad to error out like in https://buildkite.com/grapl/grapl-provision/builds/482#0181639c-8188-4fb4-8a7d-8596b64f34f9

pulumi preview.

### DIFF
--- a/pulumi/infra/nomad_job.py
+++ b/pulumi/infra/nomad_job.py
@@ -1,6 +1,6 @@
 import json
 from pathlib import Path
-from typing import Any, Mapping, Optional, Union, cast
+from typing import Any, Mapping, Optional, Union, cast, get_args
 
 import pulumi_nomad as nomad
 from infra.config import STACK_NAME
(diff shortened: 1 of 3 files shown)


pulumi/infra/nomad_job.py
```diff
@@ -91,9 +91,14 @@ def _fix_pulumi_preview(self, vars: NomadVars) -> NomadVars:
...
+                # This checks to see if this is a pulumi.Output[str] using _undocumented python implementation details_
+                # https://twitter.com/chompie1337/status/1435775022694555652?cxt=HHwWiICz0dXx8uwnAAAA
+                # SO thread on the python undocumented implementation details in question:
+                # https://stackoverflow.com/questions/57706180/generict-base-class-how-to-get-type-of-t-from-within-instance/60984681#60984681
+                if isinstance(value, pulumi.Output) and get_args(
+                    value.__orig_class__
+                ) == (str,):
                     value = pulumi_preview_replacement_string
                     if key == "redis_endpoint":
                         value = redis_endpoint
```
wimax-grapl, line 104: i'm surprised this works, but if it does, sweet! 

i do wonder if it might be better to simply say "hey, here's a list of keys i'd like to swap out" instead of "globally apply this to pulumi.Output[str]"

I don't think I fully understand the
"Without this, the vars gets unset if there's a single key with an unresolved output"
of this function; could you explain this case more?

originally introduced in https://github.com/grapl-security/grapl/pull/1189

wimax-grapl (approved):

PR #1772, opened 2022-06-17
Implement the event-source management service.

wimax-grapl:
# Quick diatribe on naming
I like "event source" or "event source service" over "event source management" or "event source management service." we should probably figure out a standardized way to name these services. I find management to be a lot of typing for not much lexical meaning-add. like, OrganizationService vs OrganizationManagementService? What is OrganizationService for if not managing organizations? 

# Back to usual PR stuff

https://github.com/grapl-security/issue-tracker/issues/847

1. Build out the Event Source Management.
2. Integration test for it
3. Seriously reduce repeated code in `grapl-local-infra` with regards to postgres dbs !!!

integration tests added

### DIFF
--- a/src/rust/event-source/src/client.rs
+++ b/src/rust/event-source/src/client.rs
@@ -31,8 +31,7 @@ impl FromEnv<EventSourceServiceClient, EventSourceServiceClientError> for EventS
             Duration::from_millis(10000),
             Duration::from_millis(500),
         )
-        .await
-        .expect("plugin-registry never reported healthy");
+        .await?;
 
         Self::connect(endpoint, None).await
     }
--- a/src/rust/event-source/src/client.rs
+++ b/src/rust/event-source/src/client.rs
@@ -2,7 +2,10 @@ use std::time::Duration;
 
 use async_trait;
 use rust_proto_new::{
-    graplinc::grapl::api::{event_source::v1beta1::client::{EventSourceServiceClient, EventSourceServiceClientError}},
+    graplinc::grapl::api::event_source::v1beta1::client::{
+        EventSourceServiceClient,
+        EventSourceServiceClientError,
+    },
     protocol::{
         healthcheck::client::HealthcheckClient,
         service_client::NamedService,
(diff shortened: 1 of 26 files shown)


src/rust/event-source/src/client.rs
```diff
@@ -0,0 +1,39 @@
...
+        let endpoint = std::env::var(ADDRESS_ENV_VAR).expect(ADDRESS_ENV_VAR);
+
+        HealthcheckClient::wait_until_healthy(
+            endpoint.clone(),
+            Self::SERVICE_NAME,
+            Duration::from_millis(10000),
+            Duration::from_millis(500),
+        )
+        .await
+        .expect("plugin-registry never reported healthy");
```
colin-grapl, line 35: It seems odd to panic here given that this function returns a Result, but my guess is that you don't want to add something like "never connected" to the error type used in a bunch of places? nbd if you don't want to change

src/rust/event-source/src/server.rs
```diff
@@ -0,0 +1,116 @@
...
+    config::EventSourceConfig,
+    db::EventSourceDbClient,
+    error::EventSourceError,
+};
+
+pub async fn exec_service(config: EventSourceConfig) -> Result<(), Box<dyn std::error::Error>> {
+    let api_impl = EventSourceApiImpl::try_from(&config).await?;
+    // I haven't quite figured out the right place to put the migrate.
+    // Perhaps when we construct the db client?!
+    api_impl.db_client.migrate().await?;
```
colin-grapl, line 25: I think you want to use the `migrate!` macro

colin-grapl (approved):

src/rust/event-source/src/server.rs
wimax-grapl: look at the implementation of this function

src/rust/event-source/src/client.rs
wimax-grapl: more importantly, copy-paste fail here :)

PR #1776, opened 2022-06-24
finish migrating pipeline and integration tests (mostly)

jgrillo-grapl:

loosely: https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/587
and: https://app.zenhub.com/workspace/o/grapl-security/issue-tracker/issues/368

This PR migrates the remaining pipeline services to Kafka, migrates some holdout services to `rust-proto-new`, removes `rust-proto`, renames `rust-proto-new` to `rust-proto`, and refactors the integration tests into `test-rust-integration` and `test-python-integration`. This is a big one, sorry. I recommend looking at each commit individually.

Automated tests.


nomad/integration-tests-new.nomad
```diff
@@ -158,6 +168,15 @@ job "integration-tests-new" {
 
         KAFKA_BOOTSTRAP_SERVERS = var.kafka_bootstrap_servers
 
+        ORGANIZATION_MANAGEMENT_BIND_ADDRESS   = "0.0.0.0:1004" # not used but required due to clap
```
christophermaier, line 181: Can you explain this comment a bit? If it's not used, why is it required? If it's required, why is it not used?

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
```diff
@@ -14,7 +20,25 @@ static PLUGIN_CONFIG_PATH: &str = "/etc/systemd/system/plugin.service.d/override
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let (_env, _guard) = grapl_config::init_grapl_env!();
 
-    let mut bootstrap_client = PluginBootstrapClient::from_env().await?;
+    let endpoint = std::env::var("PLUGIN_BOOTSTRAP_CLIENT_ADDRESS")?;
+    let plugin_bootstrap_polling_interval_ms: u64 =
+        std::env::var("PLUGIN_BOOTSTRAP_POLLING_INTERVAL_MS")?.parse()?;
+
+    tracing::info!(
+        message = "waiting 5s for plugin-bootstrap to report healthy",
```
christophermaier, line 28: This message depends on the value of `plugin_bootstrap_polling_interval_ms`, doesn't it?

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
```diff
@@ -14,7 +20,25 @@ static PLUGIN_CONFIG_PATH: &str = "/etc/systemd/system/plugin.service.d/override
...
+        endpoint = %endpoint,
+    );
+
+    HealthcheckClient::wait_until_healthy(
+        endpoint.clone(),
+        "graplinc.grapl.api.plugin_bootstrap.v1beta1.PluginBootstrapService",
+        Duration::from_secs(5), // TODO: parametrize
```
christophermaier, line 35: This was parameterized in this PR, correct?

src/rust/rust-proto/src/graplinc/grapl/api/suspicious_svchost_analyzer/v1beta1.rs
```diff
@@ -0,0 +1,166 @@
...
+};
+
+//
+// AnalyzeRequest
+//
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct AnalyzeRequest {
+    pub merged_graph: MergedGraph,
+}
```
colin-grapl, line 28: What's the purpose of the suspicious_svchost_analyzer stuff?

colin-grapl (approved):

pulumi/rust_integration_tests/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: integration-tests
```
christophermaier, line 1: Probably need to rename this to `rust-integration-tests`, and make a similar change for the Python ones.

christophermaier (approved):

nomad/integration-tests-new.nomad
jgrillo-grapl: I think I was mistakenly under the impression the organization-management client used this configuration: https://github.com/grapl-security/grapl/blob/internal-587-analyzers/src/rust/organization-management/src/lib.rs#L7-L8

but it doesn't appear to. I think we can remove this configuration value.

colin-grapl: I'm assuming it's because integration tests are going to have their ports assigned to them internally, but our configs require a valid endpoint

jgrillo-grapl: Oh, wait, actually it's used here: https://github.com/grapl-security/grapl/blob/internal-587-analyzers/src/rust/organization-management/tests/test_create_organization.rs#L7

So if that environment variable is missing the clap configuration will fail to parse. The tests apparently only use that configuration for its database credentials and address.

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
jgrillo-grapl: I think only the healthcheck polling interval was parametrized. This value is a timeout. The logic is that the heathcheck client will poll the healthcheck every `*_POLLING_INTERVAL_MS` until we hit the timeout, then fail.

src/rust/plugin-bootstrap/src/plugin_bootstrap_init.rs
jgrillo-grapl: No, this is that non-parametrized timeout value also

PR #1783, opened 2022-06-30
Enable pyupgrade as a pants backend

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/971

This adds a new tool called [pyupgrade](https://github.com/asottile/pyupgrade), which will rewrite python code to newer python syntax. This has been added as part of the pants suite, and will make the upgrade to python 3.10 MUCH easier

```
./pants lint ::
```


pants.toml
```diff
@@ -207,6 +208,9 @@ known_versions = [
   "v2.10.0|linux_x86_64|8ee6ff537341681f9e91bae2d5da451b15c575691e33980893732d866d3cefc4|2301804"
 ]
 
+[pyupgrade]
+args = ["--py38-plus"]
```
colin-grapl, line 212: I think we may be on 3.7 in some places, like CI. We should update those first.

christophermaier (approved):
Nice addition... worth the price of admission for the `Optional[X]` -> `X | None` update alone :joy:

etc/ci_scripts/dump_artifacts/nomad_artifacts.py
```diff
@@ -69,7 +69,7 @@ def _dump_hashicorp_agent_logs(artifacts_dir: Path) -> None:
 class NomadNamespace:
     name: str
 
-    def __init__(self, input: Dict[str, Any]) -> None:
+    def __init__(self, input: dict[str, Any]) -> None:
```
colin-grapl, line 72: Same here, I've always assumed that `List` and `Dict` were preferred to `list` and `dict` but I could absolutely be wrong

pants.toml
christophermaier: Things in CI are set up to honor `.python-version`, so we should be good.

@colin-grapl What specific instances of 3.7 are you seeing and concerned about?

colin-grapl (approved):

etc/ci_scripts/dump_artifacts/nomad_artifacts.py
```diff
@@ -18,14 +18,14 @@
 LOGGER.addHandler(logging.StreamHandler(stream=sys.stdout))
 
 OutOrErr = Literal["stdout", "stderr"]
-OUTPUT_TYPES: List[OutOrErr] = ["stdout", "stderr"]
+OUTPUT_TYPES: list[OutOrErr] = ["stdout", "stderr"]
```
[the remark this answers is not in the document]
twunderlich-grapl: So this is apparently preferred per [PEP 585](https://peps.python.org/pep-0585/#abstract). Which honestly after skimming makes sense

Dockerfile.pulumi
```diff
@@ -1,4 +1,6 @@
-FROM python:3.8-slim-bullseye as grapl-local-pulumi
+FROM python:3.10-slim-bullseye as grapl-local-pulumi
```
christophermaier, line 1: Once we've got everything on 3.10 --- that is, when our `.python-version` file contains a 3.10.x version --- then we could rig something up so that we pass that version in as a build argument and incorporate that into the base image identifier. That would be one less thing we'd have to keep in sync. (That goes for the `src/python/Dockerfile`, as well.)

(That's not a change for this PR, of course... just mentioning the idea for the future.)

christophermaier: Prior art for Rust: https://github.com/grapl-security/grapl/pull/1788/commits/bdefc4d1ffeb5fade35ae203d04f0aa113feeec2

christophermaier: Actually, let's just make an issue, shall we? 😅 https://github.com/grapl-security/issue-tracker/issues/973

PR #1795, opened 2022-07-07
make generator-executor a sidecar

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/977
Implementing some changes from this RFC change: https://github.com/grapl-security/grapl-rfcs/pull/33

- Turn `generator-executor` from a single service into a sidecar service for each plugin.
- Change many of the protos to reflect the RFC change.
- One consequence of these proto changes is that it necessitates removal of `tenant_id` from plugin-work-queue, so there's a decent amount of work around doing that.
- 

After running the rust-integration-tests, which deploy `example-generator`, we successfully have the sidecar running stablely alongside the plugin:
image

I can see in the logs that the execution sidecar:
1. successfully healthchecks against the plugin service
2. successfully calls "give me more work" from plugin-work-queue


nomad/grapl-core.nomad
```diff
@@ -1174,10 +1128,9 @@ job "grapl-core" {
         PLUGIN_REGISTRY_KERNEL_ARTIFACT_URL             = var.plugin_registry_kernel_artifact_url
         PLUGIN_REGISTRY_ROOTFS_ARTIFACT_URL             = var.plugin_registry_rootfs_artifact_url
         PLUGIN_REGISTRY_HAX_DOCKER_PLUGIN_RUNTIME_IMAGE = var.container_images["hax-docker-plugin-runtime"]
-        # Plugin Execution code/image doesn't exist yet; change this once it does!
-        PLUGIN_EXECUTION_CONTAINER_IMAGE      = "grapl/plugin-execution-sidecar-TODO"
-        PLUGIN_REGISTRY_BUCKET_AWS_ACCOUNT_ID = var.plugin_registry_bucket_aws_account_id
-        PLUGIN_REGISTRY_BUCKET_NAME           = var.plugin_registry_bucket_name
+        PLUGIN_EXECUTION_IMAGE                          = var.container_images["generator-executor"] # TODO: add support for analyzer too
```
wimax-grapl, line 1131: I suspect what we'll end up doing is changing this to a `plugin-executor` that can make both GetExecuteAnalyzer and GetExecuteGenerator requests; but this PR is too big already IMO!

jgrillo-grapl (approved):

PR #1798, opened 2022-07-08
Switch to using multiple lockfiles

ghost:
This migrates us from using a constraints file -> requirements.txt to using multiple lockfiles.

This is a dependency of our python 3.10 upgrade. Because we have the
pants-plugin repo which depends on pants, which doesn't support 3.10
we need one lockfile for the pants-plugin which can support older versions
of python and one for the rest of grapl which can support 3.10 (and
above?)
This also has some nice benefits including some supply-chain security
improvements

Automated tests


christophermaier (commented):
This would impact the usage of `build-support/manage_virtualenv.sh` (possibly eliminating it?), correct?

I believe there's also an impact on code in `Dockerfile.pulumi`, as well.

Might be worth adding a makefile target for upgrading all the lockfiles, too.

PR #1800, opened 2022-07-09
Set up autoflake via pants

twunderlich-grapl:
https://github.com/grapl-security/issue-tracker/issues/978

This sets up autoflake as part of the pants suite

make lint and make format


src/python/graplctl/graplctl/aws/commands.py
```diff
@@ -18,7 +18,7 @@ def aws(
     ctx: click.Context,
 ) -> None:
     """commands for managing grapl aws resources"""
-    pass
+    pass  # noqa
```
twunderlich-grapl, line 21: I personally like the explicit pass, but I'm open to having autoflake remove this automatically since strictly speaking its unnecessary.

jgrillo-grapl: I also prefer the explicit `pass`. I use it like `todo!()` in rust.

pants.toml
```diff
@@ -97,6 +98,11 @@ inits = true
 [python-protobuf]
 mypy_plugin = true
 
+[autoflake]
```
jgrillo-grapl, line 101: How do we actually invoke this? Is it something that should be added to `make format-python`?

jgrillo-grapl (approved):

PR #1804, opened 2022-07-12
Set PYTHON_VERSION for buildkite jobs

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/980
Fixes error in the CI pipeline: https://buildkite.com/grapl/grapl-merge/builds/760#0181f39e-3825-4e19-a267-0a28be427a4a

This fixes an error in https://buildkite.com/grapl/grapl-merge/builds/760#0181f39e-3825-4e19-a267-0a28be427a4a
Hardcode the python version in the buildkite pipeline. Previously we only defined PYTHON_VERSION in the makefile, but we have at least one job that doesn't go through Make and therefore needs the env variable defined twice
Long-term it would be nice to be able to grab this dynamically

Manual buildkite run


wimax-grapl (approved):

.buildkite/pipeline.verify.yml
```diff
@@ -5,6 +5,7 @@ env:
   # TODO: Figure out a way for this to be sourced from our
   # rust-toolchain file
   RUST_VERSION: 1.62.0
+  PYTHON_VERSION: 3.10.5
```
wimax-grapl, line 8: if it's gonna be duplicated from .python-version, would be nice to have a 
`# also update .python-version and the same in .python-version`

.buildkite/scripts/lib/lang_version.sh
```diff
@@ -0,0 +1,13 @@
...
+# Get the python version
+python_version() {
+    cat .python-version
+}
+# Hacky way to extract a value from a TOML file T_T
+#
+# This at least automatically keeps things in sync with with our Rust
+# toolchain.
+rust_version() {
+    grep channel src/rust/rust-toolchain.toml | sed -E 's/channel = "(.*)"/\1/g'
```
christophermaier, line 12: Let's not duplicate things: https://github.com/grapl-security/grapl/blob/main/build-support/rust_version.sh

christophermaier (changes requested):

.buildkite/scripts/lib/lang_version.sh
twunderlich-grapl: AH I didn't realize we had that. Updated

christophermaier (approved):
I think Cue will eventually make all this madness go away soon, but for now, this is still a vast improvement over the status quo.

PR #1807, opened 2022-07-13
Plugin-execution-sidecar: `trait PluginWorkProcessor` - genericize the core main loop, easy to plug in analyzer/generator

wimax-grapl:
None. 

Trying to wrap up plugin-execution-sidecar. 
This required:
1. rename every file
2. genericize out the loop that was previous generator-executor.rs specific into something that both Generator and Analyzer can consume.

I still need to implement analyzer - which still won't be fully complete because we don't have an Analyzer SDK yet. 

existing tests should capture failures since I added the getpluginhealth check!


src/rust/plugin-execution-sidecar/src/plugin_executor.rs
```diff
@@ -0,0 +1,69 @@
+use std::time::Duration;
```
wimax-grapl, line 1: most of this file is copied from `generator-executor.rs` but with the generator-specific parts replaced with PluginWorkProcessor

src/rust/plugin-execution-sidecar/src/plugin_executor.rs
```diff
@@ -0,0 +1,69 @@
...
+                self.plugin_work_processor
+                    .ack_work(
+                        &mut self.plugin_work_queue_client,
+                        self.plugin_id,
+                        request_id,
+                        success,
+                    )
+                    .await?;
+            } else {
+                tokio::time::sleep(Duration::from_secs(5)).await;
```
colin-grapl, line 62: This is a long sleep. I'd suggest cutting it to 1s.

src/rust/plugin-execution-sidecar/src/plugin_executor.rs
```diff
@@ -0,0 +1,69 @@
...
+                continue;
+            }
+        }
+        // Should we let the process exit if that while-let fails?
```
colin-grapl, line 66: yes, itll get restarted so its fine

PR #1810, opened 2022-07-14
Use reflection to get Nomad types during pulumi preview

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/981

This adds:
1. A small lib that parses hcl2 types after the python-hcl2 special format (types are wrapped in `${}`).
2. A function that replaces HCL2 primitives with fake values. ie MOCK_STRING for anything that's a string.
3. Updates _fix_pulumi_preview to use the mocking function

Since this issue stems from a type mismatch during pulumi preview where Pulumi object != number, the more robust way to fix this is to read the type that's expected for each Nomad input variable and pass that type in during pulumi preview.

Alternate approaches:
1. Skip the pulumi preview step entirely
2. Create a list of all the variables and their types in python and basically maintain the list in two separate places 🤮 

`make up` on this branch after I removed the skip-preview
Automated regression tests


3rdparty/python/requirements.txt
```diff
@@ -9,14 +9,15 @@ click
...
+lark-parser
 opentelemetry-api
 opentelemetry-exporter-zipkin-json
 opentelemetry-instrumentation
 opentelemetry-instrumentation-botocore
 opentelemetry-sdk
 protobuf>=3.15.8,<3.16.0
-pulumi>=3.0.0,<4.0.0
-pulumi-aws>=5.0.0,<6.0.0
+pulumi==3.33.1
```
twunderlich-grapl, line 19: I'm locking these since the upgrade causes pulumi preview/up to break. One bug at a time

Dockerfile.pulumi
```diff
@@ -1,7 +1,5 @@
 ARG PYTHON_VERSION
 FROM python:${PYTHON_VERSION}-slim-bullseye as grapl-local-pulumi
-# This upgrades to 3.10 because per https://bugs.python.org/issue42233, GenricAlias doesn't support Unions until 3.10.
```
twunderlich-grapl, line 3: I forgot to remove this in an earlier PR

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
         (
           cd grapl
           pulumi stack init local-grapl --non-interactive
-          pulumi up --yes --skip-preview --stack=local-grapl $$TRACING_ARGS
```
twunderlich-grapl, line 25: I've removed the skip preview since its no longer necessary and it acts as a check for new pulumi preview breakage

pulumi/hcl2_type_reflection/hcl2_type_reflection/hcl2_type_reflection.py
```diff
@@ -0,0 +1,80 @@
+from typing import Any
+
+from lark import Lark, Transformer
+
```
twunderlich-grapl, line 4: I'll note that I've only implemented what we're using so far. There are more HCL2 types. If we decide to open-source this as a standalone lib I'll finish implementing them

pulumi/python_integration_tests/__main__.py
```diff
@@ -103,7 +103,6 @@ def require_str(key: str) -> str:
             return cast(str, ref.require_output(key))
 
         self.aws_env_vars_for_local = require_str("aws-env-vars-for-local")
-        self.analyzer_bucket = require_str("analyzers-bucket")
```
twunderlich-grapl, line 106: These aren't used anymore and cause KeyErrors.

src/python/mypy.ini
```diff
@@ -35,5 +35,5 @@ ignore_missing_imports = True
 [mypy-confluent_kafka]
 ignore_missing_imports = True
 
-[mypy-netifaces]
```
twunderlich-grapl, line 38: This was supposed to be deleted a while back

twunderlich-grapl:
This seems to spike the mypy times dramatically. @wimax-grapl any idea what would cause mypy to take so much longer on this branch?

pulumi/hcl2_type_reflection/hcl2_type_reflection/hcl2_type_reflection.py
```diff
@@ -0,0 +1,80 @@
...
+        elif hcl2_type == "number":
+            return mock_number
+        else:
+            return hcl2_type
+    if isinstance(hcl2_type, dict):
+        mocked_dict = {
+            mock_hcl2_type(k): mock_hcl2_type(v) for (k, v) in hcl2_type.items()
+        }
+        return mocked_dict
+    return hcl2_type
```
wimax-grapl, line 80: just for the record - we're on py10 now!
```
def asdf(a):
  match a:
    case str(_): print("str")
    case dict(d): print(f"dict {d}")
    case _: print("idk")

asdf("hello")
asdf({"hello": 123})
```

wimax-grapl (approved):
this is so overengineered but also it rules and i like it. 11/10.

pulumi/hcl2_type_reflection/hcl2_type_reflection/hcl2_type_reflection.py
twunderlich-grapl: Python got switch statements? 🤯

PR #1811, opened 2022-07-15
Plugin-execution-sidecar now has a Producer and produces messages to `generated-graphs`

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/977

Previously the `plugin-execution-sidecar` would not push messages to Kafka. Now it does. 

CI. No messages flow through to the producer quite just yet, unfortunately, but this should be exercised once generator-dispatcher is ready to go.


colin-grapl (approved):

jgrillo-grapl (changes requested):
I have some concerns about this. Originally the `plugin-work-queue` was supposed to mediate communication between the pipeline and the plugins. This breaks that. In particular, I expect there to be many plugins, therefore many plugin sidecars, entailing (in this case) many Kafka producers.

This will definitely run us up against the client connection limit in Confluent Cloud. See "total client connections" here: https://docs.confluent.io/cloud/current/clusters/cluster-types.html

PR #1816, opened 2022-07-19
Fix sessiondb integration tests

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/983

The sessiondb integration tests require the ability to connect to the
session db. This adds the dynamodb name for the sessiondb. Its unclear why this passes locally but fails when run
against AWS, although I suspect its a localstack vs requiring the
actual dynamodb connection

Automated tests to catch any issues.


jgrillo-grapl (approved):
LGTM

PR #1819, opened 2022-07-19
Remove opentelemetry during prod, in Rust

wimax-grapl:
none, followup to https://github.com/grapl-security/grapl/pull/1817

We were seeing in CI that grapl-core failed to deploy due to 
> OpenTelemetry trace error occurred. Exporter jaeger encountered the following error(s): thrift agent failed with not open

This gates the rust opentelemetry stuff behind "does the OTEL_EXPORTER_JAEGER_AGENT_HOST env var exist?"

I then standardized how we feed in the otel jaeger/zipkin stuff into services. It'll populate those env vars in local but not in prod, just like aws_env_vars_for_local. 

I overrode `observability_env_vars_for_local` locally so it'd look like prod; stuff started up!

### DIFF
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -29,6 +29,14 @@ weird nomad state parse error.
 EOF
 }
 
+variable "observability_env_vars" {
+  type        = string
+  description = <<EOF
+With local-grapl, we have to inject env vars for Opentelemetry.
+In prod, this is currently disabled.
+EOF
+}
+
 variable "aws_region" {
   type = string
 }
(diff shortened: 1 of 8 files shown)


twunderlich-grapl:
I'll note that with Nomad 1.3.2, I should have a way forward for prod

nomad/grapl-core.nomad
```diff
@@ -1131,12 +1163,11 @@ job "grapl-core" {
         PLUGIN_EXECUTION_IMAGE                          = var.container_images["generator-execution-sidecar"] # TODO: add support for analyzer too
         PLUGIN_REGISTRY_BUCKET_AWS_ACCOUNT_ID           = var.plugin_registry_bucket_aws_account_id
         PLUGIN_REGISTRY_BUCKET_NAME                     = var.plugin_registry_bucket_name
+        PLUGIN_EXECUTION_OBSERVABILITY_ENV_VARS         = var.observability_env_vars
```
twunderlich-grapl, line 1166: Why does this need a separate env var that just repeats the existing env var?

twunderlich-grapl: Ah, this is passed into the plugins. Got it.

twunderlich-grapl (approved):

nomad/grapl-core.nomad
wimax-grapl: So, we pass through some vars from
`grapl-core.nomad` -> plugin-registry -> `hax_docker_plugin.nomad` -> plugin binaries

what we're doing to plugin-registry is:
- passing it env vars: `OTEL_BLAH_BLAH=1, ...` that plugin-registry uses at runtime
- passing it a template string: "OTEL_BLAH_BLAH=1\n..." that plugin-registry passes down to the plugin binaries

By passing in the template string, we basically **remove some complexity around the optional-ness of the env vars** (the reason we did the aws_env_vars_for_local stuff in the first place) + make hax_docker_plugin.nomad more-resemble grapl-core.nomad

PR #1822, opened 2022-07-20
Add Sign In With Google to Grapl web

inickles-grapl:

This adds the ability for users to authenticate to Grapl web via Sign In With Google. In doing so I've refactored the authentication code a bit to separate handling of the web requests and the authentication logic.

This also introduces a (very basic) integration test, with a helper utility to make it easier to add new tests. A test for username + password authentication was not included because the future of that is uncertain. A test for Sign In With Google was not included because it's not yet clear how we should create a long-lived Google token we can create that can be used.

Room for improvements in this PR (looking for input):
1. I figure the Google client ID should probably be a parameter to a Grapl Pulumi config, but I'm not sure the best way to do that. I'm happy to include that in this PR with some direction from reviewers.
2. The frontend UI placement of the Sign In With Google. It seems okay for now, we may just want to fix this up in a separate PR, but I'm open to input here.

1. I added my email address to `GRAPL_TEST_USER_NAME` in pulumi/infra/config.py to populate the user auth database, and then brought up with environment with `make up`, and tested the new auth flow locally.
2. Rust integration tests.


pulumi/grapl/__main__.py
```diff
@@ -209,6 +209,12 @@ def main() -> None:
 
     observability_env_vars = observability_env_vars_for_local()
 
+    # This Google client ID is used by grapl-web-ui for authenticating users via Sign In With Google.
+    # TODO: This should be moved to Pulumi config somehwo, but I'm not sure the best way to do that atm.
+    google_client_id = (
+        "340240241744-6mu4h5i6h9j7ntp45p3aki81lqd4gc8t.apps.googleusercontent.com"
+    )
+
```
colin-grapl, line 217: I assume we don't want this leaking?

inickles-grapl: client ID is fine, there's a secret we wouldn't want to expose, but it's not needed here.

src/rust/grapl-web-ui/tests/api/test_app.rs
```diff
@@ -0,0 +1,45 @@
...
+        Ok(Self {
+            bind_address,
+            client: awc::Client::default(),
+        })
+    }
+
+    pub fn post(&self, path: &str) -> awc::ClientRequest {
+        let bind_address = self.bind_address.as_str();
+        self.client
+            .post(format!("http://{bind_address}/{path}"))
```
colin-grapl, line 42: I think this is fine but probably the ideal would be to use a Url builder? The concern being some sort of injection attack, but again, I don't think that in this case it's attacker controlled data. Or maybe it is idk.

colin-grapl (approved):

src/rust/grapl-web-ui/tests/api/test_app.rs
inickles-grapl: The inputs are not attacker controlled and this is only used for integration tests. I think this is fine.

PR #1825, opened 2022-07-21
[grapl-web-ui] Remove unused backend services and clean up routes

inickles-grapl:
### What issue does this correspond to?

https://github.com/grapl-security/issue-tracker/issues/902

This moves API endpoints behind the "/api" path and refactors how routes are handled in grapl-web-ui.

I tested signing in with a password user the credentials for the local user. I was able to sign in successfully and without other errors.

I also tried signing in with google, which I expected to fail with a 401, as it did, showing the routes match what they were before.


src/rust/grapl-web-ui/src/routes.rs
```diff
@@ -0,0 +1,23 @@
...
+
+use actix_web::{
+    guard,
+    web,
+};
+
+pub(super) fn config(cfg: &mut actix_web::web::ServiceConfig) {
+    grapl_logo::config(cfg);
+    index::config(cfg);
+    // login::config(cfg);
```
inickles-grapl, line 14: oops, will remove

wimax-grapl (approved):

colin-grapl (approved):

PR #1826, opened 2022-07-21
`rust-proto-clients`: a crate for easily getting clients of our rust-proto services

wimax-grapl:
# How to review
First see [get_grpc_client](https://github.com/grapl-security/grapl/pull/1826/files#diff-26bff64baaaf8cac87a0e9c1a7b7665037ced7dec08c90f6ab72ba700b96ba4fR18)
it takes in a generic GrpcClientConfig which [you can see here](https://github.com/grapl-security/grapl/pull/1826/files#diff-a7f09b071dec3330059a68e8e4157165fd82c6840b4cd390ca6abc5fc0717fe2R11). 

The rest of the PR is just:
- upfixing previous client-gets to use this new form
- implementing GrpcClientConfig for each service
---

None. I've wanted to do this for eons.

Previously we did not have a pattern for getting grpc clients; it was a bunch of repeated HealthCheck code and bare ENV_VAR_NAMES scattered around the codebase. This crate gives us a low-dependencies easy way to construct a client for a given service.

Existing CI


src/rust/event-source/tests/test_crud_event_source.rs
```diff
@@ -2,13 +2,18 @@
...
 use grapl_utils::future_ext::GraplFutureExt;
 use rust_proto::graplinc::grapl::api::event_source::v1beta1 as es_api;
+use rust_proto_clients::{
+    get_grpc_client,
+    services::EventSourceClientConfig,
+};
 
 #[test_log::test(tokio::test)]
 async fn test_create_update_get() -> Result<(), Box<dyn std::error::Error>> {
-    let mut client = es_api::client::EventSourceServiceClient::from_env().await?;
```
wimax-grapl, line 11: tldr these FromEnvs were scattershot everywhere, i've now made them standardized and `impl GrpcClientConfig`

src/rust/rust-proto-clients/src/grpc_client.rs
```diff
@@ -0,0 +1,46 @@
...
+    if !address.starts_with("http") {
+        panic!("Address should start with http, but found: '{address}'")
+    }
+
+    // TODO: Add a `rust-proto` wrapper around tonic Endpoint
+    let endpoint = Endpoint::from_shared(address)?
+        .timeout(Duration::from_secs(10))
+        .concurrency_limit(30);
+
+    HealthcheckClient::wait_until_healthy(
```
colin-grapl, line 37: There's no need to wait until a service is healthy, this just makes client connections much slower. In tests it's fine but there's no benefit in establishing a client connection.

src/rust/rust-proto-clients/src/grpc_client.rs
```diff
@@ -0,0 +1,46 @@
...
+        endpoint.clone(),
+        service_name,
+        Duration::from_millis(10000),
+        Duration::from_millis(healthcheck_polling_interval_ms),
```
wimax-grapl, line 41: all these hardcoded timeouts etc will eventually make their way in to GenericClientConfig

colin-grapl (changes requested):
Remove client healthcheck and we're good to go

src/rust/rust-proto-clients/src/grpc_client.rs
wimax-grapl: If the service isn't healthy it usually isn't serving traffic. I've spent /days/ debugging shit where the root cause was "it wasn't healthy yet"

colin-grapl (approved):

src/rust/generator-dispatcher/tests/test_utils/context.rs
```diff
@@ -28,27 +28,16 @@ impl AsyncTestContext for GeneratorDispatcherTestContext {
...
+            },
         )
         .await
-        .expect("pipeline-ingress never reported healthy");
-
-        tracing::info!("connecting pipeline-ingress gRPC client");
-        let pipeline_ingress_client = PipelineIngressClient::connect(endpoint.clone())
-            .await
-            .expect("could not configure gRPC client");
+        .expect("pipeline_ingress_client");
```
jgrillo-grapl, line 40: this message doesn't tell me what _happened_, whereas the previous one (in the context of the logged messages) tells me exactly what went wrong.. we waited for 10s but the service never reported healthy.

src/rust/rust-proto-clients/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "rust-proto-clients"
```
jgrillo-grapl, line 2: why is this a separate crate?

src/rust/rust-proto-clients/src/grpc_client.rs
```diff
@@ -0,0 +1,66 @@
...
+impl Default for GetGrpcClientOptions {
+    fn default() -> Self {
+        GetGrpcClientOptions {
+            perform_healthcheck: false,
+            healthcheck_polling_interval: Duration::from_millis(500),
+        }
+    }
+}
+
+pub async fn get_grpc_client<C: GrpcClientConfig>(
```
jgrillo-grapl, line 32: can we instead call these something like `create_*` or `build_*`? They don't actually really "get" anything, instead they build a new thing every time you call one.

That said, wouldn't this make more sense as a _static method on the actual client_ than as a disembodied constructor in a different crate?

jgrillo-grapl (commented):
Ugh apparently this has already been merged. I have some structural concerns here, and some nits.

PR #1830, opened 2022-07-22
[grapl-web-ui] Add integration testing for authenticated web APIs

inickles-grapl:

Primarily this adds the ability to test authenticated APIs in the grapl-web-ui. Each integration test can initialize a new instance of TestApp, which provisions a test user in the auth database for the test. This adds an integration test for the check_login API when the user is authenticated. 

Additionally, this PR has a number of other small changes:
- Increase memory of Nomad jobs for web-ui and rust-intergation-tests - was getting OOM errors.
- I removed what looks to be similar authentication tests from Python.
- Add a health check to grapl-web-ui at `api/health' and configure Nomad to use that to determine service health.
- Add a ConfigBuilder to make it easier to configure the bind address.

I ran the Rust integration tests and verified they were running.


src/rust/Dockerfile
```diff
@@ -157,7 +157,7 @@ RUN --mount=type=cache,target="${CARGO_TARGET_DIR}",sharing=locked \
         --features "${RUST_INTEGRATION_TEST_FEATURES}" \
         --no-run \
         --message-format=json \
-        --test "*" | \
+        --test "*" -- --nocapture | \
```
inickles-grapl, line 160: do not capture stdout from tests

nomad/rust-integration-tests.nomad
```diff
@@ -159,6 +154,11 @@ job "rust-integration-tests" {
               destination_name = "event-source"
               local_bind_port  = 1005
             }
+
+            upstreams {
+              destination_name = "web-ui"
+              local_bind_port  = 1007
```
wimax-grapl, line 160: noti mportant, but why not 1006

inickles-grapl: Ah, no good reason: originally I had another upstream using 1006, I just hadn't updated this one after removing the other. I'll fix that to use 1006.

src/rust/grapl-web-ui/tests/api/auth/check_login.rs
```diff
@@ -0,0 +1,27 @@
+#[actix_web::test]
+async fn auth_unauthenticated_check_login() -> eyre::Result<()> {
+    let app = crate::test_app::TestApp::init().await?;
+
+    let response = app.post("api/auth/checkLogin").send().await?;
+
+    println!("{:?}", response);
```
wimax-grapl, line 7: don't love the bare println but whatever

src/rust/grapl-web-ui/tests/api/config.rs
```diff
@@ -0,0 +1,56 @@
+use grapl_config::env_helpers::FromEnv;
+use rusoto_dynamodb::DynamoDbClient;
+
+type Result<T> = std::result::Result<T, TestConfigError>;
+
+pub struct TestConfig {
```
wimax-grapl, line 6: fwiw, in the rest of our codebase we've mostly been using Clap for inputting env var stuff like this, but no biggie either way

wimax-grapl (approved):

src/rust/grapl-web-ui/tests/api/config.rs
inickles-grapl: Yeah, that's a good point - it's what I should have done. I think I'll make that a separate PR after this, where I'll migrate this and the main app to clap as well.

I migrated the main app to use clap in a later draft PR (https://github.com/grapl-security/grapl/pull/1839) - I'll start by bringing that part into its own.

src/rust/grapl-web-ui/tests/api/auth/check_login.rs
inickles-grapl: Yeah, they could at least say what they are. 👍

PR #1838, opened 2022-07-26
E2E tests for the new Kafka/plugin-y world

wimax-grapl:
This introduces a new integration test that:
- uploads 36_eventlog.xml to pipeline-ingress
- tries to "listen in" on every step of the pipeline - scan the Kafka pipelines between services, scan the plugin work queue, etc (so far, up to post-graph-merger) and make sure expected messages are flowing through them.


src/rust/Dockerfile
```diff
@@ -139,7 +138,7 @@ FROM base AS build-test-integration
 # https://github.com/rust-lang/cargo/issues/1924
 # https://github.com/rust-lang/cargo/issues/3670
 
-ENV RUST_INTEGRATION_TEST_FEATURES="generator-dispatcher/integration_tests,graph-merger/integration_tests,node-identifier/integration_tests,pipeline-ingress/integration_tests,plugin-registry/integration_tests,plugin-work-queue/integration_tests,sysmon-generator/integration_tests,organization-management/integration_tests,grapl-web-ui/integration_tests"
+ENV RUST_INTEGRATION_TEST_FEATURES="generator-dispatcher/integration_tests,graph-merger/integration_tests,node-identifier/integration_tests,pipeline-ingress/integration_tests,plugin-registry/integration_tests,plugin-work-queue/integration_tests,sysmon-generator/integration_tests,organization-management/integration_tests,grapl-web-ui/integration_tests,e2e-tests/integration_tests"
```
wimax-grapl, line 141: p.s. i hate this and i'm gonna fix it

wimax-grapl: in a different PR

src/rust/plugin-execution-sidecar/generator-execution-sidecar.rs
```diff
@@ -12,6 +12,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let _guard = setup_tracing(SERVICE_NAME)?;
     let plugin_executor_config = PluginExecutorConfig::parse();
 
+    // Give the plugin a little time to become available.
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
jgrillo-grapl, line 16: Is there any way we can make this deterministic?

jgrillo-grapl (approved):
LGTM. It looks like the integration tests are already configured to be able to read from all the topics in AWS so this should run 🤞 : https://github.com/grapl-security/confluent-cloud-infrastructure/blob/97919cae2fce725ef3a50491f4d3c28f864a32de/pulumi/ccloud_bootstrap/index.ts#L1181-L1189

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
```diff
@@ -0,0 +1,177 @@
...
+
+use crate::test_utils::predicates::{
+    events_36lines_merged_graph_predicate,
+    events_36lines_node_identity_predicate,
+};
+
+mod test_fixtures {
+    use super::*;
+
+    /// Send 1 line (well, event) at a time
```
inickles-grapl, line 35: Any particular reason we're doing this instead of just sending up the whole file?

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
```diff
@@ -0,0 +1,177 @@
...
+    pub fn get_36_eventlog_xml_separate_lines() -> Result<Vec<String>, std::io::Error> {
+        let filename = "/test-fixtures/36_eventlog.xml"; // This path is created in rust/Dockerfile
```
inickles-grapl, line 37: IMO we should just move that file under this package, or sysmon-generator + symlink to that from this package, and not rely on sync with the Dockerfile. That file is only used for e2e testing and doesn't need to be in a repo-level etc directory, right?

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    pub fn get_sysmon_generator() -> Result<Bytes, std::io::Error> {
+        std::fs::read("/test-fixtures/sysmon-generator").map(Bytes::from)
+    }
+}
+
+#[test_context(E2eTestContext)]
+#[tokio::test]
+async fn test_sysmon_log_e2e(ctx: &mut E2eTestContext) -> Result<(), Box<dyn std::error::Error>> {
```
inickles-grapl, line 49: My preference is to use eyre::Result for tests because just using std Result with Box<dyn std::error::Error> won't get you nearly as much, including the error's Display.

src/rust/plugin-execution-sidecar/generator-execution-sidecar.rs
wimax-grapl: discussion here
https://grapl-internal.slack.com/archives/C03JNJ57A9G/p1659471556377899

basically plugins and plugin-execution-sidecar may have a dependency cycle of health checks. fun!

wimax-grapl: so i phoned it in for now

src/rust/rust-proto/src/protocol/service_client.rs
```diff
@@ -17,12 +17,12 @@ pub trait Connectable {
...
 #[derive(Debug, thiserror::Error)]
 pub enum ConnectError {
-    #[error("failed to connect {0}")]
+    #[error("Failed to connect {0}")]
     ConnectionError(#[from] tonic::transport::Error),
 
-    #[error("healthcheck failed {0}")]
-    HealtcheckFailed(#[from] HealthcheckError),
+    #[error("Healthcheck failed for {0}: {1}")]
+    HealthcheckFailed(String, HealthcheckError),
```
inickles-grapl, line 24: My preference here would be to do the following:
```
#[error("healthcheck failed for `{service_name}`: {source})]
HealthcheckFailed {
    service_name: String,
    source: HealthcheckError
}
```

This way we don't discard the Error's source, which could be useful information.

> The Error trait's source() method is implemented to return whichever field has a #[source] attribute or is named source, if any. This is for identifying the underlying lower level error that caused your error.
-- https://github.com/dtolnay/thiserror

inickles-grapl (commented):
Wrapping up what I have now since it's already been merged.

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
wimax-grapl: There was a very, very long conversation about this in #plugin-system starting with https://grapl-internal.slack.com/archives/C03JNJ57A9G/p1659408013217299 up until like noon today

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
wimax-grapl: That's great to know, I'll look into it

src/rust/rust-proto/src/protocol/service_client.rs
wimax-grapl: Can do

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
wimax-grapl: i can very much see a world where perhaps another test that's not e2e would want to rely on these - in fact i've been in those scenarios before. I'd like to maybe punt on this until the future of the repo is a bit more obvious.

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
inickles-grapl: oh boy...

PR #1842, opened 2022-07-27
Add client-executor crate

colin-grapl:
Draft


src/rust/client-executor/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+futures = "0.3.21"
+pin-project = "1.0.11"
+recloser = "1.0.0"
+tokio = { version = "1.20.0", features = ["full"] }
```
wimax-grapl, line 12: rest of our codebase is 1.17

src/rust/client-executor/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    Sleeping(#[pin] Sleep),
+}
+
+impl<A: Action> ExecuteState<A> {
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> ExecuteFuturePoll<A> {
+        match self.project() {
+            ExecuteStateProj::Running(future) => {
+                match future.poll(cx) {
+                    Poll::Ready(Err(recloser::Error::Inner(e))) => ExecuteFuturePoll::Running(Poll::Ready(Err(Error::Inner(e)))),
+                    Poll::Ready(Err(recloser::Error::Rejected)) => ExecuteFuturePoll::Running(Poll::Ready(Err(Error::Rejected))),
```
inickles-grapl, line 38: At first I was wondering why this crate's Error just mirrors recloser::Error instead of just using it with thiserror's `#[from] recloser::Error`, but since that'd be this crate's only variant it seems like you could just type alias recloser::Error instead? At least for now, until you add more variants you might have plans for. Thoughts?

inickles-grapl: Ah, now I see recloser::Error doesn't impl std::error::Error and it doesn't require the E generic does either, as you do.

colin-grapl: `recloser::Error` does not implement `std::error::Error`, and I wanted to avoid re-exporting the crate.

src/rust/client-executor/Cargo.toml
colin-grapl: `version = "1.X"` will choose any version above `1.x`. So if we specify 1.17 elsewhere and 1.20 here we'll end up with 1.20 everywhere, and it'll be a lockfile failure otherwise. You can confirm this in the lockfile - you'll see 1.20 is the version in use for tokio.

src/rust/client-executor/src/lib.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// For example when more than 50% (`0.5f32`) of the recorded calls have failed.
+    pub fn threshold(mut self, threshold: f32) -> Self {
+        self.builder = self.builder.error_rate(threshold);
+        self
+    }
+
+    /// How many calls will the circuit remain closed for before the failure
+    /// rate is recalculated and the circuit may open
+    pub fn closed_len(mut self, closed_len: usize) -> Self {
+        assert_ne!(closed_len, 0, "closed_len must be non-zero");
```
inickles-grapl, line 250: 1) seems like we should probably just std::num::NonZeroUsize instead 
2) shouldn't we return a Result with a new `ExecutorConfigError` type instead of asserting?

src/rust/client-executor/src/lib.rs
```diff
@@ -0,0 +1,421 @@
...
+
+impl ExecutorConfig {
+    /// Creates a new `ExecutorConfig` from a `timeout`. Note that the `timeout`
+    /// is *not* a global timeout, but will be applied to each individual call
+    /// of the underlying future.
+    /// The rest of the parameters - threshold, closed_len, etc, are initialized
+    /// to defaults.
+    pub fn new(timeout: Duration) -> Self {
+        let builder = Recloser::custom()
+            .error_rate(0.5)
```
jgrillo-grapl, line 225: Why 0.5? Similar questions apply to the other hardcoded numbers here.

jgrillo-grapl (approved):
LGTM

src/rust/client-executor/src/lib.rs
colin-grapl: https://resilience4j.readme.io/docs/circuitbreaker

They're just based on the resilience4j defaults, or very slightly modified (I decreased the CircuitOpen timeout).

PR #1843, opened 2022-07-27
Move 'scan_for_plugin_message_in_pwq()` to PluginWorkQueue's test-utils

wimax-grapl:
Also, change the feature to "test-utils"; Depending on another package's "integration_tests" was an antipattern that was causing code to unexpectedly run.

### DIFF
--- a/src/rust/plugin-work-queue/src/psql_queue.rs
+++ b/src/rust/plugin-work-queue/src/psql_queue.rs
@@ -82,7 +82,7 @@ impl PsqlQueue {
         Ok(Self::new(pool))
     }
 
-    #[instrument(skip(pipeline_message), err)]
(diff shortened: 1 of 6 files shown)


src/rust/plugin-work-queue/src/psql_queue.rs
```diff
@@ -82,7 +82,7 @@ impl PsqlQueue {
         Ok(Self::new(pool))
     }
 
-    #[instrument(skip(pipeline_message), err)]
+    #[instrument(skip(self, pipeline_message), err)]
```
wimax-grapl, line 85: an unrelated good change

jgrillo-grapl (approved):

PR #1846, opened 2022-07-28
Fix panic in generator_ids_cache

wimax-grapl:
[blargh 

](https://grapl-internal.slack.com/archives/C03JNJ57A9G/p1658860185498869)

### DIFF
--- a/src/rust/rust-proto/src/protocol/status.rs
+++ b/src/rust/rust-proto/src/protocol/status.rs
@@ -129,7 +129,7 @@ impl From<GrpcCode> for Code {
 #[derive(thiserror::Error)]
 pub struct Status {
     /// The gRPC status code, found in the `grpc-status` header.
-    code: Code,
+    pub code: Code,
     /// A relevant error message, found in the `grpc-message` header.
     message: String,
 }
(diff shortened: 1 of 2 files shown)


src/rust/rust-proto/src/protocol/status.rs
```diff
@@ -129,7 +129,7 @@ impl From<GrpcCode> for Code {
 #[derive(thiserror::Error)]
 pub struct Status {
     /// The gRPC status code, found in the `grpc-status` header.
-    code: Code,
+    pub code: Code,
```
wimax-grapl, line 132: i think the 'matching on a Code' is a common enough operation that making this pub makes sense

colin-grapl: I'm fine with it, but `pub` also means that it can be mutated. Getters are more restrictive. I get no preference.

colin-grapl (approved):

src/rust/rust-proto/src/protocol/status.rs
jgrillo-grapl, line 132: Maybe a getter `pub fn code() -> &Code`? I've started doing that for new code in rust-proto (in the branch that I haven't PRed yet). Making this `pub` is fine for now, though.

jgrillo-grapl (approved):

src/rust/rust-proto/src/protocol/status.rs
wimax-grapl: could we match on that?

jgrillo-grapl: Yes, I believe so?

colin-grapl (approved):

PR #1855, opened 2022-08-01
Postgres client abstraction that safely handles credentials, removes boilerplate

inickles-grapl:

This relates to https://github.com/grapl-security/issue-tracker/issues/990 and is a follow up to https://github.com/grapl-security/grapl/pull/1848.

This has all services using Postgres use a secrets type for safer handling of Postgres credentials.

Additionally, this provides a Postgres client trait that intends to reduce boilerplate with some default implementations. A trait is used in-part so users can override the default functionality.

This also introduces a trait, `ToPostgresUrl`, to be used with Postgres service configs. The need for this trait is somewhat unfortunate, and we can probably move away from it and allow all services to use a common Postgres config type if/when clap::Praser derive macro supports supplied prefixes when `flatten`ing configs. The following code comment in this PR explains details of this situation:

> /// Ideally we could have a single Postgres config type that all Postgres clients could use,
/// instead of needing this trait, but we can't exactly do that right now while keeping consistent
/// with Grapl's use of clap derive for argument processing.
///
/// The clap::Parser derive macro does not support supllied prefixes when flattening a config
/// structure. This means all fields of the Postgres config would need to be same across all uses
/// of it (ex: DB_ADDRESS). At the time of this writing that wouldn't be a problem for running our
/// services themselves, because 1) we don't have any services that uses multiple Postgres servers,
/// and 2) we could set the common name in our grapl-core.nomad for each service (ex: DB_USERNAME
/// = var.example.username). However, our Rust integration tests are all running in a single
/// environment, and those tests need to use the same config builders as the services themselves.
/// So for now we need to have the argument variable names include uniqueness, such as prefixing
/// with the service's name (EXAMPLE_DB_HOSTNAME), meaning we can't use a common config. It looks
/// like this issue is being tracked at https://github.com/clap-rs/clap/issues/3513.
/// 
/// We could stray from using clap for this and have code that takes a prefix for loading Postgres
/// environment configs, but this isn't great for a couple reasons:
/// 1. It hides required arguments for a service in a different package. Some services have a
///    src/config.rs, which clearly defines all the parameters for the service. But by not using
///    clap here we'd break that for users, and we'd break the help formatter produced by clap when
///    one of the clap arguments are not supplied.
/// 2. Clap has good error reporting when required arguments are not supplied. We'd have to
///    recreate that, which isn't a big deal, but still.

Rust integration tests

### DIFF
--- a/src/rust/event-source/src/server.rs
+++ b/src/rust/event-source/src/server.rs
@@ -1,5 +1,6 @@
 use std::time::Duration;
 
+use grapl_config::PostgresClient;
 use rust_proto::{
     graplinc::grapl::api::event_source::{
         v1beta1 as native,
@@ -19,10 +20,8 @@ use crate::{
 };
 
 pub async fn exec_service(config: EventSourceConfig) -> Result<(), Box<dyn std::error::Error>> {
-    let api_impl = EventSourceApiImpl::try_from(&config).await?;
-    // I haven't quite figured out the right place to put the migrate.
-    // Perhaps when we construct the db client?!
-    api_impl.db_client.migrate().await?;
+    let api_impl = EventSourceApiImpl::try_from(config.clone()).await?;
+
     let (server, _shutdown_tx) = EventSourceServer::new(
         api_impl,
         TcpListener::bind(config.service_config.event_source_bind_address.clone()).await?,
@@ -47,9 +46,8 @@ pub struct EventSourceApiImpl {
 }
 
 impl EventSourceApiImpl {
-    pub async fn try_from(config: &EventSourceConfig) -> Result<Self, EventSourceError> {
-        let config = config.clone();
-        let db_client = EventSourceDbClient::try_from(config.db_config.clone()).await?;
+    pub async fn try_from(config: EventSourceConfig) -> Result<Self, EventSourceError> {
+        let db_client = EventSourceDbClient::init_with_config(config.db_config.clone()).await?;
         Ok(Self { config, db_client })
     }
 }
--- a/src/rust/grapl-config/src/postgres.rs
+++ b/src/rust/grapl-config/src/postgres.rs
@@ -1,0 +1,119 @@
+use grapl_utils::future_ext::GraplFutureExt;
+use secrecy::ExposeSecret;
+
+#[derive(Debug, Clone)]
+pub struct PostgresUrl {
+    pub address: String,
+    pub username: String,
+    pub password: super::SecretString,
+}
+
+impl std::fmt::Display for PostgresUrl {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "postgresql://{username}:<REDACTED>@{address}",
+            username = self.username,
+            address = self.address
+        )
+    }
+}
+
+impl PostgresUrl {
+    /// Connect to Postgres with a timeout of 5 seconds.
+    pub async fn connect(&self) -> Result<sqlx::Pool<sqlx::Postgres>, PostgresDbInitError> {
+        let timeout = std::time::Duration::from_secs(5);
+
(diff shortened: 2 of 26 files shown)


src/rust/event-source/src/server.rs
```diff
@@ -49,7 +48,7 @@ pub struct EventSourceApiImpl {
 impl EventSourceApiImpl {
     pub async fn try_from(config: &EventSourceConfig) -> Result<Self, EventSourceError> {
         let config = config.clone();
-        let db_client = EventSourceDbClient::try_from(config.db_config.clone()).await?;
+        let db_client = EventSourceDbClient::init_with_config(config.db_config.clone()).await?;
         Ok(Self { config, db_client })
```
colin-grapl, line 52: should prob just take config by `move` if we clone it anyways

src/rust/grapl-config/src/postgres.rs
```diff
@@ -0,0 +1,121 @@
...
+#[non_exhaustive]
+pub enum PostgresDbInitError {
+    #[error("sqlx migration error: {0}")]
+    MigrateError(#[from] sqlx::migrate::MigrateError),
+    #[error("unable to connect: {0}")]
+    Sqlx(#[from] sqlx::Error),
+    #[error("connection timeout after {timeout:?}")]
+    Timeout {
+        timeout: std::time::Duration,
+        // keep the folllwing to prevent Colin from constructing this variant
```
colin-grapl, line 64: gdi

src/rust/organization-management/src/lib.rs
```diff
@@ -9,11 +9,19 @@ pub struct OrganizationManagementServiceConfig {
...
-    pub organization_management_db_port: u16,
+    pub organization_management_db_address: String,
     #[clap(long, env)]
     pub organization_management_db_username: String,
     #[clap(long, env)]
-    pub organization_management_db_password: String,
+    pub organization_management_db_password: grapl_config::SecretString,
+}
+
+impl grapl_config::ToPostgresUrl for OrganizationManagementServiceConfig {
```
colin-grapl, line 19: 99% chance that i proc macro this before EOD

src/rust/organization-management/src/server.rs
```diff
@@ -70,25 +70,23 @@ pub struct OrganizationManagement {
...
+    fn new(pool: sqlx::Pool<sqlx::Postgres>) -> Self {
+        Self { pool }
+    }
 
-        Ok(Self {
-            pool: sqlx::PgPool::connect(&postgres_address)
-                .timeout(Duration::from_secs(5))
-                .await??,
-        })
+    async fn migrate(pool: &sqlx::Pool<sqlx::Postgres>) -> Result<(), sqlx::migrate::MigrateError> {
```
colin-grapl, line 82: Add #[tracing::instrument]

colin-grapl (approved):

src/rust/grapl-config/src/postgres.rs
```diff
@@ -0,0 +1,121 @@
+use grapl_utils::future_ext::GraplFutureExt;
+use secrecy::ExposeSecret;
+
+#[derive(Debug)]
+pub struct PostgresUrl {
```
colin-grapl, line 5: Should this be Clone?

colin-grapl (approved):

src/rust/organization-management/src/lib.rs
inickles-grapl: i was thinking that if we are to proc macro for ToPostgresUrl that we could just as well derive the fields in the first place, where we take a prefix.

```
#[derive(grapl::PostgresConfig(prefix = "EXAMPLE_SERVICE")]
```

maybe that's too much marco tho idk

src/rust/grapl-config/src/postgres.rs
inickles-grapl: sure

PR #1857, opened 2022-08-01
Fix kafka scanner

jgrillo-grapl:
NA

This resolves some flakiness we've been seeing with the integration tests.

I'm still working out _why_ exactly this fixes it. I'm not 100% sure. It had something to do with lots of spammy retries. Will work on adding a delay into the retry app in a separate PR.

Automated tests

### DIFF
--- a/src/rust/generator-dispatcher/src/lib.rs
+++ b/src/rust/generator-dispatcher/src/lib.rs
@@ -121,12 +121,21 @@ impl GeneratorDispatcher {
                                     .await
                                 {
                                     Ok(Some(generator_ids)) => {
-                                        // cache hit
-                                        enqueue_plugin_work(
-                                            plugin_work_queue_client.clone(),
-                                            generator_ids,
-                                            envelope
-                                        ).await?;
+                                        if generator_ids.is_empty() {
+                                            tracing::warn!(
+                                                message = "unrecognized event source",
+                                                tenant_id = %envelope.tenant_id(),
+                                                trace_id = %envelope.trace_id(),
+                                                event_source_id = %envelope.event_source_id(),
+                                            );
+                                        } else {
+                                            // cache hit
+                                            enqueue_plugin_work(
+                                                plugin_work_queue_client.clone(),
+                                                generator_ids,
+                                                envelope
+                                            ).await?;
+                                        }
 
                                         Ok(())
                                     },
@@ -134,6 +143,13 @@ impl GeneratorDispatcher {
                                         // cache miss, but an update was
                                         // successfully enqueued so we'll retry
                                         // the message
+                                        tracing::debug!(
+                                            message = "generator IDs cache miss, retrying message",
+                                            tenant_id = %envelope.tenant_id(),
+                                            trace_id = %envelope.trace_id(),
+                                            event_source_id = %envelope.event_source_id(),
+                                        );
+
                                         retry_message(
                                             &raw_logs_retry_producer,
                                             envelope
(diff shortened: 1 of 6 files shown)


src/rust/generator-dispatcher/src/lib.rs
```diff
@@ -121,19 +121,35 @@ impl GeneratorDispatcher {
...
-                                            plugin_work_queue_client.clone(),
-                                            generator_ids,
-                                            envelope
-                                        ).await?;
+                                        if generator_ids.is_empty() {
+                                            tracing::warn!(
+                                                message = "unrecognized event source",
+                                                tenant_id = %envelope.tenant_id(),
+                                                trace_id = %envelope.trace_id(),
+                                                event_source_id = %envelope.event_source_id(),
```
colin-grapl, line 129: I'd suggest just adding these to a span

jgrillo-grapl: ok, I'll read up on that

src/rust/generator-dispatcher/src/lib.rs
```diff
@@ -121,19 +121,35 @@ impl GeneratorDispatcher {
...
+                                        }
 
                                         Ok(())
                                     },
                                     Ok(None) => {
                                         // cache miss, but an update was
                                         // successfully enqueued so we'll retry
                                         // the message
+                                        tracing::debug!(
+                                            message = "generator IDs cache miss, retrying message",
```
wimax-grapl, line 147: maybe "sending to retry topic"

colin-grapl (approved):

src/rust/generator-dispatcher/tests/generator_dispatcher_integration_tests.rs
```diff
@@ -56,7 +56,50 @@ async fn test_dispatcher_inserts_job_into_plugin_work_queue(
     };
 
     // Send in the Raw Log Event
-    let log_event: Bytes = r#"i am some sort of raw log event"#.into();
+    let log_event: Bytes = r#"
```
wimax-grapl, line 59: could we perhaps stop copy pasting this log event :pray:

should be easy to e.g. `include_bytes!()` from a file or something

src/rust/pipeline-ingress/tests/integration_test.rs
```diff
@@ -74,14 +74,57 @@ async fn test_publish_raw_log_sends_message_to_kafka(
 ) -> Result<(), Box<dyn std::error::Error>> {
     let event_source_id = Uuid::new_v4();
     let tenant_id = Uuid::new_v4();
-    let log_event: Bytes = "test".into();
+    let log_event: Bytes = r#"
```
wimax-grapl, line 77: yeah, again

wimax-grapl (approved):

src/rust/generator-dispatcher/tests/generator_dispatcher_integration_tests.rs
jgrillo-grapl: I don't think we should. If we `include_bytes!()` it's easy to fundamentally alter the test invariants "at a distance" by changing _that file_. This would cause the assertions a test makes to change (possibly radically) without a code change which obviously impacts that test.

Also, changing that file would impact _multiple tests_ simultaneously, which seems difficult to reason about. While this event _is_ copy-pasta'd everywhere, and therefore _looks_ the same everywhere, I'm not sure it actually _is the same_.

src/rust/generator-dispatcher/src/lib.rs
jgrillo-grapl: I have a plan to do this for every message in the `kafka` lib, will implement in a follow-up PR

src/rust/generator-dispatcher/src/lib.rs
jgrillo-grapl: Will update in a follow-up PR

PR #1867, opened 2022-08-02
Warn if a multi-event log comes through sysmon-generator[-legacy]

wimax-grapl:
Currently,

sysmon-generator treats an input of
```
<Event> stuff </Event>
<Event> other stuff </Event>
```
the same as

```
<Event> stuff </Event>
```

that is really hard to debug and led me down multiple goose chases. My hope is this 'warn' here will help future devs!

### DIFF
--- a/src/rust/generators/sysmon-generator/src/api.rs
+++ b/src/rust/generators/sysmon-generator/src/api.rs
@@ -22,7 +22,9 @@ impl GeneratorApi for SysmonGenerator {
         &self,
         request: RunGeneratorRequest,
     ) -> Result<RunGeneratorResponse, Self::Error> {
-        let sysmon_event = SysmonEvent::from_str(std::str::from_utf8(&request.data)?)?;
+        let input_utf8 = std::str::from_utf8(&request.data)?;
+        let events: Vec<_> = sysmon_parser::parse_events(input_utf8).collect();
+        let sysmon_event: SysmonEvent = expect_one_event(events)?;
 
         match models::generate_graph_from_event(&sysmon_event)? {
             Some(graph_description) => Ok(RunGeneratorResponse {
@@ -35,4 +37,21 @@ impl GeneratorApi for SysmonGenerator {
             }
         }
     }
+}
+
+pub fn expect_one_event(
+    events: Vec<Result<sysmon_parser::SysmonEvent, sysmon_parser::Error>>,
+) -> Result<sysmon_parser::SysmonEvent, sysmon_parser::Error> {
+    if events.len() != 1 {
+        tracing::warn!(
+            message =
+                "sysmon-generator expects inputs of exactly 1 event - dropping all other events!",
+            found_events = events.len()
+        );
+    }
+    // Take first, or EventNotFound
+    events
+        .into_iter()
+        .next()
+        .unwrap_or_else(|| Err(sysmon_parser::Error::SysmonEventNotFound))
 }
--- a/src/rust/generators/sysmon-generator/src/main_legacy.rs
+++ b/src/rust/generators/sysmon-generator/src/main_legacy.rs
@@ -21,11 +21,11 @@ use rust_proto::graplinc::grapl::{
         RawLog,
     },
 };
-use sysmon_parser::SysmonEvent;
 use tracing::instrument::WithSubscriber;
 
 mod error;
 mod models;
+use sysmon_generator::api::expect_one_event;
 
 use crate::error::SysmonGeneratorError;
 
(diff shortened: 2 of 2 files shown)


jgrillo-grapl (approved):

inickles-grapl:
> Currently,
> 
> sysmon-generator treats an input of
> 
> ```
> <Event> stuff </Event>
> <Event> other stuff </Event>
> ```
> 
> the same as
> 
> ```
> <Event> stuff </Event>
> ```
> 
> that is really hard to debug and led me down multiple goose chases. My hope is this 'warn' here will help future devs!

This isn't true, or at least it wasn't, the generator absolutely handled multiple events just fine. Looks like it was changed at some point to only handle one event?

src/rust/generators/sysmon-generator/src/api.rs
```diff
@@ -22,7 +22,9 @@ impl GeneratorApi for SysmonGenerator {
         &self,
         request: RunGeneratorRequest,
     ) -> Result<RunGeneratorResponse, Self::Error> {
-        let sysmon_event = SysmonEvent::from_str(std::str::from_utf8(&request.data)?)?;
+        let input_utf8 = std::str::from_utf8(&request.data)?;
+        let events: Vec<_> = sysmon_parser::parse_events(input_utf8).collect();
+        let sysmon_event: SysmonEvent = expect_one_event(events)?;
```
inickles-grapl, line 27: This is not a limitation of the generator - it can produce a graph result from many events just fine.

Enforcing one event here is purely artificial? Seems better served elsewhere.

inickles-grapl (changes requested):

src/rust/generators/sysmon-generator/src/api.rs
inickles-grapl: Looks like this has indeed been changed to only process a single event (https://github.com/grapl-security/grapl/blob/main/src/rust/generators/sysmon-generator/src/api.rs#L27),

but IMO we should revert the change that made it only handle one at a time, and if we want to enforce that we should do it elsewhere and keep this as flexible as possible.

jgrillo-grapl: I think where we ended up in our discussion today was that the pipeline (e.g. everything after pipeline-ingress) operates on single events. So in the absence of a technical requirement to support processing more than one event per `RawLog`, and with the clear expectation that clients won't send more than one event per `RawLog`, it's hard to see the value in maintaining code which _expects_ more than one event per `RawLog`... that seems like it'll suffer bitrot.

See https://grapl-internal.slack.com/archives/C03JNJ57A9G/p1659457692934169

inickles-grapl: The previous version did not expect more than one event, it was flexible in being able to handle one or more events. If the backend wanted to send a single event as a time, that was fine, the code worked just the same and there was no additional maintenance needed for that flexibility, and there were no bits rotting.

If anything, this PR is additional maintenance, one to help enforce a requirement of a different system, which IMO it shouldn't do.

inickles-grapl: It looks like the discussion you linked to runs too deep for this PR tho, I'll remove my request for changes.

jgrillo-grapl: Yeah I think it's fine if this code can handle more than one event, we just need to make it sure with tests that exercise both the one event and the many events case.

inickles-grapl: > In that case, a parser which understands more than one event at a time might make sense?

The parser and generator operated on multiple events, which it got from `sysmon_parser::parse_events`, just as it would now in this PR. Previously: 
https://github.com/grapl-security/grapl/blob/2e1558f47fc01cb5e249dd46624a4686e7d928ff/src/rust/generators/sysmon-generator/src/serialization.rs#L16

where the generator folded multiple `GraphDescription`s into a single: 
https://github.com/grapl-security/grapl/blob/2e1558f47fc01cb5e249dd46624a4686e7d928ff/src/rust/generators/sysmon-generator/src/generator.rs#L86

inickles-grapl (approved):
Approving to unblock - we can change things later if desired.

src/rust/generators/sysmon-generator/src/api.rs
inickles-grapl: > Yeah I think it's fine if this code can handle more than one event, we just need to make it sure with tests that exercise both the one event and the many events case.
> 
> And, to be clear, the original change which surfaced this bug was my mistake 😞. I was in a hurry to get a generator running in the Kafka pipeline and made a bad call.

All good, no worries - we can revisit this when we're ready, but .. I don't think today is that day. 👍

wimax-grapl:
Ian let's chat in the morning. A large discussion was had on this (should generator input be fat or skinny? stored directly in kafka or off in S3?) in the slack channel today

PR #1871, opened 2022-08-04
More than one partition

jgrillo-grapl:
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/992

This PR updates the local Grapl deployment to operate Kafka under a better approximation of realistic conditions. In particular, we now have a cluster of 3 brokers, and we set `min.insync.replicas=2`, `partitions=2`, and `replication=3` on each topic. Additionally, we now run 2 instances of each consumer service.

Also, we're now committing our offsets in the `kafka` library. Also, before we had set `auto.offset.reset="earliest"`, now we use `"latest"`. I've also made a bunch of updates to the `KafkaTopicScanner`:

1. Now we have a unique consumer group for every `KafkaTopicScanner` instance. This ensures that each scanner consumes messages from every partition.
2. The `KafkaTopicScanner#contains*` methods return a `JoinHandle<Vec<Envelope<T>>>` containing _all_ the messages matching the predicate instead of just the first one. This makes the tool a little more versatile because now we can make assertions about things like the number of matching messages and interrelationships between messages. Also, the `KafkaTopicScanner` does no error handling internally. Every error panics. These panics are propagated to the test function so we can see useful things like the panic message, stack traces, etc.
3. The `KafkaTopicScanner` now commits its offsets.

Automated tests


nomad/local/grapl-local-infra.nomad
```diff
@@ -3,24 +3,60 @@ variable "image_tag" {
   description = "The tag for all container images we should deploy. This is ultimately set in the top-level Makefile."
 }
 
-variable "kafka_broker_port" {
+variable "kafka0_broker_port" {
```
wimax-grapl, line 6: I think I'd much prefer a 
```
variable "kafka_broker_args" {
  type = list(object({
    broker_port_listen = number
    broker_port_host_os = number
    jmx_port = number
  }))
}
```

but I won't push the matter;  notably this would require you constructing that arg as json in startdetach

nomad/local/grapl-local-infra.nomad
```diff
@@ -201,29 +237,41 @@ job "grapl-local-infra" {
...
 
       resources {
-        memory = 2000
+        memory = 1200
       }
 
       env {
-        kafka_broker_port       = 9092 # Only used by healthcheck
-        KAFKA_BROKER_ID         = 1
+        kafka0_broker_port      = 9092 # Only used by healthcheck
```
wimax-grapl, line 273: is this ever used?

nomad/local/grapl-local-infra.nomad
```diff
@@ -270,7 +318,132 @@ job "grapl-local-infra" {
           }
         }
       }
+    }
+
+    task "kafka1" {
```
wimax-grapl, line 323: seeing how much is repeated here I'd really prefer if we used a `dynamic "task" {}` which is literally perfect for these purposes. it iterates over the aforementioned `variable "kafka_broker_args"` and you can inject `i.broker_port`, `i.jmx_port` etc.

wimax-grapl: see `    for_each = local.database_descriptors` for an example

nomad/local/grapl-local-infra.nomad
```diff
@@ -380,6 +553,10 @@ job "grapl-local-infra" {
               }
             }
           }
+
+          resources {
+            memory = 500
+          }
```
wimax-grapl, line 559: remove so we can see if the other postgres fix did the trick?

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
```diff
@@ -1,4 +1,6 @@
-#![cfg(feature = "integration_tests")]
+//#![cfg(feature = "integration_tests")]
```
wimax-grapl, line 1: this seems wrong

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
```diff
@@ -41,40 +51,55 @@ async fn test_sysmon_log_e2e(ctx: &mut E2eTestContext) -> Result<(), Box<dyn std
...
+        Duration::from_secs(30),
+    )
+    .contains_for_tenant(tenant_id, 36, |_graph: GraphDescription| true)
+    .await;
+
+    let node_identifier_scanner_handle = KafkaTopicScanner::new(
+        ConsumerConfig::with_topic("identified-graphs"),
+        Duration::from_secs(30),
+    )
+    .contains_for_tenant(tenant_id, 36, |_graph: IdentifiedGraph| true)
```
wimax-grapl, line 72: can we bring back my predicates?

wimax-grapl: oh i see they're done below in-line, which i'm happy about; but i'm a bit confused about the role of the predicate here now if we do filtering later on

src/rust/kafka/src/lib.rs
```diff
@@ -493,9 +527,21 @@ where
...
+                if let Ok(_) = result {
+                    Ok(self.consumer.commit()?)
+                } else {
+                    result
+                }
+            })
     }
 }
 
+/// A RetryProcessor handles messages on a service's retry topic, imposes a
```
wimax-grapl, line 540: thank you for the additional docs

nomad/local/grapl-local-infra.nomad
jgrillo-grapl: Oh awesome, thanks! I'll do that and use a dynamic task.

src/rust/kafka/src/lib.rs
```diff
@@ -528,5 +584,12 @@ impl RetryProcessor {
                     result.map(|_| ())
                 }
             })
+            .then(|result| async {
+                if let Ok(_) = result {
+                    Ok(self.consumer.commit()?)
+                } else {
+                    result
+                }
```
wimax-grapl, line 592: i think `result.and_then(|| self.consumer.commit())` would do the trick but it's fine as is

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
jgrillo-grapl: Yeah, the idea is to just do _filtering_ in the `contains`, then make assertions on the filtered results. It just makes it a lot easier to tell what went wrong when tests fail.

I think maybe I should change the `contains` names to be `filter`, actually. Does that sound OK?

src/rust/kafka/src/test_utils/topic_scanner.rs
```diff
@@ -9,133 +13,184 @@ use rust_proto::{
...
-    > {
+        filter_predicate: impl FnMut(Envelope<T>) -> bool + Send + Sync + 'static,
+        stop_predicate: impl FnMut(usize, Envelope<T>) -> bool + Send + Sync + 'static,
+    ) -> JoinHandle<Vec<Envelope<T>>> {
         // we'll use this channel to communicate that the consumer is ready to
         // consume messages
         let (tx, rx) = oneshot::channel::<()>();
-
-        let predicate = std::sync::Arc::new(std::sync::Mutex::new(predicate));
+        let tx_mutex = Mutex::new(Some(tx));
```
wimax-grapl, line 108: gonna skip reviewing this part re: rx-tx, but the stop/filter predicate _is_ cool

wimax-grapl: i guess `contains` is no longer the right term though, maybe just `scan` or `filter`. we can fix that up later though

nomad/local/grapl-local-infra.nomad
jgrillo-grapl: Yeah it's used in the healthcheck on L309.

src/rust/node-identifier/tests/integration_test.rs
```diff
@@ -144,10 +113,39 @@ async fn test_sysmon_event_produces_identified_graph(
         .expect("received error response");
 
     tracing::info!("waiting for kafka_scanner to complete");
-    kafka_scanner
-        .get_listen_result()
-        .instrument(tracing::debug_span!("kafka_scanner"))
-        .await??;
+    let envelopes = handle.await?;
```
wimax-grapl, line 116: this is so much more ergonomic :+1:

src/rust/pipeline-ingress/tests/integration_test.rs
```diff
@@ -118,36 +114,43 @@ async fn test_publish_raw_log_sends_message_to_kafka(
...
-            envelope_tenant_id == tenant_id
-                && envelope_event_source_id == event_source_id
-                && raw_log.log_event() == expected_log_event
-        })
-        .await?;
+    tracing::info!(
+        message = "sending publish_raw_log request",
+        tenant_id =% tenant_id,
+        event_source_id =% event_source_id,
+    );
```
wimax-grapl, line 131: doesn't the span fill in those two things for us already?

jgrillo-grapl: The span only exists in places where we've consumed an Envelope<T> from Kafka. It's automatically applied to everything that runs in the StreamProcessor's closure, but that's about it. Here we don't have it yet because we don't consume anything (that happens in the topic scanner).

nomad/local/grapl-local-infra.nomad
jgrillo-grapl: done

nomad/local/grapl-local-infra.nomad
jgrillo-grapl: done

nomad/local/grapl-local-infra.nomad
jgrillo-grapl: done, still seeing issues that appear to be caused by lack of postgres memory. In particular, seeing this repeatedly in the `plugin-work-queue-db` error logs: 
```
2022-08-08 17:49:46.642 UTC [1] LOG:  server process (PID 111) was terminated by signal 9: Killed
```

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
jgrillo-grapl: done

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
jgrillo-grapl: done. Changed them to `scan` and `scan_for_tenant`

src/rust/kafka/src/lib.rs
jgrillo-grapl: done

src/rust/kafka/src/test_utils/topic_scanner.rs
jgrillo-grapl: fixed

src/rust/node-identifier/tests/integration_test.rs
jgrillo-grapl: thanks!

wimax-grapl (approved):
approved assuming we can get the postgres stuff working

src/rust/kafka/src/lib.rs
```diff
@@ -578,12 +578,6 @@ impl RetryProcessor {
...
                 }
             })
-            .then(|result| async {
-                if let Ok(_) = result {
-                    Ok(self.consumer.commit()?)
-                } else {
-                    result
-                }
-            })
+            .then(move |result| async { result.and_then(|_| Ok(self.consumer.commit()?)) })
```
wimax-grapl, line 581: I'd imagine you could remove the Ok() and the ?

wimax-grapl: but if not, whatever

wimax-grapl: maybe the ? is doing some conversion for us

wimax-grapl (approved):

src/rust/kafka/src/lib.rs
jgrillo-grapl: Yes, the `?` is converting the `CommitError` for us

nomad/local/nomad-agent-conf.nomad
```diff
@@ -42,6 +41,10 @@ client {
     # Turn on consul connect proxy debug logs. Consul connect sidecars now have access logs, etc.
     connect.log_level = "debug"
   }
+  gc_interval              = "30m"
+  gc_disk_usage_threshold  = 99
```
twunderlich-grapl, line 45: We're tuning nomad gc to prevent logs from being collected, right? Might be nice to put a comment pertaining to that above.

jgrillo-grapl: Yeah, will do

jgrillo-grapl: done

ghost (approved):
Approving the Nomad, postgres config and pulumi code. I'll take a quick scan through the rust code, but I imagine that you'll get better reviews from Max et al

PR #1849, opened 2022-07-29
Client executor

colin-grapl:
draft pr

PR #1872, opened 2022-08-05
'client-executor' crate for performing retries on grpc clients; port Plugin Registry to this pattern

wimax-grapl:
Resurrecting Colin's pr here.
https://github.com/grapl-security/grapl/pull/1849

I cannot speak to literally anything in `client-executor` myself, but I can speak about the rust-proto macro change - a lot of it is not really a good fit for interfaces and is a much better fit for a repeatable macro. 

I haven't tackled the streaming RPC yet, but that's an edge case!

### DIFF
--- a/src/rust/rust-proto/src/client_macros.rs
+++ b/src/rust/rust-proto/src/client_macros.rs
@@ -16,19 +16,28 @@ macro_rules! execute_client_rpc {
         $native_response_type: ty,
     ) => {{
         {
+            let backoff = client_executor::strategy::FibonacciBackoff::from_millis(100);
+            let num_retries = 10;
+
             let proto_request = <$proto_request_type>::try_from($native_request)?;
+
+            // We can revisit this; potentially passing in a retry_condition
+            // per-RPC and not globally applied.
+            let retry_condition = |status: &tonic::Status| {
+                // Only retry if the status code is Internal Error.
+                status.code() == tonic::Code::Internal
+            };
 
             let proto_response = $self
                 .executor
-                .spawn(
-                    client_executor::strategy::FibonacciBackoff::from_millis(10)
-                        .map(jitter)
-                        .take(20),
+                .spawn_conditional(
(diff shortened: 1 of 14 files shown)


src/rust/rust-proto/src/graplinc/grapl/api/plugin_registry/v1beta1_client.rs
```diff
@@ -41,6 +43,15 @@ pub enum PluginRegistryServiceClientError {
     Elapsed,
 }
 
+// This likely should be removed in favor of making all SerDe conversions
+// TryFrom<Error = SerDeError> instead of From.
+// Currently it's kind of scattershot.
```
jgrillo-grapl, line 48: It's not scattershot. The things that are fallible are `TryFrom` and the things that are infallible are `From`. If this code needs to paper over that, then this is a fine approach. I don't think we should compromise the clarity of the `From`/`TryFrom` distinction in the actual SerDe types to accommodate things like this though.

src/rust/rust-proto/src/client_macros.rs
```diff
@@ -0,0 +1,38 @@
+/// This macro implements boilerplate code to:
+/// - translate between the native types and tonic/prost types
+///   in the transport layer
+/// - Automatically hook these client RPCs up to the Executor, which provides
+///   retries (among other good behaviors).
+///
+/// It makes some expectations of the structure of `self` - like that it
+/// has a `self.executor` and `self.proto_client`.
+#[macro_export]
+macro_rules! execute_client_rpc {
```
jgrillo-grapl, line 10: Can we call this `execute_rpc` instead? The name `execute_client_rpc` sounds like it's executing a remote procedure call _against the client_ which doesn't really make sense. Instead, we're executing a remote procedure call against the server.

jgrillo-grapl (approved):
👍 just minor nits

PR #1874, opened 2022-08-08
Disable ANSI color codes in rust logs

wimax-grapl:
https://doc.rust-lang.org/cargo/reference/config.html#termcolor

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -39,3 +39,7 @@ inherits = "dev"
 # Results in a ~20x binary size reduction, for example
 # sysmon-generator: 279M -> 16M
 strip = "debuginfo"
+
+[term]
+# Disable ANSI color codes in test logs
+color = "never"


jgrillo-grapl (approved):

src/rust/Cargo.toml
```diff
@@ -39,3 +39,7 @@ inherits = "dev"
 # Results in a ~20x binary size reduction, for example
 # sysmon-generator: 279M -> 16M
 strip = "debuginfo"
+
+[term]
+# Disable ANSI color codes in test logs
+color = "never"
```
christophermaier, line 45: Would it be better to preserve the use of ANSI color codes in general (e.g., when working locally), but disable them in CI using the `CARGO_TERM_COLOR` environment variable?

wimax-grapl: Good point/idea; I'm actually coming across them in the following workflow:
1. run make test-integration-rust
2. inspect artifacts in test_artifacts

which is basically the same flow whether or not it's local or CI

PR #1881, opened 2022-08-10
use "client-executor" for Generator client

wimax-grapl:
https://github.com/grapl-security/grapl/pull/1872 introduced the client-executor crate, and introduced using it for plugin-registry. Now I'm porting generator to it.

I made a macro that deduplicates some boilerplate for the "get a proto client" stuff.

in the course of adding this to generator I realized the error types were exactly the same, so, I standardized.

### DIFF
--- a/src/rust/rust-proto/src/client_macros.rs
+++ b/src/rust/rust-proto/src/client_macros.rs
@@ -48,12 +48,12 @@ macro_rules! execute_client_rpc {
     }};
 }
 
-/// This macro implements boilerplace code to connect to
+/// This macro implements boilerplate code to connect to
 /// a gRPC service (and do retries if needed).
 /// Unfortunately, each $proto_client_type doesn't share traits, so a
 /// macro is the quickest solution.
 #[macro_export]
-macro_rules! get_proto_client {
+macro_rules! create_proto_client {
     (
         $executor: ident,
         $proto_client_type: ty,
--- a/src/rust/rust-proto/src/client_macros.rs
+++ b/src/rust/rust-proto/src/client_macros.rs
@@ -31,7 +31,9 @@ macro_rules! execute_client_rpc {
             let proto_response = $self
                 .executor
                 .spawn_conditional(
-                    backoff.map(jitter).take(num_retries),
+                    backoff
+                        .map(client_executor::strategy::jitter)
+                        .take(num_retries),
                     || {
                         let mut proto_client = $self.proto_client.clone();
                         let proto_request = proto_request.clone();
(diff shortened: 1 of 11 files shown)


src/rust/rust-proto/src/client_macros.rs
```diff
@@ -45,3 +47,38 @@ macro_rules! execute_client_rpc {
...
+/// macro is the quickest solution.
+#[macro_export]
+macro_rules! get_proto_client {
+    (
+        $executor: ident,
+        $proto_client_type: ty,
+        $endpoint: ident,
+    ) => {{
+        {
+            let backoff = client_executor::strategy::FibonacciBackoff::from_millis(10);
```
jgrillo-grapl, line 63: This is an excessive delay.. the 20th Fibonacci number is 4181 or 6765, depending on how you start counting.. if we're multiplying that by _ten_ here that's a 41s to 67s delay.

I think starting with a 10ms delay and backing off according to this Fibonacci strategy is fine, but we should
1. make sure each delay gets some random jitter added to it on the order of magnitude of the _previous_ step in the sequence and
2. make sure the total delay is capped at a reasonable value (e.g. 5s + jitter)

jgrillo-grapl (approved):
LGTM in form, but I have a few specific comments. Won't block for now if this is urgent, we can fix these things later.

src/rust/rust-proto/src/client_macros.rs
wimax-grapl: I agree the numbers here are funky and in general I prefer exponential; I was just reusing the numbers colin gave me

wimax-grapl: For a total timeout of 5000ms I'm doing
(fibonacci of 10 total) = 1 + 1 + 2 + 3 + 5 + 8 + 13 + 21 + 34 + 55 = 143
5000/143 = 35

so 35ms taken 10 times.

src/rust/rust-proto/src/client_macros.rs
```diff
@@ -45,3 +47,38 @@ macro_rules! execute_client_rpc {
         }
     }};
 }
+
+/// This macro implements boilerplace code to connect to
+/// a gRPC service (and do retries if needed).
+/// Unfortunately, each $proto_client_type doesn't share traits, so a
+/// macro is the quickest solution.
+#[macro_export]
+macro_rules! get_proto_client {
```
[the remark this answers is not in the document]
wimax-grapl: done

src/rust/rust-proto/src/client_macros.rs
jgrillo-grapl: What I mean by "capped at 5s" is the _longest_ timeout in the sequence should be 5s but once we hit that it shouldn't increase further. So it should go 35ms + O(1ms) jitter, 35ms + O(1ms) jitter, 70ms + O(1ms) jitter, 105ms + O(10ms) jitter ..., 5000ms + O(100ms) jitter, 5000ms + O(100ms) jitter, 5000ms + O(100ms) jitter...

src/rust/rust-proto/src/client_macros.rs
```diff
@@ -45,3 +47,38 @@ macro_rules! execute_client_rpc {
         }
     }};
 }
+
+/// This macro implements boilerplace code to connect to
```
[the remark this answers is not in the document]
wimax-grapl: done

src/rust/rust-proto/src/client_macros.rs
wimax-grapl: got it got it

PR #1884, opened 2022-08-11
One-time, not automated: use 'cargo sort --workspace' to sort deps

wimax-grapl:
this is likely easily automated in the future, but i think for now this is fine. i'd hate to reject a build just because my precious deps were out of order tbh


src/rust/plugin-sdk/generator-sdk/Cargo.toml
```diff
@@ -21,12 +21,12 @@ grapl-tracing = { path = "../../grapl-tracing" }
 moka = { version = "0.8", features = ["future"] }
 rust-proto = { path = "../../rust-proto" }
 serde = { version = "1.0", features = ["derive"] }
+# Only for test_utils
+test-context = { version = "0.1", optional = true }
```
wimax-grapl, line 25: i was very impressed it moved the comments correctly too!

jgrillo-grapl (approved):
this would be great to have as an automated lint

PR #1886, opened 2022-08-12
Move most RPCs tonic error types to Unknown, not Internal. Let each RPC specify which statuscode to retry on.

wimax-grapl:
As discussed around here: https://grapl-internal.slack.com/archives/C03JNJ57A9G/p1660317993208369

I arbitrarily decided to just do it based on Code; you're _RARELY_ looking at any other property of a Status. Like if you have to inspect the message, that seems bad


jgrillo-grapl (approved):
seems legit

PR #1891, opened 2022-08-14
increase a bunch of timeouts...

jgrillo-grapl:
NA

This PR attempts to fix the integration tests by increasing a bunch of timeouts. So far it's not working. The e2e test is failing locally with 
```
---- test_sysmon_log_e2e stdout ----
Error: ErrorStatus(Status { code: Internal, message: "Timeout expired" })
thread 'test_sysmon_log_e2e' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/test/src/lib.rs:185:5
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at ./rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/panicking.rs:142:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at ./rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/panicking.rs:181:5
   4: test::assert_test_result
             at ./rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/test/src/lib.rs:185:5
   5: sysmon_log_e2e_test::test_sysmon_log_e2e::{{closure}}
             at ./grapl/rust/e2e-tests/tests/sysmon_log_e2e_test.rs:42:1
   6: core::ops::function::FnOnce::call_once
             at ./rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/ops/function.rs:248:5
   7: core::ops::function::FnOnce::call_once
             at ./rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/ops/function.rs:248:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    test_sysmon_log_e2e

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 61.79s
```

### DIFF
--- a/src/rust/rust-proto/src/client_macros.rs
+++ b/src/rust/rust-proto/src/client_macros.rs
@@ -32,10 +32,10 @@ macro_rules! execute_client_rpc {
         $opts: expr,
     ) => {{
         {
-            let backoff = client_executor::strategy::FibonacciBackoff::from_millis(100)
+            let backoff = client_executor::strategy::FibonacciBackoff::from_millis(10)
                 .max_delay(Duration::from_millis(5000))
                 .map(client_executor::strategy::jitter);
-            let num_retries = 10;
+            let num_retries = 1000;
 
             let proto_request = <$proto_request_type>::try_from($native_request)?;
 
@@ -52,7 +52,11 @@ macro_rules! execute_client_rpc {
                     || {
                         let mut proto_client = $self.proto_client.clone();
                         let proto_request = proto_request.clone();
-                        async move { proto_client.$rpc_name(proto_request).await }
+                        async move {
+                            let mut tonic_request = tonic::Request::new(proto_request);
+                            tonic_request.set_timeout(Duration::from_secs(60));
+                            proto_client.$rpc_name(tonic_request).await
+                        }
                     },
                     executor_retry_condition,
                 )
--- a/src/rust/rust-proto/src/graplinc/grapl/api/pipeline_ingress/v1beta1/server.rs
+++ b/src/rust/rust-proto/src/graplinc/grapl/api/pipeline_ingress/v1beta1/server.rs
@@ -145,6 +145,9 @@ where
             )
             .await;
         Ok(Server::builder()
+            .max_concurrent_streams(100)
+            .concurrency_limit_per_connection(1)
+            .timeout(Duration::from_secs(60))
             .add_service(health_service)
             .add_service(PipelineIngressServiceServerProto::new(GrpcApi::new(
                 self.api_server,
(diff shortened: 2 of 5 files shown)


colin-grapl (approved):
Approved, but I'd suggest also increasing the error threshold to something like 95% in tests.

jgrillo-grapl:
> Approved, but I'd suggest also increasing the error threshold to something like 95% in tests.

What is the error threshold?

src/rust/rust-proto/src/client_macros.rs
```diff
@@ -32,10 +32,10 @@ macro_rules! execute_client_rpc {
         $opts: expr,
     ) => {{
         {
-            let backoff = client_executor::strategy::FibonacciBackoff::from_millis(100)
+            let backoff = client_executor::strategy::FibonacciBackoff::from_millis(10)
                 .max_delay(Duration::from_millis(5000))
                 .map(client_executor::strategy::jitter);
-            let num_retries = 10;
+            let num_retries = 1000;
```
wimax-grapl, line 38: 1000 retries feels wildly incorrect, especially when after fib(20)*10 of them each retry will delay up to 5sec long.

If this keeps retrying we'd just find outselves waiting for ~5sec * 980

wimax-grapl (changes requested):

src/rust/rust-proto/src/client_macros.rs
colin-grapl: Agree, if an RPC is failing for even a matter of minutes I think we're beyond "retry" and we're in "a dependency is failing, we are no longer healthy".

colin-grapl (changes requested):
Let's keep num_retries at 100, I don't think we want to block on RPCs for hours.

src/rust/rust-proto/src/client_macros.rs
jgrillo-grapl: Yes, I agree as well. This was an attempt to eliminate this variable as a possible cause for the observed failure.

wimax-grapl:
even then colin it'd be ~= 5sec * 80 or 6 minutes? for a single RPC?

jgrillo-grapl:
Closing this because it didn't help to isolate the error

PR #1893, opened 2022-08-15
Slice out the rust-proto part of Graph Query Service PR

wimax-grapl:
Just updates the messages in rust-proto to account for Graph Query Service. Colin asked I do this separately to unblock people.

### DIFF
--- a/src/proto/graplinc/grapl/api/graph_query_service/v1beta1/graph_query_service.proto
+++ b/src/proto/graplinc/grapl/api/graph_query_service/v1beta1/graph_query_service.proto
@@ -1,0 +1,369 @@
+syntax = "proto3";
+
+package graplinc.grapl.api.graph_query_service.v1beta1;
+
+import "graplinc/common/v1beta1/types.proto";
+import "graplinc/grapl/api/graph/v1beta1/types.proto";
+import "graplinc/grapl/common/v1beta1/types.proto";
+
+// IntegerProperty wraps one of the underlying integer property
+// types
+message IntegerProperty {
+  // The inner property
+  oneof property {
+    // An integer that only grows
+    graph.v1beta1.IncrementOnlyIntProp increment_only = 1;
+    // An integer that only shrinks
+    graph.v1beta1.DecrementOnlyIntProp decrement_only = 2;
+    // An integer that never changes after its first write
+    graph.v1beta1.ImmutableIntProp immutable = 3;
+  }
+}
+
+// A filter for querying an integer
+message IntFilter {
+  // The operation to filter with
+  enum Operation {
+    // An unknown operation
+    OPERATION_UNSPECIFIED = 0;
+    // If the property is set at all
+    OPERATION_HAS = 1;
+    // If the property is equal to `value`
+    OPERATION_EQUAL = 2;
+    // If the property is less than `value`
+    OPERATION_LESS_THAN = 3;
+    // If the property is less than or equal to `value`
+    OPERATION_LESS_THAN_OR_EQUAL = 4;
+    // If the property is greater than `value`
+    OPERATION_GREATER_THAN = 5;
+    // If the property is greater than or equal to `value`
+    OPERATION_GREATER_THAN_OR_EQUAL = 6;
+  }
+  // The filter operation to apply
+  Operation operation = 1;
+  // The value to compare against. Unset if operation is `Has`
+  int64 value = 2;
+  // Negation of the operation
+  // defaults to `false`
+  bool negated = 3;
+}
+
+// AndIntFilters represents a group of filters that must all
+// be satisfied in order to match
+message AndIntFilters {
+  // The internal filters
+  repeated IntFilter int_filters = 1;
+}
+
+// `OrIntFilters` represents groups of 'And'ed constraints, of which
+// any group can match.
+message OrIntFilters {
+  // The groups of And filters
+  repeated AndIntFilters and_int_filters = 1;
+}
+
+// `StringFilter` represents a filter against a string property
+message StringFilter {
+  // The operation to apply as a filter
+  enum Operation {
+    // The operation is unspecified
+    OPERATION_UNSPECIFIED = 0;
+    // If the property is set at all
+    OPERATION_HAS = 1;
+    // If the property is equal to `value`
+    OPERATION_EQUAL = 2;
+    // If the property contains `value`
+    OPERATION_CONTAINS = 3;
+    // If the property matches the regex `value`
+    OPERATION_REGEX = 4;
+  }
+  // The operation to apply as a filter
+  Operation operation = 1;
+  // The value to compare against, or an empty string if `Has`
+  string value = 2;
+  // Whether to negate the filter or not
+  // defaults to `false`
+  bool negated = 3;
+}
+
+// A group of StringFilter that must all match to satisfy a query
+message AndStringFilters {
+  // The underlying filters
+  repeated StringFilter string_filters = 1;
+}
+
+// A group of AndStringFilters for which only one group must match
+// in order to satisfy a query
+message OrStringFilters {
+  // The underlying filters
+  repeated AndStringFilters and_string_filters = 1;
+}
+
+// UidFilter represents a filter operation on a `Uid`
+message UidFilter {
+  // The operation to apply
+  enum Operation {
+    // If the operation is unknown or unset
+    OPERATION_UNSPECIFIED = 0;
+    // If the uid property is equal to `value`
+    OPERATION_EQUAL = 1;
+  }
+  // The operation to apply as a filter
+  Operation operation = 1;
+  // The value to compare against
+  graplinc.grapl.common.v1beta1.Uid value = 2;
+}
+
+// The filters to apply, of which if any match the query is satisfied
+message UidFilters {
+  // The underlying uid filters
+  repeated UidFilter uid_filters = 1;
+}
+
+// QueryId is a unique identifier for a node query
+message QueryId {
+  // A non-zero value, guaranteed to be unique within a single query
+  uint64 value = 1;
+}
+
+// A NodePropertyQuery is a group of filters for a node
+// where all filters must match in order for the overall query
+// to be satisfied
+message NodePropertyQuery {
+  // The type of the node
+  graplinc.grapl.common.v1beta1.NodeType node_type = 1;
+  // The unique identifier of this query within an overall GraphQuery
+  QueryId query_id = 2;
+  // A mapping of string property names to OrStringFilters
+  // Note that the key is `string` but when serializing and deserializing
+  // the keys must be treated as PropertyName
+  map<string, OrStringFilters> string_filters = 3;
+  // A mapping of integer property names to OrIntFilters
+  // Note that the key is `string` but when serializing and deserializing
+  // the keys must be treated as PropertyName
+  map<string, OrIntFilters> int_filters = 4;
+  // todo: Add unsigned 64bit property types
+  // The UidFilters
+  UidFilters uid_filters = 5;
+}
+
+// An entry in a map, consisting of composite keys and a value
+// Used in the `EdgeQueryMap`
+message EdgeQueryEntry {
+  // The QueryId for the source of this edge
+  // Part of the key
+  QueryId query_id = 1;
+  // The name of the edge
+  // Part of the key
+  graplinc.grapl.common.v1beta1.EdgeName edge_name = 2;
+  // The QueryIds for destination node queries
+  repeated QueryId neighbor_query_ids = 3;
+}
+
+// A Map of (QueryId, EdgeName) to List[QueryId]
+message EdgeQueryMap {
+  // The underlying map entries, stored as a list
+  repeated EdgeQueryEntry entries = 1;
+}
+
+// An entry in an EdgeNameMap, holding the associated forward
+// and reverse edge names
+message EdgeNameEntry {
+  // The forward edge name
+  // Key
+  graplinc.grapl.common.v1beta1.EdgeName forward_edge_name = 1;
+  // The reverse edge name
+  // Value
+  graplinc.grapl.common.v1beta1.EdgeName reverse_edge_name = 2;
+}
+
+// A Map of EdgeName to associated EdgeName
+message EdgeNameMap {
+  // The underlying map entries, encoded as a list
+  repeated EdgeNameEntry entries = 1;
+}
+
+// An entry in the NodePropertyQueryMap, associating
+// a QueryId with a NodePropertyQuery
+message NodePropertyQueryEntry {
+  // The unique identifier for this NodePropertyQuery
+  // The Key
+  QueryId query_id = 1;
+  // The NodePropertyQuery itself
+  // The value
+  NodePropertyQuery node_property_query = 2;
+}
+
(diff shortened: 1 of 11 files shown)


src/proto/graplinc/grapl/api/graph_query_service/v1beta1/graph_query_service.proto
```diff
@@ -0,0 +1,369 @@
...
+  repeated StringProperty properties = 1;
+}
+
+// IntProperty wraps a property name and its associated integer value
+message IntProperty {
+  // The name of this property
+  graplinc.grapl.common.v1beta1.PropertyName property_name = 1;
+  // The value
+  int64 property_value = 2;
+}
```
wimax-grapl, line 241: Notably, this is seemingly never used. @colin-grapl ?

src/rust/rust-proto/src/graplinc/grapl/api/graph_query_service/v1beta1/messages.rs
```diff
@@ -0,0 +1,1351 @@
...
+        }
+    }
+}
+
+impl From<StringCmp> for StringFilter {
+    fn from(string_cmp: StringCmp) -> StringFilter {
+        match string_cmp {
+            StringCmp::Has => StringFilter {
+                operation: StringOperation::Has,
+                value: "".to_string(),
```
wimax-grapl, line 241: @colin-grapl  what's this about

src/rust/rust-proto/src/graplinc/grapl/api/graph_query_service/v1beta1/messages.rs
```diff
@@ -0,0 +1,1351 @@
...
+        Self {}
+    }
+}
+
+impl From<NoMatchWithUid> for proto::NoMatchWithUid {
+    fn from(value: NoMatchWithUid) -> Self {
+        let NoMatchWithUid {} = value;
+        Self {}
+    }
+}
```
wimax-grapl, line 1229: why the let statement?

src/proto/graplinc/grapl/api/graph_query_service/v1beta1/graph_query_service.proto
```diff
@@ -0,0 +1,340 @@
...
+
+import "graplinc/common/v1beta1/types.proto";
+import "graplinc/grapl/common/v1beta1/types.proto";
+
+// A filter for querying an integer
+message IntFilter {
+  // The operation to filter with
+  enum Operation {
+    // An unknown operation
+    OPERATION_UNSPECIFIED = 0;
```
christophermaier, line 13: Is the `OPERATION_` prefix really needed on all these? I'd think the fact that they're all part of an `Operation` enum would be sufficient.

src/rust/rust-proto/src/graplinc/grapl/api/graph_query_service/v1beta1/server.rs
```diff
@@ -0,0 +1,190 @@
...
+        Self {
+            server: GraphQueryServiceServerProto::new(service),
+            addr,
+            shutdown_rx,
+        }
+    }
+
+    /// Consumes the builder and returns a new `GraphQueryServiceServer`.
+    /// Note: Panics on invalid build state
+    pub fn build(self) -> GraphQueryServiceServer<T, E> {
```
christophermaier, line 183: I'm not clear on what this builder is actually doing... it looks like it's just another name for the server it's supposed to be building. Am I missing some magic somewhere?

src/proto/graplinc/grapl/api/graph_query_service/v1beta1/graph_query_service.proto
```diff
@@ -0,0 +1,340 @@
...
+  oneof inner {
+    MatchedGraphWithUid matched = 1;
+    NoMatchWithUid missed = 2;
+  }
+}
+
+// The Response associated with a QueryGraphWithUidRequest
+// containing a matched graph and the associated uid
+message QueryGraphWithUidResponse {
+  MaybeMatchWithUid maybe_match = 1;
```
colin-grapl, line 310: Missing comment

src/proto/graplinc/grapl/api/graph_query_service/v1beta1/graph_query_service.proto
```diff
@@ -0,0 +1,340 @@
...
+  // This is *not* necessarily the `node_uid` from the request
+  graplinc.grapl.common.v1beta1.Uid root_uid = 2;
+}
+
+// Represents a 'miss' for a QueryGraphWithUidRequest
+message NoMatchWithUid {
+  // empty
+}
+
+message MaybeMatchWithUid {
```
colin-grapl, line 299: Missing comment

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
```diff
@@ -0,0 +1,245 @@
...
+
+impl TryFrom<String> for PropertyName {
+    type Error = &'static str;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        // todo: Validate this thing
+        Ok(PropertyName { value })
+    }
+}
+
+impl TryFrom<PropertyNameProto> for PropertyName {
```
christophermaier, line 61: Why is this `TryFrom` when it's infallible?

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        Ok(Self {
+            value: raw.to_owned(),
+        })
+    }
+}
+
+impl TryFrom<String> for PropertyName {
+    type Error = &'static str;
```
christophermaier, line 54: Shouldn't this also be a `SerDeError`, in keeping with the above `TryFrom` implementation?

This actually goes for all such conversion traits in this file... what determines that one has an Error type of `SerDeError`, while another uses `&'static str`?

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
```diff
@@ -0,0 +1,245 @@
...
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        // todo: Validate this thing
+        Ok(PropertyName { value })
```
christophermaier, line 57: Does the above `TryFrom` _have_ to be on a `&'static str`, or can it be just on a `&str`? If the latter, you could use the same validation logic here.

christophermaier: Or maybe you can recast the validation logic to be in terms of an `AsRef<str>`... then you could externalize it and share it across all the conversion functions for this and anything else that has to be a non-empty, less-than-32-character string.

colin-grapl (approved):

src/proto/graplinc/grapl/api/graph_query_service/v1beta1/graph_query_service.proto
wimax-grapl: Actually, yes! Our linter enforces this. It's because some languages don't do namespacing well iirc
Screenshot 2022-08-16 10 47 19 AM

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+impl std::fmt::Display for PropertyName {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.value)
+    }
+}
+
+impl PropertyName {
+    pub fn new_unchecked(value: String) -> Self {
```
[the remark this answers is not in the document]
wimax-grapl: Unfortunately this was carved out of a larger PR where lots of this is used

src/rust/rust-proto/src/graplinc/grapl/api/graph_query_service/v1beta1/server.rs
wimax-grapl: I agree that this builder doesn't follow the patterns we use elsewhere in the code and I'm tempted to rip it out for being different. @colin-grapl  ?

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
wimax-grapl: Changed

src/proto/graplinc/grapl/api/graph_query_service/v1beta1/graph_query_service.proto
colin-grapl: Yeah, there's no choice here - tonic helpfully removes the redundancy in generated code though.

src/rust/rust-proto/src/graplinc/grapl/api/graph_query_service/v1beta1/server.rs
colin-grapl: It's a builder that takes the required arguments in its constructor. If we add other arguments, such as "with_timeout", they would go here. If you want to rip it out you can go for it, we'll probably need it later but I'm fine just writing it again.

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
colin-grapl: The reason it's not a SerDeError is because it's not deserializing anything. The SerDeError that would be appropriate is the SerDeError::InvalidField, which requires the field name, which is not provided.

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
wimax-grapl: Let me add a comment. Discussed with Colin over DM he says there will be future validation here

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
wimax-grapl: this should def be serde error

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
colin-grapl: Do you mean does the error have to be &'static? If so, it either has to be static or a String, since errors are returned all the way up through main so they must always be `'static`.

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
colin-grapl: This will be fallible in the future when validation is added to ensure that PropertyName is constrained.

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
wimax-grapl: Great call, I made it a simple
```
impl TryFrom<&str> for EdgeName {
    type Error = SerDeError;
    fn try_from(raw: &str) -> Result<Self, Self::Error> {
        EdgeName::try_from(raw.to_owned())
    }
}
```

src/rust/rust-proto/src/graplinc/grapl/api/graph_query_service/v1beta1/server.rs
christophermaier: I'd rather remove it, since it's not really a "builder" at the moment. If there's cause to have a builder later, then it's easy to add. Better to not have dead / confusing / "maybe someday" code hanging around.

wimax-grapl: eh i'll leave it be

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
christophermaier: @colin-grapl By that logic, `impl TryFrom<&'static str> for PropertyName` shouldn't have `SerDeError` as its error type either.

My broader point is that there's really nothing materially different, on a business logic level, between converting from a `&str` and from a `String`, so I'd expect these implementations to look effectively identical. Having small differences like this just make things more confusing.

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
christophermaier: @colin-grapl I'm not talking about the error type here; I'm talking about the source type for the conversion (`TryFrom<&'static str> for PropertyName` vs. `TryFrom<String> for PropertyName`). I think you could just have this be `TryFrom<AsRef<str>> for PropertyName` (or whatever the syntax / type parameterization is). You could also externalize the validation logic to be reused across all these string-like things that have to be non-empty and less than 32 characters.

colin-grapl: Ah ok. In that case I would suggest two implementations, one for String, one for &str. It doesn't have to be static, you're correct.

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
colin-grapl: Yes, I think that they both should not return SerDeError. You can see in the case where it does that the field_name is just "PropertyName", which is not correct.

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
colin-grapl: The reason is because PropertyName requires an owned String, so taking AsRef<str> would require cloning even when the caller can provide a String, which is the common case.

src/rust/rust-proto/src/graplinc/grapl/api/graph_query_service/v1beta1/server.rs
wimax-grapl: oh missed chris ocmment. yeah i can delete it

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
colin-grapl: FWIW the validation is documented in the proto file - that's why many of these are infallible but will absolutely be fallible in the very near future.

src/rust/rust-proto/src/graplinc/grapl/common/v1beta1/types.rs
wimax-grapl: Colin, unfortunately you're shoving too much validation logic into something that should just be a ser/de layer

src/rust/rust-proto/src/graplinc/grapl/api/graph_query_service/v1beta1/server.rs
wimax-grapl: this ended up being very much the right move - the server impl was pretty different from the rest of our services

colin-grapl (approved):

PR #1906, opened 2022-08-18
Use nomad-native attribute for host IP

ghost:
⚠️ This will break anyone's local environment using Nomad prior to 1.3.2. ⚠️  
This switches us to using Nomad's native attribute to get the host IP since Nomad 1.3.2 has been out for over a month

Now that https://github.com/hashicorp/nomad/pull/12817 is merged in as
of Nomad 1.3.2, we should stop using the hax to get the host IP and
instead use Nomad's native method to get the host IP.

Tested locally with make up and checking Jaeger, regression tests may also pick up any issues but I'm not sure if Pulumi/Nomad will pick it up.


nomad/template_smple.json
```diff
@@ -0,0 +1,12 @@
+{
```
christophermaier, line 1: Was this supposed to have been committed?

christophermaier (approved):
Looks good, with one question.

colin-grapl (approved):

nomad/template_smple.json
twunderlich-grapl: Nope, removing now

PR #1917, opened 2022-08-25
Add mutation service

wimax-grapl:
### Which issue does this PR correspond to?
https://github.com/grapl-security/issue-tracker/issues/914

### What changes does this PR make to Grapl? Why?
Creates a service that will create nodes/edges in our scylla db

### How were these changes tested?
Just checked - looks like currently there are none. I'm gonna see if I can use the graph-query tests as a base for mutation tests.
I will not commit this without tests.

PR #1914, opened 2022-08-23
Slice uid allocator

colin-grapl:
### Which issue does this PR correspond to?
913

### What changes does this PR make to Grapl? Why?
Implements RPCs for creating and incrementing tenant uid namespaces.

### How were these changes tested?
There's integration tests in tests/

PR #1918, opened 2022-08-25
UID Allocator: wasn't running migrations, and wasn't running integration tests

wimax-grapl:
Followup to https://github.com/grapl-security/grapl/pull/1914

Found this out while doing the mutation service PR https://github.com/grapl-security/grapl/pull/1917 - it complained that the 'counters' relation didn't exist.


src/rust/uid-allocator/tests/uid_allocator.rs
```diff
@@ -1,31 +1,62 @@
...
+    // client:  "give me 100 ids for tenant ABC"
+    // server1: "ok i've reserved 1-10_000, here's 1-100"
+    // client:  "give me 100 ids for tenant ABC"
+    // server2: "ok i've reserved 10_001-20_000, here's 10_001-10_101"
+    // client:  "give me 100 ids for tenant ABC"
+    // server1: "I already have 1-10_000 reserved, so heres 101-200"
+
+    tracing::info!("allocating ids");
+    let num_requested_uids: u64 = 11_000;
+    // Just a sanity check that we're not allocating insane uids
```
colin-grapl, line 44: I'd just note that "if you're seeing this test fail, check that the number of instances is still 2 and the preallocation size is still 10k"

colin-grapl (approved):

src/rust/uid-allocator/src/allocator.rs
```diff
@@ -41,7 +40,7 @@ impl PreAllocation {
         // take a u32 because the requested size should never be close to that
         let mut requested_size = requested_size as u64;
 
-        // If we have exhausted the PreAllocation, return None.
+        // If we have not exhausted the PreAllocation
         if self.current < self.end {
```
colin-grapl, line 44: I feel like this probably should be <= but idk

colin-grapl (approved):

PR #1922, opened 2022-08-26
Migrate tracing to Lightstep via otel-collector

ghost:

This migrates tracing to Lightstep from Jaeger.
Why:
- Lightstep is hosted outside of our environment so that if we run into cluster problems we still have telemetry
- Lightstep can store and correlate traces, metrics and logs in one location
- Lightstep has a developer mode for local development

We are using otel-collector to collect the traces in both Jaeger and Zipkin format, translate them into OTEL format and forward them to Lightstep.
Why:

- Vector's tracing support is in alpha and likely to change
- Vector doesn't support OTEL format (not even for metrics)
- otel-collector is well-supported.
- We could in theory switch rust and python to use OTEL format and send the data directly to Lightstep. However Consul doesn't yet support OTEL (although its getting close with recent changes being merged into Envoy)

make up after setting the api key in Lightstep and then logging in


pulumi/infra/observability_env_vars.py
```diff
@@ -25,3 +27,47 @@ def observability_env_vars_for_local() -> str:
         OTEL_EXPORTER_ZIPKIN_ENDPOINT   = {zipkin_endpoint}
         OTEL_BSP_MAX_EXPORT_BATCH_SIZE = {otel_bsp_max_export_batch_size}
     """
+
+
+# lightstep_token should be pulumi.Output[str], but the additional type causes pulumi.Output.all to blow up during
+# typechecking
+def otel_config(lightstep_token: pulumi.Output) -> pulumi.Output[str]:
+    return pulumi.Output.all(lightstep_token=lightstep_token).apply(
+        lambda args: f"""
```
wimax-grapl, line 36: very funky but i'll take your word for it

wimax-grapl (approved):

PR #1913, opened 2022-08-22
Slice provision

colin-grapl:

PR #1924, opened 2022-08-29
Scylla provisioner (with 'provision_graph_for_tenant')

wimax-grapl:
Commandeering https://github.com/grapl-security/grapl/pull/1913

Somehow we didn't have an issue for this.

This service provisions Scylla for a given tenant. It replaces the hardcoded stuff in the graph_query_service integration tests.

Integrated the provisioner into the graph_query_service integration tests; also has its own (questionably worthwhile) integration tests.


pulumi/grapl/Pulumi.testing.yaml
```diff
@@ -4,15 +4,6 @@ config:
   grapl:cloudsmith-repository-name: grapl/testing
   grapl:confluent-environment-name: testing
   grapl:env_vars:
```
wimax-grapl, line 6: these have been useless for quite some time; though i'll keep the mechanism in there i guess

colin-grapl (approved):

wimax-grapl:
colin please stop approving ur own code :) (it makes it look like it's already approved and I don't need an approve from anyone)

colin-grapl (approved):

pulumi/grapl/Pulumi.testing.yaml
colin-grapl, line 6: Should we just remove `grapl:env_vars:` entirely?

src/rust/scylla-provisioner/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+rust-proto = { path = "../rust-proto" }
+scylla = "0.4.7"
+secrecy = "0.8.0"
+thiserror = "1.0.31"
+tokio = { version = "1.20.0", features = ["full"] }
+tracing = "0.1.36"
+uuid = { version = "1.1.2", features = ["v4"] }
+
+[dev-dependencies]
+test-log = { version = "0.2", default-features = false, features = ["trace"] }
```
inickles-grapl, line 21: Does this print terminal escape characters?

I've seen them in Rust integration test stdout before, and remember thinking it might have been due to this.

If it is, I suggest we configure it in such a way that doesn't print those, or use something else if that can't be avoided.

Maybe what I saw was the how the tracing subscriber was configured instead? Not sure, but can be addressed later.

src/rust/scylla-provisioner/src/server.rs
```diff
@@ -0,0 +1,156 @@
...
+pub enum ScyllaProvisionerError {
+    #[error("Scylla Error {0}")]
+    ScyllaError(#[from] QueryError),
+}
+
+impl From<ScyllaProvisionerError> for Status {
+    fn from(error: ScyllaProvisionerError) -> Self {
+        match error {
+            ScyllaProvisionerError::ScyllaError(error) => {
+                Status::unknown(format!("Scylla Error: {}", error).as_str())
```
inickles-grapl, line 44: I'd just call `error.to_string()` instead of using the `format` marco, because it looks like the result of that macro would be "Scylla Error: Scylla Error ...", where those dots are whatever the Display of QueryError would be. Formatting the same thing twice here.

Either that, or just make the error variant #[error(transparent)] and remove the format string from it there. It's not really adding any info there anyway.

inickles-grapl (approved):
Just some comments.

pulumi/grapl/__main__.py
```diff
@@ -66,6 +66,7 @@ def _container_images(artifacts: ArtifactGetter) -> Mapping[str, DockerImageId]:
 
     return {
         "analyzer-execution-sidecar": DockerImageId("TODO implement analzyer executor"),
+        "scylla-provisioner": builder.build_with_tag("scylla-provisioner"),
```
twunderlich-grapl, line 69: Nit: Do we want to keep this in alphabetical order here?

ghost (approved):

src/rust/scylla-provisioner/Cargo.toml
wimax-grapl: I'm not sure what's responsible for it tbh

src/rust/scylla-provisioner/src/server.rs
wimax-grapl: In general we seem to follow a pattern of making our errors over-the-wire as, well, vague and devoid of detail as possible; including the details of the real error is kind of an aberration here. I'll change it for now, but something to consider.

PR #1929, opened 2022-08-30
Move Graph DB related services to their own Nomad file

wimax-grapl:
None

Split apart our `grapl-core.nomad` by a reasonable boundary (stuff related to querying and mutating our graph databases).

integration tests :')

### DIFF
--- a/pulumi/grapl/__main__.py
+++ b/pulumi/grapl/__main__.py
@@ -462,14 +462,6 @@ def main() -> None:
         )
         pulumi.export("stage-url", api_gateway.stage.invoke_url)
 
-    grapl_core_vars: Final[NomadVars] = dict(
-        event_source_db=event_source_db.to_nomad_service_db_args(),
-        organization_management_db=organization_management_db.to_nomad_service_db_args(),
-        plugin_registry_db=plugin_registry_db.to_nomad_service_db_args(),
-        plugin_work_queue_db=plugin_work_queue_db.to_nomad_service_db_args(),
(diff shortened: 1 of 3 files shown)


pulumi/grapl/__main__.py
```diff
@@ -479,6 +502,9 @@ def main() -> None:
         vars=grapl_core_vars,
         opts=pulumi.ResourceOptions(
             provider=nomad_provider,
+            depends_on=[
+                nomad_graph_db.job,
+            ],
```
wimax-grapl, line 507: we _could_ try deploying them at the same time; i actually thought this was a decent way to get some "ordering" here (you wouldn't want our other services starting up before our graph dbs were ready, right?)

pulumi/grapl/__main__.py
```diff
@@ -479,6 +502,9 @@ def main() -> None:
         vars=grapl_core_vars,
         opts=pulumi.ResourceOptions(
             provider=nomad_provider,
+            depends_on=[
```
twunderlich-grapl, line 505: Does the order of whether we deploy grapl-core vs the graph-db services actually matter here or is this just to make sure that we have a consistent deployment order?

twunderlich-grapl (approved):

pulumi/grapl/__main__.py
wimax-grapl: Doesn't really matter. And could potentially slowdown the overall deploy. So I'm fine ripping that out

PR #1935, opened 2022-08-30
Add new web API for Plugin Registry.

inickles-grapl:

This introduces the Plugin Registry API to Grapl web, exposing `create` and `get_metadata`.

I made an integration test that exercises `create` and then runs `get_metadata` and validates the returned plugin name is the same that we specified when creating.


src/python/provisioner/provisioner/app.py
```diff
@@ -128,6 +128,6 @@ def provision() -> None:
             username=username,
             cleartext=password,
             role="owner",
-            org="test_org",
+            org="00000000-0000-0000-0000-000000000000",
```
wimax-grapl, line 131: is an org_id a tenant_id? if so, can we deduplicate these two names?

wimax-grapl: potentially as a followup PR

src/rust/grapl-web-ui/src/lib.rs
```diff
@@ -27,6 +28,8 @@ pub fn run(config: config::Config) -> Result<Server, std::io::Error> {
             jsonwebtoken_google::Parser::new(&config.google_client_id),
         ));
         let graphql_endpoint = Data::new(config.graphql_endpoint.clone());
+        let plugin_registry_client = Data::new(Mutex::new(config.plugin_registry_client.clone()));
```
wimax-grapl, line 31: fwiw i believe our tonic Clients are cheap to clone already, no lock needed?

src/rust/grapl-web-ui/src/routes/api/plugin/create.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        return Ok(body.into());
+    }
+
+    Err(PluginError::BadRequest {
+        message: "missing form data for plugin/create plugin artifact".to_string(),
+    })
+}
+
+// It'd be greate if we didn't have to do this and PluginType jsut derived serde::Deserialize itself
```
wimax-grapl, line 162: thats a solid point; we should bring it up w jesse. all our enums are not currently SerDe

src/rust/grapl-web-ui/src/routes/api/plugin/get_metadata.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct GetPluginMetadataResponse {
+    pub plugin_id: Uuid,
+    pub display_name: String,
+    pub plugin_type: String,
+    pub event_source_id: Option<uuid::Uuid>,
+}
+
+#[tracing::instrument(skip(plugin_registry_client, user, data), fields(
+    username = tracing::field::Empty
+))]
```
wimax-grapl, line 31: does username ever get assigned?

src/rust/grapl-web-ui/tests/api/plugin/create.rs
```diff
@@ -0,0 +1,76 @@
...
+async fn plugin_create() -> eyre::Result<()> {
+    let app = crate::test_app::TestApp::init().await?;
+
+    app.login_with_test_user().await?;
+
+    let plugin_name = "Test Plugin";
+    let create_metadata_body = serde_json::json!({
+            "plugin_name": plugin_name,
+            "plugin_type": "generator",
+            "event_source_id": "00000000-0000-0000-0000-000000000000"
```
wimax-grapl, line 16: add a TODO?

wimax-grapl: i mean, you could make a real event source with the event source client

src/rust/grapl-web-ui/tests/api/plugin/create.rs
```diff
@@ -0,0 +1,76 @@
...
+    });
+    let form = reqwest::multipart::Form::new()
+        .part(
+            "metadata",
+            reqwest::multipart::Part::text(create_metadata_body.to_string()),
+        )
+        .part(
+            "plugin_artifact",
+            reqwest::multipart::Part::bytes("junk".as_bytes()),
+        );
```
wimax-grapl, line 26: if you really wanted to exercise it, you _could_ do the `get_sysmon_generator()` - we may want to make a test_fixtures crate to share this everywhere

wimax-grapl: that fn exists already btw

wimax-grapl (commented):
overall looks good

src/python/provisioner/provisioner/app.py
inickles-grapl: > is an org_id a tenant_id? if so, can we deduplicate these two names?

Yeah, as far as I know, they are the same.

src/rust/grapl-web-ui/src/lib.rs
inickles-grapl: PluginRegistryServiceClient::create_plugin takes a mutable self, which I can't get behind the web::Data without a Mutex. https://docs.rs/actix-web/latest/actix_web/web/struct.Data.html.

src/rust/grapl-web-ui/src/routes/api/plugin/get_metadata.rs
inickles-grapl: as good catch

src/rust/grapl-web-ui/tests/api/plugin/create.rs
inickles-grapl: Sounds to me more like something for an e2e test. wdyt?

src/rust/grapl-web-ui/tests/api/plugin/create.rs
inickles-grapl: hmm yeah. I agree we should probably do that, though I'm thinking in doing so I should expose the `deploy` and `tear_down` APIs as well. I think I'll do that.

src/rust/grapl-web-ui/src/lib.rs
wimax-grapl: ah got it

src/rust/grapl-web-ui/tests/api/plugin/create.rs
wimax-grapl: that line keeps blurring :) i worry about the hardcoded event source id since, well, couldn't somebody else start using that and corrupting this test?

src/rust/grapl-web-ui/src/lib.rs
inickles-grapl: Yes-ish, I'd need `plugin_registry_client.get_ref().clone()`, but regardless I think you're right that I should just do that instead. Thanks!

src/rust/grapl-web-ui/tests/api/plugin/create.rs
inickles-grapl: Kk, I'm not just calling into `e2e-tests::test_fixtures::get_sysmon_generator` for now.

src/rust/grapl-web-ui/tests/api/plugin/create.rs
inickles-grapl: Kk, I'm generating a random Uuid instead.

I don't want to use the event source client, that sounds like overkill for this integration test.

src/rust/grapl-web-ui/src/authn/error.rs
```diff
@@ -27,6 +27,8 @@ pub enum AuthenticationError {
     SessionTokenNotFound,
     #[error("session expired")]
     SessionExpired,
+    #[error("error parsing tenant_id '{input}': {source}")]
```
jgrillo-grapl, line 30: this error text refers to a `tenant_id` but the enum variant calls it an `org_id`.. which is it?

src/rust/grapl-web-ui/src/routes/api/plugin.rs
```diff
@@ -0,0 +1,61 @@
...
+    let get_plugin_request = GetPluginRequest::new(plugin_id, authenticated_tenant_id);
+
+    tracing::debug!(message = "getting plugin metadata", ?get_plugin_request);
+
+    let get_plugin_response = client.get_plugin(get_plugin_request).await?;
+    let plugin_metadata = get_plugin_response.plugin_metadata().to_owned();
+
+    if authenticated_tenant_id != plugin_metadata.tenant_id() {
+        tracing::warn!(
+            message = "user requested to deploy plugin owned by another tenant/organization",
```
jgrillo-grapl, line 53: same question regarding `tenant` and `organization`.. are they the same thing?

src/rust/grapl-web-ui/src/routes/api/plugin/get_deployment.rs
```diff
@@ -0,0 +1,96 @@
...
+        deployed: response.deployed(),
+    };
+
+    Ok(HttpResponse::Ok().json(web_response))
+}
+
+//TODO: It'd be great if rust_proto::graplinc::grapl::api::plugin_registry::v1beta1::PluginDeploymentStatus
+// implemented Display, or serde::Serialize
+fn serialize_deployment_status<S>(
+    health_status: &PluginDeploymentStatus,
```
wimax-grapl, line 60: i tihnk PluginType did this in rust-proto, the mapping to a string - might make sense to do that in rust-proto for this as well?

src/rust/grapl-web-ui/tests/api/plugin.rs
```diff
@@ -0,0 +1,243 @@
...
+        &response
+    );
+
+    let response_body = response.json::<GetPluginHealthResponse>().await?;
+
+    println!("response body: {:?}", response_body);
+
+    Ok(response_body)
+}
+
```
wimax-grapl, line 228: `// TODO uncomment once analyzers are done`

src/rust/grapl-web-ui/src/routes/api/plugin/get_deployment.rs
```diff
@@ -0,0 +1,96 @@
...
+        plugin_id: response.plugin_id(),
+        timestamp: response.timestamp(),
+        status: response.status(),
+        deployed: response.deployed(),
+    };
+
+    Ok(HttpResponse::Ok().json(web_response))
+}
+
+//TODO: It'd be great if rust_proto::graplinc::grapl::api::plugin_registry::v1beta1::PluginDeploymentStatus
```
jgrillo-grapl, line 57: This idea scares me a little bit because all the stuff in `rust_proto::graplinc::grapl::api` is "internal" and not deliberately exposed publicly. In this particular case it's just some innocuous enum, but that could change later and the person changing it might not know this thing is actually public. 

I think making the public API code very deliberate about what it does expose, and ensuring by default it _does not expose_ internal data is much safer.

jgrillo-grapl (approved):
I don't think I understand everything here, but I'm happy to see tests! I left a question about some terminology I am not yet familiar with.

src/rust/grapl-web-ui/src/routes/api/plugin/deploy.rs
```diff
@@ -0,0 +1,38 @@
...
+    DeployPluginRequest,
+    PluginRegistryServiceClient,
+};
+
+use super::PluginError;
+
+#[derive(serde::Deserialize)]
+pub(super) struct DeployPluginParameters {
+    plugin_id: uuid::Uuid,
+}
```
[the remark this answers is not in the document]
jgrillo-grapl: This sounds very dangerous

src/rust/grapl-web-ui/src/routes/api/plugin.rs
```diff
@@ -1,11 +1,61 @@
...
+    cfg.route("/get_health", web::get().to(get_health::get_health));
     cfg.route("/get_metadata", web::get().to(get_metadata::get_metadata));
+    cfg.route("/tear_down", web::post().to(tear_down::tear_down));
+}
+
+/// Validate the authenticated user has permissions to operate on the supplied plugin ID.
+///
+/// Returns validated Plugin metadata.
+#[tracing::instrument(skip(client))]
+async fn verify_plugin_ownership<'a>(
```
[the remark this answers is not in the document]
inickles-grapl: If you mean the upstream Plugin Registry should do the enforcement, I 100% agree. Doing so there is more important than here. Though, I also think that even with that, this web server should enforce as well, call it belt and suspenders.

src/rust/grapl-web-ui/tests/api/plugin.rs
inickles-grapl: Yeah, I have the TODO on line 25, which calls into this. I thought that was sufficient for TODO. This needed to be commented as well to satisfy dead_code lint, but doesn't seem worth putting a TODO in each place. I'm fine with adding here as well if you prefer, just lemme know.

src/rust/grapl-web-ui/src/routes/api/plugin.rs
inickles-grapl: Per https://grapl-internal.slack.com/archives/C0174A8QV2S/p1661961943096359 they are. New references to 'organization id' have been replaced with 'tenant id'

jgrillo-grapl (approved):

PR #1938, opened 2022-08-31
[backburner] Teardown plugins after tests if the test otherwise succeeded. Also eyre-ify some stuff

wimax-grapl:
One tradeoff worth mentioning here is that we won't have logs for "successful" plugins based on how nomad_artifacts.py currently works! Just failed plugins. Seems fine for now.

### DIFF
--- a/src/rust/plugin-registry/tests/test_deploy_plugin.rs
+++ b/src/rust/plugin-registry/tests/test_deploy_plugin.rs
@@ -266,16 +266,11 @@ async fn test_teardown_plugin() -> eyre::Result<()> {
         .plugin_deployment();
 
     eyre::ensure!(plugin_deployment.plugin_id() == plugin_id, "plugins equal");
-    eyre::ensure!(plugin_deployment.deployed(), "plugin deployed");
+    eyre::ensure!(!plugin_deployment.deployed(), "plugin not deployed");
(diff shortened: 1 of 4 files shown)


src/rust/plugin-registry/tests/test_deploy_plugin.rs
```diff
@@ -246,7 +265,17 @@ async fn test_teardown_plugin() {
...
 
-    assert_eq!(plugin_deployment.plugin_id(), plugin_id);
-    assert!(!plugin_deployment.deployed());
-    assert_eq!(plugin_deployment.status(), PluginDeploymentStatus::Success);
+    eyre::ensure!(plugin_deployment.plugin_id() == plugin_id, "plugins equal");
+    eyre::ensure!(plugin_deployment.deployed(), "plugin deployed");
+    eyre::ensure!(
+        plugin_deployment.status() == PluginDeploymentStatus::Success,
+        "deployment successful"
+    );
```
wimax-grapl, line 273: the above eyre-ification came about because of the change to `-> eyre::Result` on line 211. I was like oh i guess i should remove these asserts

wimax-grapl:
Draft-ified because Jesse asked to hold off until we have a better logging story

PR #1948, opened 2022-09-02
Otel-collector deploy 2-step part 1

ghost:
Apparently the Nomad pulumi provider can't handle replacing a task
here for some reason. As such I'm commenting out the code so it gets
destroyed first and then in a follow-up PR will uncomment out the code.

Yes, I hate this.

Specifically we get the following error:
panic: fatal: An assertion has failed: Expected diff to not require
deletion or replacement during Update of
urn:pulumi:testing::grapl::grapl:NomadJob$nomad:index/job:Job::testing-grapl-observability-job

https://github.com/grapl-security/grapl/pull/1949 will be merged in afterward as part 2

### DIFF
--- a/pulumi/grapl/__main__.py
+++ b/pulumi/grapl/__main__.py
@@ -315,17 +315,17 @@ def main() -> None:
         pulumi_config.get(key="lightstep-access-token") or ""
     )
     otel_configuration = otel_config(lightstep_access_token)
-    NomadJob(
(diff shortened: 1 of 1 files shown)


christophermaier (approved):
🤔 That's unfortunate... might be worth digging into a bit / reporting it upstream.

jgrillo-grapl (approved):

PR #1952, opened 2022-09-02
Reorganize python-proto to more closely resemble rust-proto

wimax-grapl:
Some existing things about python-proto I didn't like:

- we really shouldn't do anything complex in `__init__.py`s - I'd consider it a code smell, it should be used like a mod.rs (importing) - so I ripped all the SerDe stuff out to `serde.py`. 
- `api.py` had taken over the `api` namespace; I moved its contents to `api/graph/v1beta1/messages.py`
- I changed the "import every proto type as _Thing" pattern to "proto.Thing" - I find this clearer, and a better use of namespacing anyway.

### DIFF
--- a/src/python/python-proto/python_proto/api/graph/v1beta1/messages.py
+++ b/src/python/python-proto/python_proto/api/graph/v1beta1/messages.py
@@ -3,69 +3,27 @@ from __future__ import annotations
 import dataclasses
 from typing import Mapping, Sequence, cast
 
-from graplinc.grapl.api.graph.v1beta1.types_pb2 import (
-    DecrementOnlyIntProp as _DecrementOnlyIntProp,
-)
-from graplinc.grapl.api.graph.v1beta1.types_pb2 import (
-    DecrementOnlyUintProp as _DecrementOnlyUintProp,
-)
-from graplinc.grapl.api.graph.v1beta1.types_pb2 import Edge as _Edge
-from graplinc.grapl.api.graph.v1beta1.types_pb2 import EdgeList as _EdgeList
-from graplinc.grapl.api.graph.v1beta1.types_pb2 import ExecutionHit as _ExecutionHit
-from graplinc.grapl.api.graph.v1beta1.types_pb2 import (
-    GraphDescription as _GraphDescription,
-)
-from graplinc.grapl.api.graph.v1beta1.types_pb2 import (
-    IdentifiedGraph as _IdentifiedGraph,
-)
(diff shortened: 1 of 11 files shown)


src/python/python-proto/python_proto/api/graph/v1beta1/messages.py
```diff
@@ -803,8 +769,8 @@ def from_proto(proto_execution_hit: _ExecutionHit) -> ExecutionHit:
             risky_node_keys=proto_execution_hit.risky_node_keys,
         )
 
-    def into_proto(self) -> _ExecutionHit:
-        proto_execution_hit = _ExecutionHit()
+    def into_proto(self) -> proto.ExecutionHit:
+        proto_execution_hit = proto.ExecutionHit()
```
wimax-grapl, line 773: don't forget this file, which is "too big" by default for github PR

src/python/python-proto/python_proto/serde.py
```diff
@@ -0,0 +1,83 @@
...
+    @staticmethod
+    @abstractstaticmethod
+    def deserialize(bytes_: bytes) -> SerDe[P]:
+        raise NotImplementedError
+
+    def serialize(self) -> bytes:
+        return self.into_proto().SerializeToString()
+
+    @staticmethod
+    @abstractstaticmethod
```
colin-grapl, line 37: abstractstaticmethod is deprecated, youc an just use abstractmethod

src/python/python-proto/python_proto/serde.py
```diff
@@ -0,0 +1,83 @@
...
+            hasattr(subclass, "proto_cls")
+            and hasattr(subclass, "deserialize")
+            and callable(subclass.deserialize)
+            and hasattr(subclass, "serialize")
+            and callable(subclass.serialize)
+            and hasattr(subclass, "from_proto")
+            and callable(subclass.from_proto)
+            and hasattr(subclass, "into_proto")
+            and callable(subclass.into_proto)
+        )
```
colin-grapl, line 26: We should probalby just use pydantic if we want to do this sort of thing, but this is fine for now

colin-grapl (approved):

jgrillo-grapl (approved):

src/python/python-proto/python_proto/serde.py
wimax-grapl: this is all pre-existing and i dont want to change it :')

PR #1954, opened 2022-09-03
add a list_plugins RPC method

jgrillo-grapl:
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/1003

This PR implements a `list_plugins` RPC endpoint in plugin-registry which returns all the plugins of a given `PluginType` for a given `tenant_id`. 

Automated tests


src/rust/plugin-registry/tests/test_list_plugins.rs
```diff
@@ -0,0 +1,114 @@
...
+    let client_config = PluginRegistryClientConfig::parse();
+    let mut client = build_grpc_client(client_config).await?;
+
+    let tenant_id = uuid::Uuid::new_v4();
+    let event_source_1_id = uuid::Uuid::new_v4();
+    let event_source_2_id = uuid::Uuid::new_v4();
+
+    let generator_1_create_response = {
+        let display_name = "generator 1";
+        let artifact = get_example_generator()?;
```
wimax-grapl, line 33: tbh we can just use a regular `let artifact = Bytes::from("arbitrary binary");` 

instead of the concrete, real Example Generator here, since we're not really testing the contents of the generator binary, just how they're managed.

wimax-grapl: (it is fine as-is, but I like to use intentionally-marked arbitrary data whenever possible in a test so I know what's important and what's not)

src/rust/plugin-registry/tests/test_list_plugins.rs
```diff
@@ -0,0 +1,114 @@
...
+            )
+            .timeout(std::time::Duration::from_secs(5))
+            .await??
+    };
+
+    let generator_2_id = generator_2_create_response.plugin_id();
+
+    let analyzer_create_response = {
+        let display_name = "analyzer";
+        let artifact = get_example_generator()?;
```
wimax-grapl, line 75: (This analyzer/generator dichotomy here is what first got me thinking about "arbitrary binary")

src/rust/plugin-registry/tests/test_list_plugins.rs
```diff
@@ -0,0 +1,114 @@
...
+        .await??
+        .plugins();
+
+    assert_eq!(generator_plugins.len(), 2);
+    assert_eq!(generator_plugins[0].plugin_id(), generator_1_id);
+    assert_eq!(generator_plugins[1].plugin_id(), generator_2_id);
+
+    let analyzer_plugins = client
+        .list_plugins(ListPluginsRequest::new(tenant_id, PluginType::Analyzer))
+        .timeout(std::time::Duration::from_secs(5))
```
wimax-grapl, line 106: fwiw - i know they're peppered throughout the other tests in here, but with the introduction of `client-executor` I don't think we need these `.timeout`s (and the extra `?`) anymore - but nbd if you keep it in

src/rust/plugin-registry/tests/test_list_plugins.rs
```diff
@@ -0,0 +1,114 @@
...
+        PluginType,
+    },
+};
+
+fn get_example_generator() -> Result<Bytes, std::io::Error> {
+    std::fs::read("/test-fixtures/example-generator").map(Bytes::from)
+}
+
+#[test_log::test(tokio::test)]
+async fn test_list_plugins() -> eyre::Result<()> {
```
wimax-grapl, line 23: Can you add an edge case test for "a tenant id with no plugins"? should be just "build a tenant, add no generators/analyzers, call list_plugins"

wimax-grapl (approved):

src/rust/plugin-registry/tests/test_list_plugins.rs
jgrillo-grapl: done

src/rust/plugin-registry/tests/test_list_plugins.rs
jgrillo-grapl: done

wimax-grapl (approved):

PR #1956, opened 2022-09-06
Lightstep dev mode

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/1010

This adds the ability to run Lightstep in dev mode and get data into Lightstep

Run the Lightstep docker container using the command from Lightstep
`make up`, 
Check that data makes it into the dev panel afterward

### DIFF
--- a/pulumi/infra/observability_env_vars.py
+++ b/pulumi/infra/observability_env_vars.py
@@ -32,17 +32,14 @@ def observability_env_vars_for_local() -> str:
 # lightstep_token should be pulumi.Output[str], but the additional type causes pulumi.Output.all to blow up during
 # typechecking
 def otel_config(
-    lightstep_token: pulumi.Output, lightstep_endpoint: str = "ingest.lightstep.com:443"
+    lightstep_token: pulumi.Output,
+    lightstep_endpoint: str = "ingest.lightstep.com:443",
+    lightstep_is_endpoint_secure: str = "true",
 ) -> pulumi.Output[str]:
-    # use the endpoint as a way to figure out if we're in local dev or not
-    if lightstep_endpoint == "ingest.lightstep.com:443":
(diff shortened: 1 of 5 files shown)


pulumi/infra/observability_env_vars.py
```diff
@@ -31,8 +31,19 @@ def observability_env_vars_for_local() -> str:
...
-def otel_config(lightstep_token: pulumi.Output) -> pulumi.Output[str]:
-    return pulumi.Output.all(lightstep_token=lightstep_token).apply(
+def otel_config(
+    lightstep_token: pulumi.Output, lightstep_endpoint: str = "ingest.lightstep.com:443"
+) -> pulumi.Output[str]:
+    # use the endpoint as a way to figure out if we're in local dev or not
+    if lightstep_endpoint == "ingest.lightstep.com:443":
+        is_endpoint_insecure = "false"
+    else:
+        is_endpoint_insecure = "true"
```
wimax-grapl, line 41: I guess I'm a bit confused here - do we actually care about local or not, or just if it's 443 (secure) or 80 (insecure)?

wimax-grapl (approved):

pulumi/infra/observability_env_vars.py
twunderlich-grapl: Just whether it's secure or not. I think I may switch this to just be passed in as a config instead.

wimax-grapl (approved):

pulumi/infra/observability_env_vars.py
twunderlich-grapl: @christophermaier, 
For devs, yes that should be specific to each engineer.
For the others, not yet. Right now everything should be staging. It looks like the right way to handle this is to have separate api keys, but I still need to finish testing that.

PR #1957, opened 2022-09-06
Upgrade from Mypy 0.910 to 0.960

wimax-grapl:
I ahve some code from colin that uses pattern matching and walrus := syntax - requiring a mypy upgrade. I tested that the previous fails vs succeeds with https://mypy-play.net/?mypy=0.960&python=3.10

(also, adding some extra type stubs)

### DIFF
--- a/3rdparty/python/mypy.lockfile
+++ b/3rdparty/python/mypy.lockfile
@@ -9,7 +9,7 @@
 //     "CPython==3.10.*"
 //   ],
 //   "generated_with_requirements": [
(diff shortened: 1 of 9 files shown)


ghost (approved):

christophermaier (approved):

colin-grapl (approved):

wimax-grapl:
Had to introduce a new hack due to https://github.com/python/mypy/issues/5144

colin-grapl (approved):

PR #1959, opened 2022-09-07
Add missing egress sg rule to postgres

ghost:

This fixes db timeouts because while connections can reach postgres the response can never get through

### DIFF
--- a/pulumi/infra/postgres.py
+++ b/pulumi/infra/postgres.py
@@ -84,6 +84,7 @@ class Postgres(pulumi.ComponentResource):
         # Allow communication between nomad-agents and RDS
         aws.ec2.SecurityGroupRule(
             f"{name}-nomad-agents-egress-to-rds",
+            description=f"Connect to {name} RDS",
             type="egress",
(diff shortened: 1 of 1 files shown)


pulumi/infra/postgres.py
```diff
@@ -104,6 +104,17 @@ def __init__(
             opts=pulumi.ResourceOptions(parent=self.security_group),
         )
 
+        aws.ec2.SecurityGroupRule(
+            f"{name}-rds-egress-to-nomad-agents",
+            type="egress",
+            security_group_id=self.security_group.id,
+            from_port=postgres_port,
+            to_port=postgres_port,
```
christophermaier, line 112: Is this correct? I thought clients' ports would be arbitrary, high number ones.

twunderlich-grapl: Yep you're correct. Adding just egress doesn't seem to work

twunderlich-grapl:
After manually testing I've confirmed that this isn't necessary. Closing this

PR #1963, opened 2022-09-08
Internal 1009 analyzer dispatcher

jgrillo-grapl:
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/1009

This PR adds analyzer-dispatcher and analyzer-dispatcher-retry services.

Automated tests


src/rust/analyzer-dispatcher/tests/analyzer_dispatcher_integration_tests.rs
```diff
@@ -0,0 +1,115 @@
...
+        )
+        .await?;
+
+    ctx.deploy_generator(generator_id).await?;
+
+    let analyzer_id = ctx
+        .create_analyzer(
+            tenant_id,
+            "analyzer dispatcher test analyzer".to_string(),
+            Bytes::from("fake"), // FIXME: use a real analyzer
```
wimax-grapl, line 44: For generator/analyzers, since we're not deploying them I do think it's totally fine to just have Bytes::from("fake") instead of the real deal - but whatevs

src/rust/async-cache/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+        let mut async_cache =
+            AsyncCache::new(10, Duration::from_millis(100), 10, 10, |_| async { 42 }).await;
+
+        let foo = async_cache.get(3).await.expect("should be successful");
+        assert_eq!(foo, None); // empty but update has been enqueued
+
+        tokio::time::sleep(Duration::from_millis(1)).await; // "short delay"
+
+        let foo = async_cache.get(3).await.expect("should be successful");
+        assert_eq!(foo, Some(42)); // cache updated and value is present
```
wimax-grapl, line 176: I know this is pre-existing code you just moved to its own lib - great, btw - but, is there any guarantee that the `updater_rx` will have received and followed through with updating for key 3 after 1ms, or is it a potentially-race-condition?

src/rust/e2e-tests/src/test_utils/context.rs
```diff
@@ -93,6 +93,96 @@ pub struct SetupGeneratorOptions {
 }
 
 impl E2eTestContext {
+    pub async fn create_tenant(&mut self) -> eyre::Result<Uuid> {
+        tracing::info!("creating tenant");
+        // TODO: actually create a real tenant
+        Ok(Uuid::new_v4())
+    }
```
wimax-grapl, line 100: Liking this whole refactor, thanks!

src/rust/e2e-tests/src/test_utils/context.rs
```diff
@@ -109,45 +199,39 @@ impl E2eTestContext {
...
-                    .deploy_plugin(DeployPluginRequest::new(plugin.plugin_id()))
-                    .await?;
+                self.deploy_generator(generator_id).await?;
             }
-            plugin
+
+            generator_id
         };
 
         let setup_result = SetupResult {
```
wimax-grapl, line 231: this is my naming mistake, not yours, but potentially rename to SetupGeneratorResult

src/rust/generator-dispatcher/src/lib.rs
```diff
@@ -82,13 +100,81 @@ impl GeneratorDispatcher {
...
+                                    Duration::from_millis(10000)
+                                };
+
+                                tracing::error!(
+                                    message = "error retrieving generator IDs from plugin-registry",
+                                    error =% e,
+                                    retry_delay =? backoff,
+                                );
+
+                                tokio::time::sleep(backoff).await;
```
wimax-grapl, line 159: A) `get_generators_for_event_source` already _does_ retries itself (See `macro_rules! execute_client_rpc`), so I'm a bit confused as to why we're adding another layer on top.
B) At the very least, I'd prefer to do this in a separate bare function than nested so deep here, if possible?

src/rust/plugin-work-queue/src/test_utils.rs
```diff
@@ -64,6 +69,31 @@ impl PsqlQueueTestExtensions for PsqlQueue {
...
+            plugin_id
+        )
+        .fetch_all(&self.pool)
+        .await?;
+
+        Ok(analyzer_messages)
+    }
 }
 
 pub async fn scan_for_plugin_message_in_pwq(
```
wimax-grapl, line 99: another "oh fuck, i didn't name something forward-thinking-ly": could we rename this to `scan_generator_messages`?

wimax-grapl (commented):
This will be an approve once we hash out the retried-rpc stuff

src/rust/async-cache/src/lib.rs
jgrillo-grapl: It's subject to the async runtime, so there's no guarantee. In practice 1ms appears to be long enough.

src/rust/generator-dispatcher/src/lib.rs
jgrillo-grapl: I tried following that code through and couldn't figure it out. Where does the retry happen and how can I configure it?

src/rust/e2e-tests/src/test_utils/context.rs
jgrillo-grapl: I tried to avoid having these changes impact a bunch of other services

wimax-grapl (approved):

PR #1966, opened 2022-09-08
Comment the living hell out of execute_client_rpc

wimax-grapl:
none

make this macro easier to grok; made the backoff options configurable per-RPC instead of global

compiling them :)

### DIFF
--- a/src/rust/rust-proto/src/client_macros.rs
+++ b/src/rust/rust-proto/src/client_macros.rs
@@ -1,14 +1,36 @@
 /// Return true if this RPC should be retried.
 type RetryPredicate = fn(&tonic::Status) -> bool;
-pub struct ExecuteClientRpcOptions {
+pub struct RpcConfig {
     pub retry_predicate: RetryPredicate,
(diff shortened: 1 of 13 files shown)


src/rust/rust-proto/src/client_macros.rs
```diff
@@ -29,34 +51,48 @@ macro_rules! execute_client_rpc {
...
-            let proto_response = $self
+            // Convert the Native request - something in `rust-proto` - into a Protobuf request.
+            let proto_request: ProtoRequestType = ProtoRequestType::try_from($native_request)?;
+
+            // Execute the RPC, with retries.
+            let tonic_response: tonic::Response<_> = $self
                 .executor
                 .spawn_conditional(
-                    backoff.take(num_retries),
+                    backoff.take(backoff_opts.num_retries),
```
jgrillo-grapl, line 83: in order for this to work in `async-cache` we need to be able to chart a path through this macro which _avoids_ specifying a retry count

jgrillo-grapl (approved):

PR #1976, opened 2022-09-13
Enable dnsmasq as a system job that runs everywhere

ghost:

For AWS we need a way to change Nomad jobs so that they get dns from localhost:8600 (the Consul agent). Ideally we'd be able to do dns_server = localhost:8600, but that isn't supported. Looking at the docker docs, they kinda support it using published ports (ie mapping host ports to different ports in containers ie the reverse of normal port mapping where we publish container ports to a different host port). Nomad doesn't support published ports that I can find. As such, the easiest thing to do is to use dnsmasq as a dns forwarder.

Note: We're using containerized dnsmasq instead of host-based dnsmasq that way we only have one place we're setting up dns

### DIFF
--- a/nomad/dns.nomad
+++ b/nomad/dns.nomad
@@ -1,0 +1,59 @@
+job "dns" {
+  datacenters = ["dc1"]
+  type        = "system"
+
+  group "dnsmasq" {
+    network {
+      mode = "bridge"
+      port "dns" {
+        static = 53
+        to     = 53
+      }
+    }
+
+
+    task "dnsmasq" {
+      driver = "docker"
+
+      config {
+        #This is an alpine-based dnsmasq container
+        image = "4km3/dnsmasq:2.85-r2"
+        ports = ["dns"]
+        args = [
+          # Send all queries for .consul to the NOMAD_IP
+          "--server", "/consul/${NOMAD_IP_dns}#8600",
+          # log to standard out
+          "--log-facility=-",
+        ]
+        cap_add = [
+          "NET_BIND_SERVICE",
+        ]
+        logging {
+          type = "journald"
+          config {
+            tag = "DNSMASQ"
+          }
+        }
+      }
+
+      service {
+        name         = "dnsmasq"
+        port         = "dns"
+        address_mode = "driver"
+        tags         = ["dns"]
+
+        check {
+          type     = "tcp"
+          port     = "dns"
+          interval = "10s"
+          timeout  = "2s"
+        }
+      }
+
+      resources {
+        cpu    = 50
+        memory = 100
+      }
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


nomad/dns.nomad
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+
+
+    task "dnsmasq" {
+      driver = "docker"
+
+      config {
+        #This is an alpine-based dnsmasq container
+        image = "4km3/dnsmasq:2.85-r2"
```
twunderlich-grapl, line 20: Long-term we'll want to switch this to either an official dnsmasq container or one that we build

nomad/dns.nomad
```diff
@@ -0,0 +1,59 @@
...
+        ports = ["dns"]
+        args = [
+          # Send all queries for .consul to the NOMAD_IP
+          "--server", "/consul/${NOMAD_IP_dns}#8600",
```
twunderlich-grapl, line 24: Note to self: doublecheck why we're doing it this way. It probably makes sense to switch to `attr.unique.network.ip-address`

twunderlich-grapl:
I still need to do some testing with this, but this is generally the approach I'd like to take to forward dns to the local Consul agent.

As an overview the dns call chain will look like
container -> dnsmasq on port 53 -> consul agent -> AWS dns.
Why?
This way we get access to all *.consul domains (used to forward observability requests to the otel collector) and then we get normal dns. Not either or.


Some risks that we need to take a look at:
What happens if dnsmasq fails to come up? Does it need to come up first?

nomad/dns.nomad
wimax-grapl: (how likely are we to actually follow up on this lol)

pulumi/grapl/__main__.py
```diff
@@ -314,6 +314,18 @@ def main() -> None:
...
 
+    NomadJob(
+        "dns",
+        jobspec=repository_path("nomad/dns.nomad"),
+        vars=dict(),
+        opts=pulumi.ResourceOptions(
+            provider=nomad_provider,
+            custom_timeouts=CustomTimeouts(
+                create=nomad_grapl_core_timeout, update=nomad_grapl_core_timeout
+            ),
```
wimax-grapl, line 325: honestly we prob don't need this for a single image

wimax-grapl (approved):

christophermaier (approved):

nomad/dns.nomad
twunderlich-grapl: This part seems like it will work both ways, so I'm going to leave it

twunderlich-grapl:
I've investigated alternative approaches:
1. For production, it may make sense to set up the consul terraform sync code. This is the model that k8s typically uses.
2 I've also looked into getting rid of all consul dns. Right now the only place this is being used is by Consul to figure out where to forward traces to. We could do this, but it would require a fair amount of hackery, especially since go-sockaddr isn't supported

christophermaier (approved):

PR #1983, opened 2022-09-14
Add in updates

colin-grapl:
1005

Same as https://github.com/grapl-security/grapl/pull/1982

Except I've added in the `Updates` construct. The reason there's an Updates as opposed to an Update is because the graph-merger still takes an entire graph as input,

```rust
    pub async fn handle_event(
        &mut self,
        tenant_id: uuid::Uuid,
        subgraph: IdentifiedGraph,
    ) -> Result<Updates, GraphMergerError> {
```
StreamProducer expects every iteration of the stream to yield `Result<Option<Envelope<T>>, _`, where T is SerDe, so there's no way for me to break the multiple updates up across multiple Envelope's.

In the future, if we so choose, we can update the StreamProcessor API so that this is possible, however we should almost certainly try to group some updates together as it should significantly improve Analyzer performance. I suspect that as we move towards the front of the pipeline, changing our graph types into smaller, bounded types, this will ultimately sort itself out (ie: if GraphMerger takes in one node it may still return N updates but N would be bounded to the number of properties on a node).

Existing tests, primarily. Some new serde tests.


src/python/python-proto/python_proto/tests/strategies.py
```diff
@@ -86,6 +89,12 @@
 small_text = st.text(max_size=4)
 
 
+def uids(
+    value: st.SearchStrategy[int] = st.integers(min_value=1, max_value=UINT64_MAX),
+) -> st.SearchStrategy[Uid]:
+    return st.builds(Uid, value=value)
+
```
wimax-grapl, line 96: This already exists, please import it:
```
from python_proto.tests.test_grapl_common import (
    uids,
)
```

src/python/python-proto/python_proto/tests/test_api_analyzers.py
```diff
@@ -68,13 +68,16 @@ def updates(
     return st.builds(analyzer_msgs.Update, inner=inner)
 
 
+def multiple_updates(
```
wimax-grapl, line 71: maybe lists_of_updates? I have 
```
# yeah, we ahve UidFilter and UidFilters
def uid_filters_plural(
```
lol

src/rust/graph-merger/src/config.rs
```diff
@@ -0,0 +1,6 @@
+#[derive(clap::Parser, Debug)]
+#[clap(name = "graph-merger", about = "Graph Merger Service")]
+pub struct GraphMergerConfig {
+    #[clap(env = "GRAPH_MUTATION_CLIENT_URL")]
+    pub graph_mutation_client_url: String,
+}
```
wimax-grapl, line 6: Nope. Use the pre-existing `src/rust/rust-proto/src/client_factory/services/graph_mutation.rs` like all existing code:

```
    let mutation_client_config = GraphMutationClientConfig::parse();
    let mut graph_mutation_client = build_grpc_client(mutation_client_config).await?;
```

wimax-grapl (changes requested):

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
```diff
@@ -53,6 +65,18 @@ async fn test_sysmon_log_e2e(ctx: &mut E2eTestContext) {
...
+        .create_tenant_keyspace(CreateTenantKeyspaceRequest { tenant_id })
+        .await?;
+
+    let provisioner_client_config = ScyllaProvisionerClientConfig::parse();
+    let mut provisioner_client = build_grpc_client(provisioner_client_config).await?;
+
+    provisioner_client
+        .provision_graph_for_tenant(ProvisionGraphForTenantRequest { tenant_id })
+        .await?;
+
```
wimax-grapl, line 79: These should likely be brought into an E2eTestContext function; this will be a very common functionality.

src/rust/graph-merger/src/config.rs
wimax-grapl: I fixed this

src/python/python-proto/python_proto/api/graph/v1beta1/messages.py
```diff
@@ -435,6 +436,54 @@ def into_proto(self) -> proto.EdgeList:
...
+        proto_identified_edge.edge_name = self.edge_name
+        return proto_identified_edge
+
+
+@dataclasses.dataclass(frozen=True, slots=True)
+class IdentifiedEdgeList(SerDe[proto.IdentifiedEdgeList]):
+    # TODO: seed to places where this is used:
+    # /src/python/grapl_analyzerlib/grapl_analyzerlib/view_from_proto.py
+    # /src/python/grapl_analyzerlib/grapl_analyzerlib/subgraph_view.py
+    edges: Sequence[IdentifiedEdge]
```
wimax-grapl, line 467: @colin-grapl fyi these are gone now so

src/rust/graph-merger/src/main.rs
```diff
@@ -105,56 +87,33 @@ async fn handler(
...
-                                Err(StreamProcessorError::from(e))
-                            }
-                        }
+                    GraphMergerError::Unexpected(ref reason) => {
+                        tracing::error!(
+                            message = "unexpected error",
+                            reason = %reason,
+                            error = %e,
+                        );
+                        // TODO: write message to failed topic here
```
wimax-grapl, line 106: are these TODOs still applicable?

src/rust/graph-merger/src/service.rs
```diff
@@ -55,26 +78,23 @@ impl From<&GraphMergerError> for kafka::StreamProcessorError {
 
 #[derive(Clone)]
 pub struct GraphMerger {
-    mg_client: Arc<DgraphClient>,
-    reverse_edge_resolver: ReverseEdgeResolver,
+    mg_client: GraphMutationClient,
```
wimax-grapl, line 81: `mg_client` meaning `master_graph_client` is so outdated i'm gonna rename this >:)

src/rust/graph-merger/src/service.rs
```diff
@@ -83,47 +103,123 @@ impl GraphMerger {
...
-        Ok(n) => n.as_millis(),
-        Err(_) => panic!("SystemTime before UNIX EPOCH!"),
-    };
-
-    let cur_day = cur_ms - (cur_ms % 86400);
-
-    format!("{}/{}-{}", cur_day, cur_ms, uuid::Uuid::new_v4())
+fn property_to_update(uid: Uid, property_name: String, property: &Property) -> Update {
+    match property {
+        Property::IncrementOnlyUintProp(_) => Update::Uint64Property(UInt64PropertyUpdate {
```
wimax-grapl, line 182: gonna simplify this function with ors

PR #1989, opened 2022-09-15
Replace build_grpc_client() usage with ConnectWithConfig blanket impl

wimax-grapl:
Refactor so getting a client is a bit more reasonable-looking, and something you can call on a client instead of this bare function.

CI!

### DIFF
--- a/src/rust/rust-proto/src/protocol/service_client.rs
+++ b/src/rust/rust-proto/src/protocol/service_client.rs
@@ -12,6 +12,8 @@ use crate::{
 /// Every service should implement Connectable.
 #[async_trait::async_trait]
 pub trait Connectable {
+    type Config: GrpcClientConfig + Send + Sync;
+
     /// Pass this NAME to e.g. a healthcheck client.
     const SERVICE_NAME: &'static str;
 
@@ -53,18 +55,9 @@ impl From<client_executor::Error<ConnectError>> for ConnectError {
 }
 
 #[async_trait::async_trait]
-pub trait ConfigConnectable
-where
-    Self: Connectable,
-{
-    type Config: GrpcClientConfig + Send + Sync;
-}
-
-#[async_trait::async_trait]
 pub trait ConnectWithConfig
 where
     Self: Connectable,
-    Self: ConfigConnectable,
 {
     async fn connect_with_config(client_config: Self::Config) -> Result<Self, ConnectError>
     where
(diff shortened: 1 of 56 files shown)


src/rust/rust-proto/src/protocol/service_client.rs
```diff
@@ -45,3 +51,32 @@ impl From<client_executor::Error<ConnectError>> for ConnectError {
         }
     }
 }
+
+#[async_trait::async_trait]
+pub trait ConfigConnectable
```
jgrillo-grapl, line 56: Why does this need to be a separate trait? Can't this `type` just go on `Connectable`?

src/rust/rust-proto/src/protocol/service_client.rs
```diff
@@ -45,3 +51,32 @@ impl From<client_executor::Error<ConnectError>> for ConnectError {
...
+where
+    Self: Connectable,
+{
+    type Config: GrpcClientConfig + Send + Sync;
+}
+
+#[async_trait::async_trait]
+pub trait ConnectWithConfig
```
jgrillo-grapl, line 64: Same question--is there ever a case where a `Connectable` _doesn't_ have a config? In that case, it seems like all this stuff should just go on `Connectable`.

jgrillo-grapl (approved):

src/rust/rust-proto/src/protocol/service_client.rs
wimax-grapl: I believe the answer is "potentially"; I'd previously thought about the following:
```

fn get_plugin_upstream_address(plugin_id: uuid::Uuid) -> String {
    let upstream_addr_env_var = format!("NOMAD_UPSTREAM_ADDR_plugin-{plugin_id}");
    let upstream_addr = std::env::var(&upstream_addr_env_var).expect(&upstream_addr_env_var);
    let address = format!("http://{upstream_addr}");
    address
}

/// Create a client from environment
pub async fn get_generator_client(
    plugin_id: uuid::Uuid,
) -> Result<GeneratorServiceClient, ConnectError> {
    let address = get_plugin_upstream_address(plugin_id);
    let client_config = GeneratorClientConfig {
        generator_client_address: address.parse().expect("generator_client_address"),
    };
    GeneratorServiceClient::connect_with_config(client_config).await
    ```
    
  where we manually construct the Config instead of doing anything clap/parse-y with it

PR #1992, opened 2022-09-16
Update pants run syntax after 2.13 change

ghost:

This updates a pants run command to be more explicit about which Pants target to use due to a change in Pants 2.13

For reference:
https://buildkite.com/grapl/grapl-provision/builds/603#0183466b-f092-4e89-80c6-736920e24673

```
Currently, running a `pex_binary` by default will not include the source files in the PEX, and will instead put them in a temporary sandbox.

In Pants 2.14, the default will change to instead build the PEX like you had run the `package` goal, and then execute that PEX. This is more consistent and intuitive behavior.
```
This results in

```
TooManyTargetsException: The `run` goal only works with one valid target, but was given multiple valid targets:

  * etc/ci_scripts/dump_artifacts/__init__.py:dump_artifacts_lib
  * etc/ci_scripts/dump_artifacts/cli.py:dump_artifacts_lib
  * etc/ci_scripts/dump_artifacts/docker_artifacts.py:dump_artifacts_lib
  * etc/ci_scripts/dump_artifacts/nomad_artifacts.py:dump_artifacts_lib
  * etc/ci_scripts/dump_artifacts:dump_artifacts
```

Basically its looking at both the python_source targets and the pex_binary targets and can't figure out which to use, so we need to be explicit

Ran `./pants run ./etc/ci_scripts/dump_artifacts -- --no-dump-agent-logs` and reproduced the failure in CI.
Ran `./pants run ./etc/ci_scripts/dump_artifacts:dump_artifacts -- --no-dump-agent-logs` and the script ran correctly

### DIFF
--- a/.buildkite/scripts/grapl_testing_stack/dump_artifacts.sh
+++ b/.buildkite/scripts/grapl_testing_stack/dump_artifacts.sh
@@ -17,5 +17,5 @@ readonly _NOMAD_ADDRESS
 
 cd "${REPOSITORY_ROOT}"
 NOMAD_ADDRESS="${_NOMAD_ADDRESS}" ./pants run \
-    ./etc/ci_scripts/dump_artifacts -- \
+    ./etc/ci_scripts/dump_artifacts:dump_artifacts -- \
     --no-dump-agent-logs
--- a/.buildkite/scripts/grapl_testing_stack/dump_artifacts.sh
+++ b/.buildkite/scripts/grapl_testing_stack/dump_artifacts.sh
@@ -16,6 +16,5 @@ _NOMAD_ADDRESS=$(pulumi stack output address --stack="${STACK}")
 readonly _NOMAD_ADDRESS
 
 cd "${REPOSITORY_ROOT}"
-NOMAD_ADDRESS="${_NOMAD_ADDRESS}" ./pants run \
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/grapl_testing_stack/dump_artifacts.sh
```diff
@@ -17,5 +17,5 @@ readonly _NOMAD_ADDRESS
 
 cd "${REPOSITORY_ROOT}"
 NOMAD_ADDRESS="${_NOMAD_ADDRESS}" ./pants run \
-    ./etc/ci_scripts/dump_artifacts -- \
+    ./etc/ci_scripts/dump_artifacts:dump_artifacts -- \
```
christophermaier, line 20: It would be better to use our alias: https://github.com/grapl-security/grapl/blob/3b8e01a2dfc10c68103d969371025e4e8aee7c8a/pants.toml#L239

This would then just be `./pants dump-artifacts --no-dump-agent-logs`

(I was wracking my brain trying to figure out how this made it through, when I was sure I'd updated this command for the new Pants syntax... guess I missed this invocation.)

christophermaier (changes requested):

christophermaier (approved):

PR #1995, opened 2022-09-20
Fix otel collector metric scraping

ghost:

This fixes a few issues with metric scraping from the otel collector
1. The scrape interval was less than the scrape timeout
2. Setting up the metrics pipeline properly
3. Getting the otel collector to use a functioning endpoint for Nomad

make up and then checking the otel-collector logs

### DIFF
--- a/pulumi/infra/config.py
+++ b/pulumi/infra/config.py
@@ -21,6 +21,11 @@ GRAPL_TEST_USER_NAME: Final[str] = f"{STACK_NAME}-grapl-test-user"
 # note: this ${} is interpolated inside Nomad
 HOST_IP_IN_NOMAD: Final[str] = "${attr.unique.network.ip-address}"
 
+# This is equivalent to what "${attr.unique.network.ip-address}" resolves to but is used for cases where variable
+# interpolation is not available. Currently this is used to get the private IP for Nomad so the otel collector can
+# scrape metrics
+LOCAL_HOST_IP: Final[str] = os.getenv("LOCAL_HOST_ETH0_IP")
+
 
 def repository_path(relative_path: str) -> Path:
     """
--- a/pulumi/infra/config.py
+++ b/pulumi/infra/config.py
@@ -21,11 +21,6 @@ GRAPL_TEST_USER_NAME: Final[str] = f"{STACK_NAME}-grapl-test-user"
 # note: this ${} is interpolated inside Nomad
 HOST_IP_IN_NOMAD: Final[str] = "${attr.unique.network.ip-address}"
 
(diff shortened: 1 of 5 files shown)


pulumi/infra/observability_env_vars.py
```diff
@@ -57,13 +59,14 @@ def otel_config(
     config:
       scrape_configs:
         - job_name: 'nomad-server'
-          scrape_interval: 10s
```
twunderlich-grapl, line 60: The scrape interval has to be greater than the scrape timeout, otherwise the scraping times out before any scraping is actually conducted.
Essentially these are absolute times, scrape_timeout is not relative to the scrape_interval

pulumi/infra/observability_env_vars.py
```diff
@@ -57,13 +59,14 @@ def otel_config(
...
-          scrape_timeout: 20s
-          metrics_path: '/v1/metrics?format=prometheus'
+          scrape_interval: 20s
+          scrape_timeout: 10s
+          metrics_path: '/v1/metrics'
           params:
             format: ['prometheus']
           static_configs:
-            - targets: ['localhost:4646']
```
twunderlich-grapl, line 66: This doesn't work from inside a container.

jgrillo-grapl (approved):
LGTM

pulumi/infra/observability_env_vars.py
christophermaier: I don't want to derail this PR, or the work you've already put into solving this problem (after all, a solution that works is much better than one that doesn't!), but are you familiar with any way of configuring the containers or their networking such that we don't have to jump through so many hoops to get this wired together?

If the answer is "no, this is how we do this", that's fine... it just seems like a lot of fiddly bits to deal with.

pulumi/infra/config.py
```diff
@@ -21,6 +21,11 @@
 # note: this ${} is interpolated inside Nomad
 HOST_IP_IN_NOMAD: Final[str] = "${attr.unique.network.ip-address}"
 
+# This is equivalent to what "${attr.unique.network.ip-address}" resolves to but is used for cases where variable
+# interpolation is not available. Currently this is used to get the private IP for Nomad so the otel collector can
+# scrape metrics
+LOCAL_HOST_IP: Final[str] = os.environ["LOCAL_HOST_ETH0_IP"]
```
christophermaier, line 27: Given that this variable is provided in contrast with `HOST_IP_IN_NOMAD`, it may be better to name it something like `HOST_IP_OUTSIDE_NOMAD` (ugh, that name _sucks_ 🤢) to more clearly tie it to the other variable, and to better highlight its intended usage.

christophermaier (approved):

PR #1998, opened 2022-09-21
Add Login Form & Sign In With Google (unstyled)

andrea-grapl:
https://github.com/grapl-security/issue-tracker/issues/1012

Move over auth code from `engagement_view`. Add login form, login service, and SSO with Google. 

Run Grapl, log in with username `local-grapl-grapl-test-user` and Secrets Managers generated auth token. Also click SSO button and make sure login is successful.


src/js/frontend/src/components/login/Login.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { signInWithGoogleService } from "../../services/login/signInWithGoogleService";
+import { loginService } from "../../services/login/loginService";
+
+const Login = () => {
+  const [state, setState] = React.useState({
+    loginFailed: false,
+  });
+
+  const validationSchema = Yup.object().shape({
+    userName: Yup.string().required("Username Required"),
```
colin-grapl, line 15: 'username' is fine, no need for 'userName'

src/js/frontend/src/components/login/Login.tsx
```diff
@@ -0,0 +1,124 @@
...
+    password: Yup.string().required("Password Required"),
+  });
+
+  return (
```
colin-grapl, line 19: This code block is really large and complicated. Can we somehow break it up? Like move the GoogleLogin 'on success' code to a function, same with the'onSubmit' callback. A comment on the 'loginFailed' that describes when it changes from true -> false or false -> true would be helpful as well.

src/js/frontend/src/components/login/Login.tsx
andrea-grapl: ✅

inickles-grapl (approved):

PR #2000, opened 2022-09-21
nomad_endpoint must be a valid hostname

ghost:

This fixes the following error in staging

```
Error: cannot unmarshal the configuration: error reading receivers configuration for "prometheus": prometheus receiver failed to unmarshal yaml to prometheus config: "http://internal-nomad-alb-31268fd-999032172.us-east-1.elb.amazonaws.com:4646" is not a valid hostname
2022/09/21 19:14:07 collector server run finished with error: cannot unmarshal the configuration: error reading receivers configuration for "prometheus": prometheus receiver failed to unmarshal yaml to prometheus config: "http://internal-nomad-alb-31268fd-999032172.us-east-1.elb.amazonaws.com:4646" is not a valid hostname

```

The nomad provider address is formatted like http://internalalb.foo.com:4646. This is an invalid hostname, because the http:// is for port 80 and we're using a different port. The otel-collector blows up when unmarshalling the yaml config, because of the invalid hostname. As such, we're stripping out the prepended http:// to make the alb url a valid hostname.

Tested by temporarily adding the following lines into the local grapl logic:

```
    nomad_endpoint = pulumi.Output.from_input("http://internal-nomad-alb-31268fd-999032172.us-east-1.elb.amazonaws.com:4646")

    nomad_endpoint = pulumi.Output.all(
        nomad_endpoint=nomad_endpoint,
    ).apply(lambda endpoint: urlparse(f"{nomad_endpoint}").netloc)
```
running `make up` and checking the logs for the otel-collector

### DIFF
--- a/pulumi/grapl/__main__.py
+++ b/pulumi/grapl/__main__.py
@@ -134,9 +134,22 @@ def main() -> None:
     nomad_provider: pulumi.ProviderResource | None = None
     consul_provider: pulumi.ProviderResource | None = None
 
+    # Use the nomad ALB address or default to the local host address.
+    # Apparently nomad templates use a different templating syntax than anywhere else
+    nomad_endpoint: pulumi.Output[
+        str
+    ] | str = '{{ env "attr.unique.network.ip-address" }}:4646'
+
     if not config.LOCAL_GRAPL:
         upstream_stacks = UpstreamStacks()
         nomad_provider = get_nomad_provider_address(upstream_stacks.nomad_server)
+
+        nomad_endpoint = nomad_provider.address
+        # we need to strip off the prepended http:// because that combined with a non-standard port is
+        nomad_endpoint = pulumi.Output.all(
+            nomad_endpoint=nomad_endpoint,
+        ).apply(lambda endpoint: f"{nomad_endpoint}".replace("http://", ""))
+
         # Using get_output instead of require_output so that preview passes.
         # NOTE wimax Feb 2022: Not sure the above is still the case
         consul_master_token_secret_id = upstream_stacks.consul.get_output(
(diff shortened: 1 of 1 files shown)


pulumi/grapl/__main__.py
```diff
@@ -134,9 +134,22 @@ def main() -> None:
...
+    # Apparently nomad templates use a different templating syntax than anywhere else
+    nomad_endpoint: pulumi.Output[
+        str
+    ] | str = '{{ env "attr.unique.network.ip-address" }}:4646'
+
     if not config.LOCAL_GRAPL:
         upstream_stacks = UpstreamStacks()
         nomad_provider = get_nomad_provider_address(upstream_stacks.nomad_server)
+
+        nomad_endpoint = nomad_provider.address
```
christophermaier, line 147: Do we need this intermediate variable? Could we just do the processing directly on `nomad_provider.address`?

pulumi/grapl/__main__.py
```diff
@@ -134,9 +134,22 @@ def main() -> None:
...
+        # we need to strip off the prepended http:// because that combined with a non-standard port is
```
christophermaier, line 148: "is..." what?

THE SUSPENSE IS KILLING ME! :joy:

pulumi/grapl/__main__.py
```diff
@@ -134,9 +134,22 @@ def main() -> None:
...
+        nomad_endpoint = pulumi.Output.all(
+            nomad_endpoint=nomad_endpoint,
+        ).apply(lambda endpoint: f"{nomad_endpoint}".replace("http://", ""))
```
christophermaier, line 151: Is there any possibility of this being `https://` at any point in the future?

I think there are URL libraries that can be used to strip the protocol portion off a URL generically... might be worthwhile here.

christophermaier (approved):
Some minor suggestions, but otherwise 👍

pulumi/grapl/__main__.py
twunderlich-grapl: No we don't, fixing now.

pulumi/grapl/__main__.py
twunderlich-grapl: I see what you're saying. I don't think it makes sense to pull in a dependency just to do this. Let me test using .split on '://'

christophermaier: I think you can just use [urllib.parse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse):
```python
urllib.parse(nomad_endpoint).netloc
```

PR #2002, opened 2022-09-22
Graph-merger now emits Updates, not MergedGraph.

wimax-grapl:
Resurrecting PR https://github.com/grapl-security/grapl/pull/1983 which was closed by accident.

# Which issue does this PR correspond to?
1005

# What changes does this PR make to Grapl? Why?
Same as https://github.com/grapl-security/grapl/pull/1982

Except I've added in the Updates construct. The reason there's an Updates as opposed to an Update is because the graph-merger still takes an entire graph as input,

    pub async fn handle_event(
        &mut self,
        tenant_id: uuid::Uuid,
        subgraph: IdentifiedGraph,
    ) -> Result<Updates, GraphMergerError> {
StreamProducer expects every iteration of the stream to yield Result<Option<Envelope<T>>, _, where T is SerDe, so there's no way for me to break the multiple updates up across multiple Envelope's.

In the future, if we so choose, we can update the StreamProcessor API so that this is possible, however we should almost certainly try to group some updates together as it should significantly improve Analyzer performance. I suspect that as we move towards the front of the pipeline, changing our graph types into smaller, bounded types, this will ultimately sort itself out (ie: if GraphMerger takes in one node it may still return N updates but N would be bounded to the number of properties on a node).

# How were these changes tested?
Existing tests, primarily. Some new serde tests.


src/python/python-proto/python_proto/api/graph/v1beta1/messages.py
```diff
@@ -342,7 +343,7 @@ def into_proto(self) -> proto.NodeDescription:
 @dataclasses.dataclass(frozen=True)
```
colin-grapl, line 343: Probably all of these should have `slots=True` too

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
```diff
@@ -195,27 +194,30 @@ async fn test_sysmon_log_e2e(ctx: &mut E2eTestContext) {
...
-            let envelope = envelope.clone();
-            let merged_graph = envelope.inner_message();
-            events_36lines_merged_graph_predicate(merged_graph)
+    assert!(graph_updates.len() >= input_log_lines.len());
+
+    // flatmap vec<vec<update>> to vec<update>
+    let mut updates: Vec<Update> = vec![];
+    graph_updates.into_iter().for_each(|updates_envelope| {
+        updates.append(&mut updates_envelope.inner_message().updates);
+    });
```
colin-grapl, line 206: There's a flatten for cases like this.

```
let updates = graph_updates.into_iter().map(|env| env.inner_message().updates).flatten().collect();
```

src/rust/graph-mutation/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(warnings)]
```
colin-grapl, line 1: Should remove

src/rust/plugin-execution-sidecar/src/work/analyzer_work_processor.rs
```diff
@@ -0,0 +1,160 @@
+use rust_proto::{
```
wimax-grapl, line 1: oh, this should be removed from this PR

colin-grapl (approved):

PR #2006, opened 2022-09-23
remove references to graphql-endpoint

andrea-grapl:
https://github.com/grapl-security/issue-tracker/issues/1029

Remove references to GraphQL Endpoint because we're no longer using Graphql
to retrieve data on the frontend.
 
Searched for references to `graphql endpoint` in our codebase. Because this PR is removing functionality, success of existing integration test should be sufficient.


pulumi/grapl/__main__.py
```diff
@@ -301,13 +300,6 @@ def main() -> None:
         opts=pulumi.ResourceOptions(provider=consul_provider),
     )
 
-    ConsulServiceDefault(
```
twunderlich-grapl, line 304: There's a possibility that we may need to add this back temporarily so we can delete the service first and then delete this. No action needed now, but that's what we'll have to do if we run into problems with pulumi

twunderlich-grapl (approved):
LGTM

PR #2014, opened 2022-09-26
Merge tmp-migration (Graph Migration) into main

wimax-grapl:
This is strictly a merge.

Primarily, it ports graph-merger and node-identifier to use scylla, and emit Updates instead of MergedGraphs


nomad/grapl-core.nomad
```diff
@@ -189,19 +194,23 @@ locals {
   dgraph_alpha_grpc_public_port_base  = 9080
 
   # DGraph Alphas (shards * replicas)
```
wimax-grapl, line 196: after this merge i'll do a separate PR to rip out any remaining dgraph stuff

colin-grapl (approved):

src/rust/graph-mutation/src/reverse_edge_resolver.rs
```diff
@@ -36,33 +37,50 @@ impl ReverseEdgeResolver {
...
 
-                let reverse_name = response.reverse_edge_name.clone();
-                entry.insert(response);
-                Ok(reverse_name)
-            }
-        }
+        let reverse_name = response.reverse_edge_name;
+        Ok(reverse_name)
+
+        // TODO: Resurrect the below once we figure out caching for Graph Mutation
```
jgrillo-grapl, line 59: commented code can be removed

jgrillo-grapl (commented):
One thing, otherwise LGTM

src/rust/graph-mutation/src/reverse_edge_resolver.rs
wimax-grapl: Fixing these is high on our list of things to do (like, this week), and the uncommented code will likely be exactly the same as what's already here [barring some type changes], so I'm intentionally keeping it as-is. However, I can guarantee I'll resolve these comments by Friday.

PR #2024, opened 2022-09-30
Fix otel-collector Lightstep export

ghost:

1. The otel-collector is checking for whether the endpoint is insecure, not whether its secure. I've renamed the variable to make that clear and flipped everything so it should just work now
2. Quote the lightstep access token

Manually made the changes in staging and checked to see if traces came into lightstep. They do.

### DIFF
--- a/pulumi/grapl/__main__.py
+++ b/pulumi/grapl/__main__.py
@@ -339,8 +339,8 @@ def main() -> None:
         pulumi_config.get(key="lightstep-access-token") or ""
     )
     lightstep_endpoint = pulumi_config.require(key="lightstep-endpoint")
-    lightstep_is_endpoint_insecure = (
-        pulumi_config.get(key="lightstep-is-endpoint-insecure") or "false"
+    lightstep_is_endpoint_insecure = pulumi_config.get_bool(
+        key="lightstep-is-endpoint-insecure", default=False
     )
 
     otel_configuration = otel_config(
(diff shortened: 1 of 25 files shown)


pulumi/grapl/__main__.py
```diff
@@ -339,15 +339,15 @@ def main() -> None:
         pulumi_config.get(key="lightstep-access-token") or ""
     )
     lightstep_endpoint = pulumi_config.require(key="lightstep-endpoint")
-    lightstep_is_endpoint_secure = (
-        pulumi_config.get(key="lightstep-is-endpoint-secure") or "true"
+    lightstep_is_endpoint_insecure = (
+        pulumi_config.get(key="lightstep-is-endpoint-insecure") or "false"
```
christophermaier, line 343: This could be simplified to `pulumi_config.get("lightstep-is-endpoint-secure", "false")`, in line with the normal Python behavior. The `key` keyword is kind of noisy here, given existing Python idioms around this kind of function.

Do these have to be strings? Because `pulumi.Config` has a `get_bool` function that returns booleans. It's probably better to have these as Booleans, and then convert them to strings at the point of ultimate consumption, if the consumer _has_ to have them as strings.

christophermaier (approved):

pulumi/grapl/__main__.py
twunderlich-grapl: Testing whether I can use booleans now.

PR #2026, opened 2022-09-30
Greatly simplify how we do consul connect with plugins by leveraging Consul Namespaces. 

wimax-grapl:
None. Somewhat precursor work for GQS Proxy.

1. Greatly simplify how we do consul connect with plugins by leveraging Consul Namespaces. 
Now the upstream does not have a dynamic name (it's just "plugin", not "plugin-uuid-12345-65453-432423")

2. Upgrade the nomad CLI bundled with Plugin Registry; it uses this to parse the Nomad file, and it needed a much more recent version for Consul Namespaces.

3. Here and there I've gotten Scylla OOMs upon startup, adding a memory_max fixed it.

CI


src/rust/Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN --mount=type=cache,target=/var/lib/apt/lists,sharing=locked,id=rust-base-apt
 # - in plugin-registry image
 WORKDIR /nomad
 RUN <<EOF
-NOMAD_VERSION="1.2.4"
+NOMAD_VERSION="1.3.5"
```
jgrillo-grapl, line 38: should this be `"1.3.5-1"` to match what's in etc/chromeos/lib/installs.sh?

src/rust/build-env.Dockerfile
```diff
@@ -27,7 +27,7 @@ EOF
 # - in plugin-registry unit tests
 WORKDIR /nomad
 RUN <<EOF
-NOMAD_VERSION="1.2.4"
+NOMAD_VERSION="1.3.5"
```
jgrillo-grapl, line 30: same

src/rust/plugin-registry/src/static_files/hax_docker_analyzer.nomad
```diff
@@ -77,14 +81,16 @@ job "grapl-plugin" {
...
+
     network {
       mode = "bridge"
       port "analyzer-execution-sidecar" {}
       port "graph-query-proxy" {}
     }
 
     service {
-      name = "analyzer-exec-sidecar-${var.plugin_id}"
+      name = "analyzer-exec-sidecar"
```
jgrillo-grapl, line 93: this can now be the un-abbreviated `"analyzer-execution-sidecar"`

src/rust/plugin-registry/src/static_files/hax_docker_analyzer.nomad
```diff
@@ -95,21 +101,37 @@ job "grapl-plugin" {
...
+              destination_name      = "plugin-work-queue"
+              destination_namespace = "default"
+              local_bind_port       = 1000
             }
 
             upstreams {
-              destination_name = "plugin-${var.plugin_id}"
-              # port unique but arbitrary - https://github.com/hashicorp/nomad/issues/7135
-              local_bind_port = 1001
+              destination_name      = "plugin"
```
jgrillo-grapl, line 110: would it be possible to call this "analyzer" instead of "plugin"? It would be nice to make the corresponding name "generator" for the generator plugins as well

jgrillo-grapl (approved):
some naming nits, LGTM

src/rust/Dockerfile
twunderlich-grapl: The `-1` at the end is a linux repository thing, and doesn't appear to be used in the releases page

etc/chromeos/lib/installs.sh
```diff
@@ -267,7 +267,7 @@ install_hashicorp_tools() {
     echo_banner "Installing hashicorp tools: consul nomad packer"
 
     CONSUL_VERSION="1.12.5-1"
-    NOMAD_VERSION="1.3.5-1"
+    NOMAD_VERSION="1.3.5-1" # Also update build-env.Dockerfile, rust/Dockerfile
```
twunderlich-grapl, line 270: Thanks for adding this. I had forgotten about these (clearly)

src/rust/plugin-registry/src/static_files/hax_docker_generator.nomad
```diff
@@ -138,6 +143,8 @@ job "grapl-plugin" {
   }
 
   group "plugin" {
+    consul { namespace = local.namespace }
```
twunderlich-grapl, line 146: This may not work as expected since per the docs it appears to be enterprise only: https://www.nomadproject.io/docs/job-specification/group#namespace

twunderlich-grapl: Actually @wimax-grapl what you probably want is to run the entire job in a Nomad namespace, which is supported per https://learn.hashicorp.com/tutorials/nomad/namespaces#run-a-job-in-a-namespace

src/rust/plugin-registry/src/static_files/hax_docker_analyzer.nomad
wimax-grapl: done

src/rust/plugin-registry/src/static_files/hax_docker_generator.nomad
wimax-grapl: Surprise! Nomad namespaces are not the same as Consul namespaces. (However, the jobs already have their own Nomad namespace with the same name)

jgrillo-grapl (approved):

wimax-grapl:
Abandoning this due to https://github.com/hashicorp/nomad/issues/14810
if we ever decide to shell out for consul enterprise, let's reconsider this again.

PR #2028, opened 2022-10-04
Remove partitioning from plugin work queue

wimax-grapl:
None

We were seeing some serious issues with the plugin work queue running locally (it'd OOM), and we can't quite figure out why. Based on the `explain` there appeared to be a serious number of partitions utilized by the `get_generator_messages` (and analyzer) query in psql_queue.rs. 

This is an experiment: to see if we get better behavior by removing the partitioning.

CI

### DIFF
--- a/nomad/local/grapl-local-infra.nomad
+++ b/nomad/local/grapl-local-infra.nomad
@@ -33,7 +33,6 @@ locals {
     {
       name       = "plugin-work-queue-db",
       port       = 5433,
-      memory_max = 1024,
     },
     {
       name = "organization-management-db",
--- a/nomad/local/grapl-local-infra.nomad
+++ b/nomad/local/grapl-local-infra.nomad
@@ -31,8 +31,8 @@ locals {
       port = 5432,
     },
     {
-      name       = "plugin-work-queue-db",
-      port       = 5433,
+      name = "plugin-work-queue-db",
+      port = 5433,
     },
     {
       name = "organization-management-db",
--- a/src/rust/plugin-work-queue/migrations/1640037910_plugin_work_queue_migration.sql
+++ b/src/rust/plugin-work-queue/migrations/1640037910_plugin_work_queue_migration.sql
@@ -71,8 +71,3 @@ CREATE INDEX IF NOT EXISTS creation_time_ix ON plugin_work_queue.generator_plugi
 
 CREATE INDEX IF NOT EXISTS execution_key_ix ON plugin_work_queue.analyzer_plugin_executions (execution_key);
 CREATE INDEX IF NOT EXISTS creation_time_ix ON plugin_work_queue.analyzer_plugin_executions (creation_time);
-
-CREATE SCHEMA IF NOT EXISTS partman;
-CREATE EXTENSION IF NOT EXISTS pg_partman WITH SCHEMA partman;
-CREATE EXTENSION IF NOT EXISTS pg_cron;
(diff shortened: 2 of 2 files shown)


nomad/local/grapl-local-infra.nomad
```diff
@@ -31,9 +31,8 @@ locals {
       port = 5432,
     },
     {
-      name       = "plugin-work-queue-db",
-      port       = 5433,
-      memory_max = 1024,
```
jgrillo-grapl, line 36: probably worth keeping the `memory_max` here unless there's a compelling reason to remove it

src/rust/plugin-work-queue/migrations/1640037910_plugin_work_queue_migration.sql
```diff
@@ -74,33 +74,5 @@ CREATE INDEX IF NOT EXISTS creation_time_ix ON plugin_work_queue.analyzer_plugin
...
-UPDATE partman.part_config
-SET infinite_time_partitions = true,
-    retention                = '1 month',
-    retention_keep_table= false
-WHERE (
-                  parent_table = 'plugin_work_queue.generator_plugin_executions' OR
-                  parent_table = 'plugin_work_queue.analyzer_plugin_executions'
-          );
-
 SELECT cron.schedule('@hourly', $$CALL partman.run_maintenance_proc()$$);
```
jgrillo-grapl, line 78: this should also be removed

src/rust/plugin-work-queue/migrations/1640037910_plugin_work_queue_migration.sql
```diff
@@ -74,33 +74,5 @@ CREATE INDEX IF NOT EXISTS creation_time_ix ON plugin_work_queue.analyzer_plugin
 
 CREATE SCHEMA IF NOT EXISTS partman;
```
jgrillo-grapl, line 75: this stuff needs to go also

colin-grapl (approved):

jgrillo-grapl (changes requested):

nomad/local/grapl-local-infra.nomad
wimax-grapl: It's more like, the only compelling reason it was there in the first place was to get around the OOMs for PWQ. By removing this memory_max= we help validate that this was, indeed, the underlying issue.

src/rust/plugin-work-queue/migrations/1640037910_plugin_work_queue_migration.sql
wimax-grapl: sure

src/rust/plugin-work-queue/migrations/1640037910_plugin_work_queue_migration.sql
wimax-grapl: i guess if we're not using it, sure

nomad/local/grapl-local-infra.nomad
jgrillo-grapl: oh, I see, it "defaults" to a non-default here: https://github.com/grapl-security/grapl/blob/main/nomad/local/grapl-local-infra.nomad#L36

so leaving this unset will in fact _not_ default to 300MB it'll get set to 512MB

>>_<<

jgrillo-grapl (approved):

PR #2029, opened 2022-10-05
Fix a minor shell bug in log.sh

wimax-grapl:
If you have `-euo pipefail` set while calling this function, but do not specify EXIT_CODE, it fails.

`-v` checks if such a variable exists. 

Yes, the syntax is, in fact, correct! No quotes, no $


twunderlich-grapl (approved):

christophermaier (approved):
Good catch.

PR #2030, opened 2022-10-05
begin adding analyzer tests to e2e tests; rename some things to make clear they're about Generators

wimax-grapl:
None

Start pushing the final part of e2e forward by:
1. deploying an analyzer
2. adding further tests down the pipeline

Also, in doing so it became clear there's a lot of places I was using the verbiage "plugin" where I really mean "generator plugin", so lots of changes there.

Passed CI on my devbox

### DIFF
--- a/src/rust/e2e-tests/src/test_utils/context.rs
+++ b/src/rust/e2e-tests/src/test_utils/context.rs
@@ -120,9 +120,9 @@ impl AsyncTestContext for E2eTestContext {
 }
 
 #[derive(Debug)]
-pub struct SetupResult {
+pub struct SetupGeneratorResult {
     pub tenant_id: Uuid,
-    pub plugin_id: Uuid,
+    pub generator_plugin_id: Uuid,
     pub event_source_id: Uuid,
 }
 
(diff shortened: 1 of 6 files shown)


jgrillo-grapl (approved):

src/rust/e2e-tests/src/test_utils/context.rs
```diff
@@ -255,11 +255,24 @@ impl E2eTestContext {
...
     }
 
+    pub async fn setup_suspicious_svchost_analyzer(
+        &mut self,
+        tenant_id: uuid::Uuid,
+        test_name: &str,
+    ) -> eyre::Result<uuid::Uuid> {
+        let analyzer_artifact = test_fixtures::get_suspicious_svchost_analyzer()?;
+        let analyzer_plugin_id = self
+            .create_analyzer(tenant_id, test_name.to_owned(), analyzer_artifact)
```
colin-grapl, line 265: This especially doesn't matter because it's a test fn, but just fwiw I would say that if you're going to require a String internally you should take a String in your API, *or* `impl Into<String>`, which would let callers pass in &str or String.

Purely style, pls do not feel like you have to fix this

wimax-grapl: cool 2 know

colin-grapl (approved):

PR #2031, opened 2022-10-05
Add a Docker Plugin Runtime image, which is just python debian bullseye slim for now.

wimax-grapl:
None

Make our plugins run inside a Docker image that has py3.10 installed.
Instead of hardcoding it to Just Python Bullseye Slim, I'd imagine we could eventually want to shoehorn more specific stuff into this Docker image, so I gave it its own dockerfile.
image

In unison with https://github.com/grapl-security/grapl/pull/2030/files
I got as far as the following with an analyzer
```
  File "/root/.pex/unzipped_pexes/106a54a03c284153816eabba69ce6564e8a7b438/python_proto/api/graph_query/v1beta1/client.py", line 30, in connect
    address = os.environ["GRAPH_QUERY_CLIENT_ADDRESS"]
  File "/usr/local/lib/python3.10/os.py", line 679, in __getitem__
    raise KeyError(key) from None
KeyError: 'GRAPH_QUERY_CLIENT_ADDRESS'
```

I need to change it to use Graph Query Proxy, anyway, but this proves our PEX is being deployed and running.


colin-grapl (approved):

docker-bake.hcl
```diff
@@ -232,9 +240,9 @@ group "all" {
   # NOTE: Please keep this list sorted in alphabetical order
   targets = [
     "all-tests",
-    "local-only-services",
-    "grapl-services",
     "export-rust-build-artifacts-to-dist",
+    "grapl-services",
+    "local-only-services",
```
christophermaier, line 245: Bless your alphabetizing ways :pray:

docker-bake.hcl
```diff
@@ -459,6 +467,23 @@ target "uid-allocator" {
...
 
 
+# Plugin Runtime
+# ----------------------------------------------------------------------
+
+# This is the Docker image that will host our Generators and Analyzers
+# until we swap over to Firecracker.
+target "docker-plugin-runtime" {
+  inherits   = ["_grapl-base"]
+  context    = "."
```
christophermaier, line 477: Not a huge deal at this point, but it may be a good idea to create a separate directory for this, just for general context management / hygiene. So rather than `Dockerfile.plugin-runtime`, we'd have `plugin-runtime/Dockerfile`, with `plugin-runtime` as the context.

Again, no big deal now, particularly given the image doesn't really do anything yet 😅

Dockerfile.plugin-runtime
```diff
@@ -0,0 +1,8 @@
+# This is the Docker image that Grapl plugins will run inside until we swap
+# over to Firecracker. 
+# It needs to be able to run Rust binaries (nothing special) and 
+# python PEXs (needs py3.10).
+
+ARG PYTHON_VERSION
+
+FROM python:${PYTHON_VERSION}-slim-bullseye AS docker-plugin-runtime
```
christophermaier, line 8: A slim-bullseye image is a good starting point... do we have any notion of what other requirements will be needed for actually executing plugins beyond a Python runtime? It'd be nice to have as small a surface area as possible. But, as said in another comment, it's obviously not a huge deal at the moment.

christophermaier (approved):
Nice formalization of this at an early stage 👍

docker-bake.hcl
wimax-grapl: I'm amenable to that.

Dockerfile.plugin-runtime
wimax-grapl: That's about it for now

PR #2032, opened 2022-10-05
Rename Protobufs from 'graph-query-service' to 'graph-query' to bring it in line with our service naming conventions

wimax-grapl:
Just a big ol' rename-o-rama, and some changes to `buf` to allow the deletion of unused, unstable packages.


src/rust/Cargo.toml
```diff
@@ -13,7 +13,7 @@ members = [
   "./generators/sysmon-generator",
   "./graph-merger",
   "./graph-mutation",
-  "./graph-query-service", # TODO rename dir
+  "./graph-query"
```
wimax-grapl, line 16: # Behold, my motivation for this change

wimax-grapl: (it's the only time 'service' shows up in this entire file)

colin-grapl (approved):

PR #2034, opened 2022-10-06
Add a version tag to all traces

ghost:

This adds a version tag to all service traces (but not to traces coming from Consul). This will allow us to easily see what version of a service bugs or performance changes were introduced and could potentially be used for automatic remediation further down the line.

I've opted to use the OTEL_RESOURCE_ATTRIBUTES env var to implement tags in open-telemetry traces. This way it's relatively easy to add new tags, even on a per-service basis. This opens up adding tags like instance_id, team or even language when we're ready 

Tested using Lightstep developer mode and confirmed that it showed up in the UI (under the details, not attributes 🤔 )
Screenshot 2022-10-10 1 01 21 PM

### DIFF
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -177,6 +177,14 @@ locals {
 
   # enabled
   rust_backtrace = 1
+
+  # Set up default tags for otel traces via the OTEL_RESOURCE_ATTRIBUTES env variable. Format is key=value,key=value
+  # We're setting up defaults on a per-job basis, but these can be expanded on a per-service basis as necessary.
+  # Examples of keys we may add in the future: language, instance_id/ip, team
+
+  # Currently we use the same version for all containers. As such we pick one container to get the version from
+  app_version                      = split(":", var.container_images["analyzer-dispatcher"])[1]
+  default_otel_resource_attributes = "service.version=${local.app_version}"
 }
 
 job "grapl-core" {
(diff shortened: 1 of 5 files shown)


nomad/grapl-core.nomad
```diff
@@ -177,6 +177,14 @@ locals {
...
   # enabled
   rust_backtrace = 1
+
+  # Set up default tags for otel traces via the OTEL_RESOURCE_ATTRIBUTES env variable. Format is key=value,key=value
+  # We're setting up defaults on a per-job basis, but these can be expanded on a per-service basis as necessary.
+  # Examples of keys we may add in the future: language, instance_id/ip, team
+
+  # Currently we use the same version for all containers. As such we pick one container to get the version from
+  app_version                      = split(":", var.container_images["analyzer-dispatcher"])[1]
+  default_otel_resource_attributes = "service.version=${local.app_version}"
```
wimax-grapl, line 187: This _might_ be a better match for the `pulumi/infra/observability_env_vars.py` mechanism, which is where all the other OTEL_ stuff is.

I don't really foresee us using multiple tags concurrently but I could be wrong

twunderlich-grapl: So that's where I initially thought about putting this and then decided against it due to:

1. Right now we give all containers the same version tag. I suspect that at some point in the future, this might change because either a) we're using a third-party container for something or b) as we grow we only update the containers that need to be rebuilt based on code changes
2. I do see a future where multiple tags are used concurrently. The next tag that I'm likely to add soon is the instance hostname, which would help us track down problematic instances. That said, I think most other tags we could add will only be useful a few years down the line (ie tagging the team for a service is really only useful when you have a lot of teams. Really useful when you're past the 100+ engineer mark, but not that useful below the 20 engineer mark.

wimax-grapl (approved):

PR #2037, opened 2022-10-11
Scylla optimize provision

colin-grapl:
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/1031

Provisioning keyspaces and tables in parallel is slow and error prone. This changes things so that we provision things once and instead our tenancy isolation is done via partition keys, not keyspaces.

This PR also updates Scylla to 0.6, which includes a number of bug fixes, including changes around the reliability of ALTER commands.

Note that the tenant-id no longer needs to be communicated to the service but this PR does not remove it. There will be follow up issues to track this work.

Further, we attempt to only provision once by storing a boolean value and short circuiting if a provision is already successful. Provisioning should already be idempotent so this shouldn't be a problem, but it should help speed up tests.

Existing tests.

### DIFF
--- a/src/rust/Cargo.lock
+++ b/src/rust/Cargo.lock
@@ -4301,11 +4301,12 @@ dependencies = [
 
 [[package]]
 name = "scylla"
-version = "0.5.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eed40c02a5c835d894044bbd6f454ef54b6937ec8af23efe3cdeaa57005bb00d"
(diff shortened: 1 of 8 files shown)


colin-grapl:
> this is a fine start, but I'd prefer that we just call it from `grapl-provision` once instead of leave all these ScyllaProvisionerClients all over the test code. Mind if I just do that in a followup PR or something?

I've created this issue to track that. The short version is I'm not exactly sure at this point.
https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/1042

twunderlich-grapl:
We're currently running on scylla 4.6.3 locally. Should we upgrade that as part of this PR or in a follow-up PR?

PR #2038, opened 2022-10-11
Set up head-based sampling

ghost:

This sets up head-based sampling at the otel-collector layer. This allows us to set a percentage of traces to collect. Currently we're defaulting to 100% but we can decrease that as necessary. Long-term, we should split the otel-collectors into agents and a dedicated processing cluster. When we do that we should switch to tail-based sampling so that we can create rules around what to sample

### DIFF
--- a/pulumi/infra/observability_env_vars.py
+++ b/pulumi/infra/observability_env_vars.py
@@ -29,10 +29,10 @@ def get_observability_env_vars() -> str:
 def otel_config(
     lightstep_token: pulumi.Output,
     nomad_endpoint: pulumi.Output | str,
-    lightstep_endpoint: str = "ingest.lightstep.com:443",
+    lightstep_endpoint: str,
     # This is optional because pulumi.config.get_bool returns Optional[bool]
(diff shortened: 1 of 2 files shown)


wimax-grapl (approved):

jgrillo-grapl (approved):

pulumi/infra/observability_env_vars.py
```diff
@@ -32,12 +32,14 @@ def otel_config(
     lightstep_endpoint: str = "ingest.lightstep.com:443",
     # This is optional because pulumi.config.get_bool returns Optional[bool]
     lightstep_is_endpoint_insecure: bool | None = False,
+    trace_sampling_percentage: float | None = 100.0,
```
christophermaier, line 35: Can this be restructured so there's only one default value for this configuration variable in the codebase?

I've had bad experiences in the past playing "default value whack-a-mole" trying to figure out exactly where a given value is coming from, because every place it was used had a fallback.

christophermaier (changes requested):

christophermaier (approved):

PR #2043, opened 2022-10-12
Fe integration tests

andrea-grapl:
https://github.com/grapl-security/issue-tracker/issues/1012

Add initial integration tests for the frontend, set up in CI/CD. 

Using the new targets `make test-integration-frontend` `make test-unit-typescript`


src/js/frontend/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+WORKDIR /home/grapl
+ENTRYPOINT ["/bin/bash", "-o", "errexit", "-o", "nounset", "-c"]
+
+# integration-tests
+################################################################################
+FROM grapl-typescript-base AS integration-tests
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+# TODO: make sure yarn version matches package.json
+ENV YARN_VERSION 3.2.3
```
christophermaier, line 27: This can be done similar to what we do with [Python](https://github.com/grapl-security/grapl/blob/main/build-support/python_version.sh) and [Rust](https://github.com/grapl-security/grapl/blob/main/build-support/rust_version.sh); the core logic would be:
```shell
jq --raw-output '.packageManager' src/js/frontend/package.json |
    awk 'BEGIN { FS="@" } { print $2 }'
```
We'd then pass it in as a build argument; see:
- https://github.com/grapl-security/grapl/blob/46f05bfbcfa8b73143bf90e1eaa198a018b6aa22/Makefile#L19-L21
- https://github.com/grapl-security/grapl/blob/46f05bfbcfa8b73143bf90e1eaa198a018b6aa22/docker-bake.hcl#L70-L76

This could be part of this PR, or a follow-up PR.

src/js/frontend/Makefile
```diff
@@ -90,12 +90,10 @@ build-code: ## Build the code if any files have changed
...
-run-tests: build-code
-run-tests: ## Run all unit tests and gather coverage statistics
-	$(DOCKER_RUN) yarn test \
+.PHONY: run-unit-tests
+run-unit-tests: build-code
+run-unit-tests: ## Run all unit tests and gather coverage statistics
+	$(DOCKER_RUN) yarn unitTest \
 		--coverage \
 		--watchAll=false \
 		--coverageDirectory=/frontend/coverage/
```
christophermaier, line 99: We're going to have to do a bit more work to ensure that the coverage statistics actually get out of the container so we can upload them to Codecov.

We can follow-up on this after this PR, though.

src/js/frontend/README.md
```diff
@@ -15,7 +15,7 @@ Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
 The page will reload if you make edits.\
 You will also see any lint errors in the console.
 
-### `yarn test`
+### `yarn integrationTest` or `yarn unitTest`
```
christophermaier, line 18: I notice we seem to have a pattern of using `:` to create "subtypes" of a command (see `prettier` and `prettier:fix`; `lint` and `lint:fix`).

Any reason to not stick with that convention and do something like `test:integration` and `test:unit`?

src/js/frontend/build-env.Dockerfile
```diff
@@ -30,5 +30,9 @@ RUN mkdir --parents \
     /frontend/.yarn/state \
     /frontend/node_modules
 
+
+
+
+
```
christophermaier, line 36: May as well remove this change.

src/js/frontend/README.md
```diff
@@ -15,7 +15,7 @@ Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
...
 
 Launches the test runner in the interactive watch mode.\
```
christophermaier, line 20: Seems like this documentation paragraph is now out of date and needs to be updated.

src/js/frontend/src/components/tests/integrationTests/Login.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+        target: {
+          value: process.env.GRAPL_TEST_USER_NAME,
+        },
+      });
+    });
+
+    await waitFor(() => {
+      fireEvent.change(password, {
+        target: {
+          value: process.env.GRAPL_TEST_USER_PASSWORD_SECRET_ID, // this is not the password secret value yet,// we have to interact with AWS SDK to get the value from the secret ID
```
christophermaier, line 40: This value isn't passed into the tests in the Nomad job.

src/js/frontend/src/components/tests/integrationTests/Login.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+        },
+      });
+    });
+
+    await waitFor(() => {
+      fireEvent.click(submitButton);
+    });
+
+    expect(process.env.GRAPL_TEST_USER_NAME).toBe("local-grapl-grapl-test-user");
+    expect(process.env.GRAPL_TEST_PASSWORD).not.toBeNull();
```
christophermaier, line 50: Won't this assertion always fail? I don't think it is set anywhere.

src/js/frontend/src/components/tests/integrationTests/Login.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+          value: process.env.GRAPL_TEST_USER_PASSWORD_SECRET_ID, // this is not the password secret value yet,// we have to interact with AWS SDK to get the value from the secret ID
+        },
+      });
+    });
+
+    await waitFor(() => {
+      fireEvent.click(submitButton);
+    });
+
+    expect(process.env.GRAPL_TEST_USER_NAME).toBe("local-grapl-grapl-test-user");
```
christophermaier, line 49: This will work in "local grapl", but will fail in AWS, because the name depends on which environment things are being run in.

Is this still here just to prove that everything is wired up properly (from Pulumi to Nomad to your container to your tests), or can you actually make assertions against the actual login process? If it's just to prove things are wired up properly, I'd be inclined to set this as a separate test; after all, this isn't actually a test for "Retrieves password and successfully logs in", and having it pass because of this (and in the absence of anything else) would be very misleading.

src/js/frontend/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "strict": true,
     "forceConsistentCasingInFileNames": true,
     "noFallthroughCasesInSwitch": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
christophermaier, line 13: Do we really want to disable this?

src/rust/Cargo.lock
```diff
@@ -1965,12 +1965,12 @@ dependencies = [
 name = "grapl-tracing"
 version = "0.0.1"
 dependencies = [
- "opentelemetry 0.17.0",
```
christophermaier, line 1968: Why are Rust dependencies being updated in this PR?

christophermaier (changes requested):

pulumi/typescript_integration_tests/__main__.py
```diff
@@ -0,0 +1,106 @@
...
+            return cast(str, ref.require_output(key))
+
+        self.aws_env_vars_for_local = require_str("aws-env-vars-for-local")
+        self.test_user_name = require_str("test-user-name")
+
+        # self.plugin_work_queue_db = cast(
+        #     NomadServicePostgresDbArgs, ref.require_output("plugin-work-queue-db")
+        # )
+
+        # self.test_user_password_secret_id = require_str("test-user-password-secret-id")
```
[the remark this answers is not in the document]
christophermaier: Actually, once https://github.com/grapl-security/grapl/pull/2046 lands, you can pull that in and eliminate this `GraplStack` implementation altogether.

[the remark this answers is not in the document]
christophermaier: Update: #2046 has been merged.

src/js/frontend/tsconfig.json
andrea-grapl: Yes, there are some, but very few cases where we want the `any` type especially for plugins when we won't necessarily know the type before they're uploaded. I'm going to be very conscious about avoiding this in any situation possible, but we might need it in certain situations. We can switch it back for this PR, but we'll eventually need an any type,

docker-bake.hcl
```diff
@@ -221,6 +221,13 @@ group "local-infrastructure" {
   ]
 }
 
+group "typescript-integration-tests" {
+  # NOTE: Please keep this list sorted in alphabetical order
+  targets = [
+    "typescript-integration-tests",
```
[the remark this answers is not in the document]
andrea-grapl: Deleted both `typescript` and `python` groups

.buildkite/pipeline.verify.yml
```diff
@@ -229,6 +229,21 @@ steps:
...
+      - label: ":typescript: Typescript Integration Tests"
+        command:
+          - make test-integration-typescript
+        plugins:
+          - grapl-security/vault-login#v0.1.3
+          - grapl-security/vault-env#v0.2.0:
+              secrets:
+                - grapl/TOOLCHAIN_AUTH_TOKEN
+        env:
+          GRAPL_LOG_LEVEL: "DEBUG"
```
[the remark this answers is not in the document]
andrea-grapl: Removed

.buildkite/pipeline.provision.yml
```diff
@@ -90,9 +120,6 @@ steps:
     build:
       commit: "${BULIDKITE_COMMIT}"
       branch: "${BUILDKITE_BRANCH}"
-
-  - wait
-
```
christophermaier, line 95: This `wait` definitely needs to be present.

.buildkite/pipeline.testing.yml
```diff
@@ -20,5 +20,6 @@ steps:
             .buildkite/scripts/grapl_testing_stack/dump_artifacts.sh
     command:
       - .buildkite/scripts/grapl_testing_stack/run_parameterized_job.sh rust-integration-tests 6
+      - .buildkite/scripts/grapl_testing_stack/run_parameterized_job.sh typescript-integration-tests 6
```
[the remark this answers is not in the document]
christophermaier: This still should be in its own dedicated step.

(By "still run the tests sequentially", I meant "use two Buildkite steps, separated by a `wait` step"... I'm still concerned about the test artifacts with both commands running in the _same_ step.)

.buildkite/pipeline.verify.yml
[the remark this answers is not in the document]
christophermaier: I was referring to the `GRAPL_LOG_LEVEL` environment variable, not the `TOOLCHAIN_AUTH_TOKEN` secret. As currently written, this is an invalid Buildkite step definition because of the incomplete `vault-env` plugin specification.

nomad/typescript-integration-tests.nomad
```diff
@@ -0,0 +1,110 @@
...
+  (See DockerImageId in Pulumi for further documentation.)
+EOF
+}
+
+variable "test_user_name" {
+  type        = string
+  description = "The name of the test user"
+}
+
+#variable "test_user_password_secret_id" {
```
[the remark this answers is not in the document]
christophermaier: Still have some dead code here.

.buildkite/pipeline.testing.yml
```diff
@@ -20,6 +20,23 @@ steps:
             .buildkite/scripts/grapl_testing_stack/dump_artifacts.sh
     command:
       - .buildkite/scripts/grapl_testing_stack/run_parameterized_job.sh typescript-integration-tests 6
+    artifact_paths:
+      - "test_artifacts/**/*"
+
+  - wait
+
+  - label: ":aws: Integration tests in AWS"
```
christophermaier: I think you'll need to give the steps different names (e.g. "Rust Integration tests in AWS", "TypeScript Integration tests in AWS")

.buildkite/pipeline.verify.yml
```diff
@@ -235,8 +235,8 @@ steps:
         plugins:
           - grapl-security/vault-login#v0.1.3
           - grapl-security/vault-env#v0.2.0:
-        env:
-          GRAPL_LOG_LEVEL: "DEBUG"
+        secrets:
+          - grapl/TOOLCHAIN_AUTH_TOKEN
```
christophermaier: Watch the indentation here; the `secrets` object needs to be indented inside of the `vault-env` item.

pulumi/typescript_integration_tests/__main__.py
```diff
@@ -99,7 +99,7 @@ def require_str(key: str) -> str:
         #     NomadServicePostgresDbArgs, ref.require_output("plugin-work-queue-db")
         # )
 
-        # self.test_user_password_secret_id = require_str("test-user-password-secret-id")
+        self.test_user_password_secret_id = require_str("test-user-password-secret-id")
```
christophermaier, line 102: If you've incorporated changes from `main`, you don't need this entire `GraplStack` implementation; you can just pull in the common one, like we do in the `python_integration_tests` and `rust_integration_tests` Pulumi projects.

.buildkite/pipeline.verify.yml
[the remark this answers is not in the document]
andrea-grapl: Removed `LOG_LEVEL`, ADDED BACK `TOOLCHAIN_AUTH_TOKEN` secret

.python-version
```diff
@@ -1 +1 @@
-3.10.5
+3.10.5
```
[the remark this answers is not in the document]
andrea-grapl: Checked out and committed file from Main

src/rust/Cargo.lock
andrea-grapl: Removed

.buildkite/pipeline.provision.yml
andrea-grapl: Added back

pulumi/typescript_integration_tests/__main__.py
```diff
@@ -0,0 +1,158 @@
...
+#             return cast(str, ref.require_output(key))
+#
+#         self.aws_env_vars_for_local = require_str("aws-env-vars-for-local")
+#         self.test_user_name = require_str("test-user-name")
+#
+#         # self.plugin_work_queue_db = cast(
+#         #     NomadServicePostgresDbArgs, ref.require_output("plugin-work-queue-db")
+#         # )
+#
+#         self.test_user_password_secret_id = require_str("test-user-password-secret-id")
```
christophermaier, line 103: Remove this commented out code entirely.

pulumi/typescript_integration_tests/__main__.py
```diff
@@ -0,0 +1,158 @@
...
+
+def main() -> None:
```
christophermaier, line 105: You have _two_ `main` definitions now!

src/js/frontend/src/components/tests/integrationTests/Login.test.tsx
andrea-grapl: resolved

src/js/frontend/src/services/test/modules/getSecretValues.ts
```diff
@@ -0,0 +1,12 @@
+import { SecretsManagerClient, GetSecretValueCommand } from "@aws-sdk/client-secrets-manager"; // ES Modules import
+import getAwsClient from "./envHelpers";
+
+const getSecrets = async () => {
```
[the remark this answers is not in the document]
andrea-grapl: This is how syntax for declaring functions using ES6 in Typescript.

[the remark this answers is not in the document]
christophermaier: How is this different from the declaration of `getAwsClient` in `envHelpers.ts`? That is declared with a `function` keyword.

src/js/frontend/src/components/login/LoginForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { yupValidationSchema } from "./yupValidationSchema";
+
+const LoginForm = () => {
+  const [state, setState] = React.useState({
+    loginFailed: false, // Boolean represented as true when user is successfully authenticated,
+    // false when there's an auth error, token has been removed or user has been logged out
+  });
+
+  const handleSubmit = async (values: { username: string; password: string }) => {
+    const loginSuccess = await loginService(values.username, values.password);
```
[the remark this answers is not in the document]
andrea-grapl: Can we meet in the middle with `successfulLoginResponse`? Because it's a boolean, I think it makes sense to name the variable based on the state of what's true and what's false.`loginResult` set to a bool is slightly confusing because it can return an object with login values or an error, but I'm happy to change it to `loginResult` if unclear.

src/js/frontend/src/components/tests/integrationTests/Login.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+import getSecrets from "../../../services/test/modules/getSecretValues";
+
+// to show the result of render(), use screen.debug() which displays HTML
+describe("Login Component", () => {
+  afterEach(cleanup);
+
+  test("Retrieves password from AWS and executes password in login form ", async () => {
+    act(() => {
+      render(
+        <GoogleOAuthProvider clientId="340240241744-6mu4h5i6h9j7ntp45p3aki81lqd4gc8t.apps.googleusercontent.com">
```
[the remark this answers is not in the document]
andrea-grapl: Per Ian, there's nothing secret about this value, but it the client ID is which domain we're logging into. Ian set up SSO so he would be a good starting point.

src/js/frontend/src/services/test/modules/getSecretValues.ts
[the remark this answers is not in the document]
andrea-grapl: Removed function keyword so all functions use ES6 now.

src/js/frontend/src/components/tests/integrationTests/Login.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    await waitFor(() => {
+      fireEvent.click(submitButton);
+    });
+
+    await waitFor(() => {
+      expect(username).not.toBeNull();
+      expect(getCreds).not.toBeNull();
+      expect(getCreds).toBeDefined();
+      expect(getCreds).toHaveProperty("SecretString");
```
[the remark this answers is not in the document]
andrea-grapl: Updated tests to test onsubmit of form with creds from the environment

src/js/frontend/src/components/login/LoginForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { Field, Form, Formik } from "formik";
+import { GoogleSSO } from "./GoogleSSO";
+
+import { loginService } from "../../services/login/loginService";
+import { yupValidationSchema } from "./yupValidationSchema";
+
+const LoginForm = () => {
+  const [state, setState] = React.useState({
+    loginFailed: false, // Boolean represented as true when user is successfully authenticated,
+    // false when there's an auth error, token has been removed or user has been logged out
```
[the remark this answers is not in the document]
andrea-grapl: Renamed to `loginStatus` where true is logged in and false is not logged in. We use `loggedIn` as a variable elsewhere so this seemed like an okay middle ground?

src/js/frontend/src/components/login/LoginForm.tsx
```diff
@@ -0,0 +1,75 @@
...
+              name="password"
+              type="password"
+              placeholder="Password"
+            />{" "}
+            <br />
+            {touched.password && errors.password && <div>{errors.password}</div>}
+            <button data-testid={"button"} name="submitButton" type="submit">
+              SUBMIT
+            </button>
+            {state.loginStatus && <div data-testid={"loginError"}>Unsuccessful Login</div>}
```
christophermaier, line 67: Won't this display "Unsuccessful Login" only when the user successfully logged in?

src/js/frontend/src/components/login/LoginForm.tsx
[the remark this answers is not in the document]
christophermaier: Reading more of the code, I think that my issue ultimately lies in the fact that our `loginService` function is returning a boolean, rather than the response itself. Because of that, we now have a lot of booleans in this scope that sound like they're the same thing, but actually aren't, which is really confusing.

If that function just returned the request, then _it_ would be simpler (it's just making a request, not making a request and figuring out what to do with it), and the code _here_ would be more clear:

```typescript
const loginResponse = await loginService(values.username, values.password);

if loginResponse["success"] {
    // do successful things
} else {
    // do failure things, including operating on any error messages that may be in the response!
    loginStatus = false;
}
```
To my eye, this reads much more clearly, and there is no ambiguity. Additionally, this formulation is _more useful_, because any error messages in a failed login response are not swallowed by the `loginService` code.

(Incidentally, `loginService.tsx` is identical to `signInWithGoogleService.ts`... I assume the latter should be deleted?)

src/js/frontend/src/components/tests/integrationTests/Login.test.tsx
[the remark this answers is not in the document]
christophermaier: It's still worth extracting this as a constant and documenting this kind of information. Better to be able to find out the information by reading the code than having to ask Ian anytime someone has a question about it.

src/js/frontend/src/components/tests/integrationTests/Login.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+          <LoginForm />
+        </GoogleOAuthProvider>,
+      );
+    });
+
+    const username = screen.getByPlaceholderText(/Username/i);
+    const password = screen.getByPlaceholderText(/Password/i);
+    const submitButton = screen.getByText("SUBMIT");
+
+    let getCreds = await getSecrets;
```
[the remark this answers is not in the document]
andrea-grapl: Thank you for the detail and clarity, I appreciate how much thought you put into this and have updated variables and function calls to be more clear. Thank you!!!

src/js/frontend/src/components/login/LoginForm.tsx
andrea-grapl: Resolved

src/js/frontend/src/components/tests/integrationTests/Login.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+      expect(testPassword).not.toBeNull();
+      expect(testPassword).toBeDefined();
+
+      expect(onSubmit).toHaveBeenCalledWith({
+        username: process.env.GRAPL_TEST_USER_NAME,
+        password: testPassword,
+      });
+    });
+
+    expect(onSubmit).toHaveBeenCalledTimes(1);
```
christophermaier, line 65: I think you need another assertion in this test. As it is, it looks like this just asserts that the user filled in some values in the login form and then hit the button; we need an assertion that the login was actually _successful_.

PR #2057, opened 2022-10-17
Script to purge dead grapl-plugin jobs from Nomad

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/1044

Script to clean up dead grapl-plugin jobs from Nomad.
We're seeing a number of dead grapl-plugin jobs, all in different namespaces. Even though these jobs are dead, they're not cleaned up automatically. As such, we need a script to clean these up

Ran the script

### DIFF
--- a/bin/nomad_plugin_purge.sh
+++ b/bin/nomad_plugin_purge.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+# This script is meant to be used to clear dead copies of the grapl-plugin job in Nomad.
+
+# Usage:
+# Make sure there is an open tunnel to Nomad (typically via ./bin/aws/ssm_nomad_server.sh in a separate tab)
+# Then run this script ./bin/nomad_plugin_purge.sh
+
+plugin_namespaces=$(nomad job stop -namespace=* -purge grapl-plugin 2>&1 | tail -n +3 | awk '{ if ($5 == "dead") { print $2 } }')
+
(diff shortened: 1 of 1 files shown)


bin/nomad_plugin_purge.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+# This script is meant to be used to clear dead copies of the grapl-plugin job in Nomad.
+
+# Usage:
+# Make sure there is an open tunnel to Nomad (typically via ./bin/aws/ssm_nomad_server.sh in a separate tab)
+# Then run this script ./bin/nomad_plugin_purge.sh
+
+plugin_namespaces=$(nomad job stop -namespace=* -purge grapl-plugin 2>&1 | tail -n +3 | awk '{ if ($5 == "dead") { print $2 } }')
```
christophermaier, line 9: Does the bare `*` cause shell globbing issues? Should it be single-quoted?

Might also be good to add a typical example of the expected output so the `tail` and `awk` commands make a bit more sense; as it is, I have no idea how to evaluate that pipeline.

twunderlich-grapl: Updated. Thanks for the suggestions, very useful especially to our future selves when hashicorp changes the return

christophermaier (approved):
Looks great; thanks for the changes! 🙇

PR #2060, opened 2022-10-17
Fix s3 uri format so that its explicitly a s3 URI

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/1045

This updates the s3 uri used with the Nomad artifact stanza in order to fix the 403 error we were seeing in AWS. This appears to be an issue with go-getter and http://, and can be seen in this thread: https://github.com/hashicorp/nomad/issues/1113
An alternate fix would be to remove the prefix entirely (ie get rid of s3:: or the previous http://)

I've removed the comments about the slack thread, since that thread has been deleted (past the 90 day threshold or whatever it is)

1. On a nomad agent instance I downloaded the internally used go-getter binary and tested the former url and the new url:
```
sh-4.2$ ./go-getter http://plugin-registry-bucket-b1abfd2.s3.amazonaws.com/plugins/tenant_id_a7b7e3a8-fdcf-4032-a0e8-d6e3aa1973d4/plugin_type-generator/3ee133b7-8147-41bb-8908-283e0d946197.bin .
2022/10/17 23:27:21 Error downloading: bad response code: 403
sh-4.2$ ./go-getter s3::http://plugin-registry-bucket-b1abfd2.s3.amazonaws.com/plugins/tenant_id_a7b7e3a8-fdcf-4032-a0e8-d6e3aa1973d4/plugin_type-generator/3ee133b7-8147-41bb-8908-283e0d946197.bin .
2022/10/17 23:27:48 Error downloading: URL is not a valid S3 URL
sh-4.2$ ./go-getter s3::plugin-registry-bucket-b1abfd2.s3.amazonaws.com/plugins/tenant_id_a7b7e3a8-fdcf-4032-a0e8-d6e3aa1973d4/plugin_type-generator/3ee133b7-8147-41bb-8908-283e0d946197.bin .2022/10/17 23:28:52 success!
```
2. I've manually updated the Nomad job with the new uri and seen that the plugin tasks starts running successfully

### DIFF
--- a/src/rust/plugin-registry/src/server/s3_uri.rs
+++ b/src/rust/plugin-registry/src/server/s3_uri.rs
@@ -2,7 +2,7 @@ use grapl_config::env_helpers::ENV_ENDPOINT;
 
 /// This gets a s3 URI to be used by Nomad's artifact stanza (in particular the grapl-plugin).
 /// The underlying library, Hashicorp's [go-getter](https://github.com/hashicorp/go-getter) lib,
-/// doesn't work with the `http://` schema prefix, but does work with an explicit s3 prefix `s://`
+/// doesn't work with the `http://` schema prefix, but does work with an explicit s3 prefix `s3://`
 /// or without a prefix. We've opted for the s3 prefix to be explicit.
 pub fn get_s3_uri(bucket: &str, key: &str) -> String {
     let local_endpoint = std::env::var(ENV_ENDPOINT).ok();
--- a/src/rust/plugin-registry/src/server/s3_uri.rs
+++ b/src/rust/plugin-registry/src/server/s3_uri.rs
@@ -1,14 +1,19 @@
 use grapl_config::env_helpers::ENV_ENDPOINT;
 
 /// This gets a s3 URI to be used by Nomad's artifact stanza (in particular the grapl-plugin).
-/// The underlying go-getter lib, doesn't work with the `http://` schema prefix, but does work with
-/// an explicit s3 prefix `s::` or without a prefix. We've opted for the s3 prefix to be explicit.
+/// The underlying library, Hashicorp's [go-getter](https://github.com/hashicorp/go-getter) lib,
+/// doesn't work with the `http://` schema prefix, but does work with an explicit s3 prefix `s://`
+/// or without a prefix. We've opted for the s3 prefix to be explicit.
 pub fn get_s3_uri(bucket: &str, key: &str) -> String {
     let local_endpoint = std::env::var(ENV_ENDPOINT).ok();
     // If the above is specified, we're running locally against Localhost
-    // Otherwise, it's against prod S3; so we can use virtual-hosted-style URLs
+    // Otherwise, it's against prod S3; so we use the s3 URIs. Notably, the underlying go-getter
+    // lib does not seem to work with s3 object URLs (urls that start with http:// or https://),
+    // including the recommended virtual host style access.
+    // https://github.com/hashicorp/go-getter/issues/387 tracks updating the documentation.
+    // This is the s3 URI format used in the s3 console.
     local_endpoint.map_or_else(
-        || format!("s3::{bucket}.s3.amazonaws.com/{key}"),
+        || format!("s3://{bucket}.s3.amazonaws.com/{key}"),
         |endpoint| format!("{endpoint}/{bucket}/{key}"),
     )
 }
--- a/src/rust/plugin-registry/src/server/s3_uri.rs
+++ b/src/rust/plugin-registry/src/server/s3_uri.rs
@@ -1,6 +1,9 @@
 use grapl_config::env_helpers::ENV_ENDPOINT;
 
-pub fn get_s3_url(bucket: &str, key: &str) -> String {
+/// This gets a s3 URI to be used by Nomad's artifact stanza (in particular the grapl-plugin).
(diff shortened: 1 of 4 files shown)


src/rust/plugin-registry/src/server/s3_url.rs
```diff
@@ -1,13 +1,11 @@
 use grapl_config::env_helpers::ENV_ENDPOINT;
 
-/// Some discussion about this function here:
-/// https://grapl-internal.slack.com/archives/C02J5JYS92S/p1648667074625629
 pub fn get_s3_url(bucket: &str, key: &str) -> String {
     let local_endpoint = std::env::var(ENV_ENDPOINT).ok();
     // If the above is specified, we're running locally against Localhost
     // Otherwise, it's against prod S3; so we can use virtual-hosted-style URLs
```
christophermaier, line 6: This comment looks like it needs updating now.

src/rust/plugin-registry/src/server/s3_url.rs
```diff
@@ -1,13 +1,11 @@
...
     local_endpoint.map_or_else(
-        || format!("http://{bucket}.s3.amazonaws.com/{key}"),
+        || format!("s3::{bucket}.s3.amazonaws.com/{key}"),
```
christophermaier, line 8: How is this a URL? I would buy `s3://...` but this just looks _weird_.

twunderlich-grapl: Ah let me change the function name to URI. And agreed that this is a very weird schema. It feels like someone was like, well our arns already use `::` as separators, why not just reuse that?

src/rust/plugin-registry/src/server/s3_url.rs
twunderlich-grapl: Updated. Unfortunately, since I can't access the slack message, the comment I've added may be incomplete but 🤷 .
Let me know if there's something you think I should add.

christophermaier: I was mainly referring to the fact that it's not an S3 virtual-host-style URL anymore (and not even a URL, or even a URI, I think :shrug:)

If it's all about how `go-getter` handles things, then a mention of that and any relevant links would be useful.

Without that kind of background information, I could _easily_ see someone in the future "helpfully" refactoring this into something that doesn't work anymore.

src/rust/plugin-registry/src/server/s3_uri.rs
```diff
@@ -1,6 +1,9 @@
 use grapl_config::env_helpers::ENV_ENDPOINT;
 
-pub fn get_s3_url(bucket: &str, key: &str) -> String {
+/// This gets a s3 URI to be used by Nomad's artifact stanza (in particular the grapl-plugin).
+/// The underlying go-getter lib, doesn't work with the `http://` schema prefix, but does work with
+/// an explicit s3 prefix `s::` or without a prefix. We've opted for the s3 prefix to be explicit.
```
christophermaier, line 5: `s::` -> `s3::`

src/rust/plugin-registry/src/server/s3_url.rs
christophermaier: Out of curiosity, would it have worked with `https://` instead of `http://`?

twunderlich-grapl: Sadly no. I tried that early on 😿

twunderlich-grapl: So apparently `s3://` does work and is officially supported by AWS. I've switched to that, especially since this is the format the s3 URI link in the console is (https://s3.console.aws.amazon.com/s3/object/plugin-registry-bucket-b1abfd2?region=us-east-1&prefix=plugins/tenant_id_024885b9-dcf0-469d-b12a-81876cc045c8/plugin_type-analyzer/51f8b573-6b7c-411c-9379-2b1070205c1f.bin and search s3 URI).
Apparently, the `s3::` was something the go-getter lib created and call [Forced Protocols](https://github.com/hashicorp/go-getter#forced-protocol). Basically its a way of saying that this is a s3 url.

src/rust/plugin-registry/src/server/s3_url.rs
twunderlich-grapl: Updated docs again. This should be better

wimax-grapl (approved):
i think i'd prever a `grapl_config::env_helpers::is_local_aws()` but I know people have Very Strong Opinions on resurrecting `IS_LOCAL`. In this case, it makes sense to me.

src/rust/plugin-registry/src/server/s3_uri.rs
```diff
@@ -0,0 +1,19 @@
+use grapl_config::env_helpers::ENV_ENDPOINT;
+
+/// This gets a s3 URI to be used by Nomad's artifact stanza (in particular the grapl-plugin).
+/// The underlying library, Hashicorp's [go-getter](https://github.com/hashicorp/go-getter) lib,
+/// doesn't work with the `http://` schema prefix, but does work with an explicit s3 prefix `s://`
```
christophermaier, line 5: `s://` -> `s3://`

christophermaier (approved):
Tiny documentation tweak, but otherwise 👍 

Thanks for digging into the guts of this issue 🎉

PR #2062, opened 2022-10-18
Bump Tonic and Prost. This now requires installing protoc in Docker images!

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/1047

- The latest Tonic/Prost no longer builds `protoc` from source - you have to supply/bundle it. I've added it in three contexts:
1. `setup_chromeos` (for people running `cargo` locally on their boxes)
2. `rust/build-env.Dockerfile` (for `make lint-rust` and `make test-unit-rust`)
3. `rust/Dockerfile` (for real building)

lastly, I've included the rest of the necessary plumbing to have a single place (`.protoc-version`) to specify version.

Passed integration tests on devbox and on CI.


src/rust/Dockerfile
```diff
@@ -35,14 +36,32 @@ RUN --mount=type=cache,target=/var/lib/apt/lists,sharing=locked,id=rust-base-apt
...
 EOF
 
+WORKDIR /tmp
+RUN <<EOF
+    echo "${PROTOC_VERSION}"
+    PB_REL="https://github.com/protocolbuffers/protobuf/releases"
+    ZIP_PATH="/tmp/protoc.zip"
+
+    # Download the zip
+    curl \
```
colin-grapl, line 54: Same thing - check how we invoke curl elsewhere.

src/rust/build-env.Dockerfile
```diff
@@ -37,6 +40,23 @@ rm "${ZIP_NAME}"
 mv /nomad/nomad /bin
 EOF
 
+WORKDIR /tmp
+RUN <<EOF
+    PB_REL="https://github.com/protocolbuffers/protobuf/releases"
+    ZIP_PATH="/tmp/protoc.zip"
+
+    # Download the zip
+    curl \
```
colin-grapl, line 49: curl args

colin-grapl (approved):

.github/CODEOWNERS
```diff
@@ -1,4 +1,5 @@
 .python-version @christophermaier
+.protoc-version @christophermaier
```
christophermaier, line 2: I see I have been volunteered :joy:

build-support/protoc_version.sh
```diff
@@ -0,0 +1,15 @@
...
+# this information (e.g., build arguments for container images to
+# ensure base images are in-sync with our Rust version)
+#
+# It can safely be invoked from anywhere in the repository, as it
+# resolves the .protoc-version file from the repository root.
+set -euo pipefail
+
+repo_root="$(git rev-parse --show-toplevel 2> /dev/null)"
+readonly repo_root
+cat "${repo_root}/.protoc-version"
```
christophermaier, line 15: I like the centralization of this logic, but in this particular case, I think that as long as we have Python code that is being generated from the Protobuf definitions, we should set the version in `pants.toml`, rather than `.protoc-version`.

See https://www.pantsbuild.org/docs/reference-protoc for more details.

We can still use this script to ensure that the value can get into our Docker images, but it would need to do some TOML-munging. I'm not aware of any shell utilities that deal with TOML, but we could potentially get away with some `grep`, or use a tiny Python script. I'm currently chatting with the Pants folks to see if there's some kind of introspection ability in Pants already.

PR #2065, opened 2022-10-18
Kafka producers compress messages with `zstd`

christophermaier:
Due to an apparent bug in the Confluent Cloud Terraform provider, it's not possible to create _topics_ with a `compression.type` setting of anything other than `producer`. This essentially means that a topic will honor whatever scheme a received message is compressed with, and won't re-compress it according to its own compression scheme.

As a result, we'll want to ensure that our messages are `zstd`-compressed when they enter our pipeline; if all our topics are configured with `compression.type` set to `producer`, then the `zstd` compression will be preserved throughout the pipeline.

In retrospect, it's actually better this way; now, we're explicitly controlling the compression at the head of the pipeline.

(An upcoming refresh of our Confluent Cloud infrastructure will ensure our topics are all set up with the `provider` compression setting, but even without that change, *this* change should be effectively transparent, as far as Grapl is concerned.)

### DIFF
--- a/pulumi/infra/kafka.py
+++ b/pulumi/infra/kafka.py
@@ -144,7 +144,7 @@ class Kafka(pulumi.ComponentResource):
                         partitions=2,
                         replication_factor=3,
                         config={
-                            "compression.type": "zstd",
+                            "compression.type": "producer",
                             "min.insync.replicas": 2,
                         },
                         opts=pulumi.ResourceOptions(provider=provider),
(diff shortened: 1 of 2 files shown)


pulumi/infra/kafka.py
```diff
@@ -144,7 +144,7 @@ def __init__(
                         partitions=2,
                         replication_factor=3,
                         config={
-                            "compression.type": "zstd",
+                            "compression.type": "provider",
```
jgrillo-grapl, line 147: this should be `"producer"`

jgrillo-grapl (approved):
One small change, otherwise LGTM

pulumi/infra/kafka.py
christophermaier: Oh good lord... 🤦 

`E_TOOMUCHPULUMI`

jgrillo-grapl: LGTM!

jgrillo-grapl (approved):

ghost (approved):

PR #2066, opened 2022-10-19
Add retries to flaky web integration tests.

inickles-grapl:

We're seeing intermittent 500 errors returned from the Consul sidecar in the grapl-web-ui integration tests.

These often occur after a test has already successfully made requests to grapl-web-ui, suggesting that service has had enough time to come up as healthy.

https://github.com/grapl-security/issue-tracker/issues/1008

This adds retry logic to the grapl-web-ui integration tests. Wen a request returns a 500 error, the request will retry up to ten times, waiting a second between each retry attempt.

CI

### DIFF
--- a/src/rust/grapl-web-ui/tests/api/auth/check_login.rs
+++ b/src/rust/grapl-web-ui/tests/api/auth/check_login.rs
@@ -2,7 +2,8 @@
 async fn auth_unauthenticated_check_login() -> eyre::Result<()> {
     let app = crate::test_app::TestApp::init().await?;
 
-    let response = app.post("api/auth/checkLogin").send().await?;
+    let request = app.post("api/auth/checkLogin");
+    let response = app.send_with_retries(request).await?;
 
     eyre::ensure!(
         response.status() == actix_web::http::StatusCode::UNAUTHORIZED,
@@ -19,7 +20,8 @@ async fn auth_authenticated_check_login() -> eyre::Result<()> {
 
     app.login_with_test_user().await?;
 
-    let response = app.post("api/auth/checkLogin").send().await?;
+    let request = app.post("api/auth/checkLogin");
+    let response = app.send_with_retries(request).await?;
 
     eyre::ensure!(
         response.status() == actix_web::http::StatusCode::OK,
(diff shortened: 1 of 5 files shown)


jgrillo-grapl (changes requested):
If I understand the slack conversations that precipitated this PR correctly, the flakiness we're observing is the result of a timing issue where these tests are running before grapl-web-ui is ready to serve requests.

If that's the case, I'd recommend fixing the flakiness temporarily by adding a big healthy sleep to the start of each test function, and prioritizing a longer term project to make sure grapl-web-ui doesn't report healthy until it's able to talk to its backend services--that is, until all its dependencies themselves report healthy.

Once grapl-web-ui has robust healthchecks, I'd recommend in each test function to block on grapl-web-ui reporting healthy, then run the test. Until such a time, a simple sleep should do the job.

Unless I'm totally misunderstanding the reason for this PR, which is entirely possible...

inickles-grapl:
> If I understand the slack conversations that precipitated this PR correctly, the flakiness we're observing is the result of a timing issue where these tests are running before grapl-web-ui is ready to serve requests.

That's not it, a number of the failures we've seen are in a call to `api/plugin/get_metadata`, which uses data in the response from a previous and successful call to `api/plugin/create`.

Notes and logs can be found here: https://github.com/grapl-security/issue-tracker/issues/1008

> If that's the case, I'd recommend fixing the flakiness temporarily by adding a big healthy sleep to the start of each test function, and prioritizing a longer term project to make sure grapl-web-ui doesn't report healthy until it's able to talk to its backend services--that is, until all its dependencies themselves report healthy.
> 
> Once grapl-web-ui has robust healthchecks, I'd recommend in each test function to block on grapl-web-ui reporting healthy, then run the test. Until such a time, a simple sleep should do the job.
> 
> Unless I'm totally misunderstanding the reason for this PR, which is entirely possible...

We already have a sleep for just this reason: https://github.com/grapl-security/grapl/blob/main/src/rust/Dockerfile#L263-L292.

Because a number of the failures we're seeing happen after a test has made successful requests already it suggests it's not due to this initial timeout being too small.

jgrillo-grapl (approved):

twunderlich-grapl:
Apparently, the next version of Consul (1.14) will add support for setting up retry policies on Envoy per https://github.com/hashicorp/consul/pull/12890. Once we're on that version, I'd like to set that up and potentially back out of this

PR #2067, opened 2022-10-19
Read protoc-version from pants.toml instead of .protoc-version

wimax-grapl:
Making two changes requested by cm in followup comments to https://github.com/grapl-security/grapl/pull/2062
- the pants.toml protoc-version
- the codeowners fix


pants.toml
```diff
@@ -209,6 +209,9 @@ lockfile = "3rdparty/python/pyupgrade.lockfile"
 # For keep-runtime-typing to work we also can't use py30-plus
 args = ["--py39-plus", "--keep-runtime-typing"]
 
+[protoc]
+version = "3.20.1"
```
christophermaier, line 213: It would be worth adding a comment here describing why we currently need to specify the version for `protoc`, even if it is the default value in Pants (which `3.20.1` currently is). We don't want somebody "helpfully" refactoring this entire stanza away.

Pants does provide a massive JSON dump of information that we can extract this information out of (even without an explicit configuration in `pants.toml`. The format is a little complex and I'm still working through it, but that seems like the ultimate way to approach this. In the meantime, of course, `grep` is perfectly fine; I really just care about a comment here at this stage. 😄

christophermaier (approved):
Good to go once we have an explanatory comment!

Thanks for revisiting this 🙇

PR #2078, opened 2022-10-25
Internal 1014 nightmare refactor from hell

jgrillo-grapl:

https://app.zenhub.com/workspaces/grapl-6036cbd36bacff000ef314f2/issues/grapl-security/issue-tracker/1014

This is a large PR. It removes the various gRPC client factory macros and replaces them with a generic blanket implementation.

The motivation for this work was to make _individual gRPC client instances_ individually configurable. Previously, because of the macro, we only had the ability to configure _every_ client instance the same way. Now we can tailor configurations individually at the use site upon client instance creation.

This brings in [figment](https://docs.rs/figment/latest/figment/) for client configurations. Moving forward, we should transition our clap-based configurations to figment, but for now I've just focused on the client configs. We can move the other configurations over later.

Automated tests.


src/rust/analyzer-dispatcher/src/main.rs
```diff
@@ -98,11 +94,14 @@ impl AnalyzerDispatcherConfig {
...
+    #[error("error configuring kafka client {0}")]
     KafkaConfiguration(#[from] KafkaConfigurationError),
 
-    #[error("error configuring analyzer IDs cache")]
-    AnalyzerIdsCacheConfiguration(#[from] ConnectError),
+    #[error("error configuring gRPC client {0}")]
+    ClientError(#[from] ClientError),
+
+    #[error("figment error {0}")]
+    FigmentError(#[from] figment::Error),
```
wimax-grapl, line 104: i'd argue that a figment configuration error is a valid scenario for a panic - that's never recoverable - but this is fine

src/rust/analyzer-dispatcher/src/main.rs
```diff
@@ -135,15 +134,21 @@ struct AnalyzerDispatcher {
...
         let graph_update_retry_producer: RetryProducer<Update> =
             RetryProducer::new(config.kafka_retry_producer_config)?;
-        let client_config = PluginRegistryClientConfig::parse();
 
-        let plugin_registry_client =
-            PluginRegistryServiceClient::connect_with_config(client_config).await?;
+        let client_config = Figment::new()
+            .merge(Env::prefixed("PLUGIN_REGISTRY_CLIENT_"))
+            .extract()?;
+        let plugin_registry_client = PluginRegistryClient::connect_with_healthcheck(
```
wimax-grapl, line 146: Colin very adamantly pushed against our having healthchecks in our code - this is a discussion best hashed out with him. (I'm in favor of the health checks, personally.)

https://grapl-internal.slack.com/archives/C03JNJ57A9G/p1658862938530339

this is from before we introduced client-executor's retries stuff

src/rust/Cargo.toml
```diff
@@ -49,11 +49,13 @@ clap = { version = "3.0", default_features = false, features = [
 ] }
 env_logger = "0.9.0"
 eyre = "0.6"
+figment = { version = "0.10", features = ["env"] }
```
wimax-grapl, line 52: thanks for workspace-deps'ing :)

src/rust/e2e-tests/src/test_utils/context.rs
```diff
@@ -65,36 +62,71 @@ impl AsyncTestContext for E2eTestContext {
...
-            GraphSchemaManagerClient::connect_with_config(GraphSchemaManagerClientConfig::parse())
-                .await
-                .expect("graph_schema_manager_client");
-
-        let pipeline_ingress_client =
-            PipelineIngressClient::connect_with_config(PipelineIngressClientConfig::parse())
-                .await
-                .expect("pipeline_ingress_client");
+        let event_source_client_config = Figment::new()
+            .merge(Env::prefixed("EVENT_SOURCE_CLIENT_"))
```
wimax-grapl, line 66: I sort of don't love that every client of a given service has to re-state what its prefix is, but it's not a big deal. Nonetheless the Env::prefixed stuff is killer very cool

src/rust/graph-merger/src/service.rs
```diff
@@ -58,8 +56,8 @@ pub enum GraphMergerError {
     #[error("failed to configure tracing {0}")]
     SetupTracingError(#[from] SetupTracingError),
 
-    #[error("GraphMutationClientError {0}")]
-    GraphMutationClientError(#[from] GraphMutationClientError),
+    #[error("gRPC client error {0}")]
+    GrpcClientError(#[from] ClientError),
```
wimax-grapl, line 60: In this PR I see both `GrpcClientError(#[from] ClientError)` and `ClientError(#[from] ClientError)`, could be worth standardizing (but not a big deal)

src/rust/plugin-execution-sidecar/src/sidecar_client/generator_client.rs
```diff
@@ -1,26 +0,0 @@
...
-        ConnectWithConfig,
-    },
-};
-
-fn get_plugin_upstream_address(plugin_id: uuid::Uuid) -> String {
-    let upstream_addr_env_var = format!("NOMAD_UPSTREAM_ADDR_plugin-{plugin_id}");
-    let upstream_addr = std::env::var(&upstream_addr_env_var).expect(&upstream_addr_env_var);
-    let address = format!("http://{upstream_addr}");
-    address
-}
```
wimax-grapl, line 15: I'm afraid I don't see where this behavior was replaced. Plugin upstreams cannot really be done via just-env-vars (well, they can: see the solution to https://github.com/hashicorp/nomad/issues/14813 )

src/rust/plugin-execution-sidecar/src/work/analyzer_work_processor.rs
```diff
@@ -45,27 +50,23 @@ impl Workload for GetExecuteAnalyzerResponse {
...
-    analyzer_service_client: AnalyzerServiceClient,
+    analyzer_client: AnalyzerClient,
 }
 
 impl AnalyzerWorkProcessor {
-    pub async fn new(config: &PluginExecutorConfig) -> Result<Self, Box<dyn std::error::Error>> {
-        let plugin_id = config.plugin_id;
-        let upstream_addr_env_var = format!("NOMAD_UPSTREAM_ADDR_plugin-{plugin_id}");
-        let upstream_addr = std::env::var(&upstream_addr_env_var).expect(&upstream_addr_env_var);
-        let address = format!("http://{upstream_addr}");
```
wimax-grapl, line 56: I don't see any `.nomad` changes in this PR, so I very much doubt that this ANALYZER_CLIENT_ env prefix stuff works, same as the previous comment about the generator plugin upstream.

wimax-grapl: (taking a PR checkpoint here since it's serious enough to send back for discussion)

src/rust/rust-proto/src/graplinc/grapl/api/client.rs
```diff
@@ -0,0 +1,381 @@
...
+    ) -> Result<NU, ClientError>
+    where
+        PT: prost::Message + From<NT> + Clone,
+        NT: ProtobufSerializable<ProtobufMessage = PT> + TryFrom<PT>,
+        PU: prost::Message,
+        NU: ProtobufSerializable<ProtobufMessage = PU> + TryFrom<PU>,
+        P: Fn(&tonic::Status) -> bool,
+        F: FnMut(C, tonic::Request<PT>) -> R + Clone,
+        R: Future<Output = Result<tonic::Response<PU>, tonic::Status>>,
+        ClientError: From<<NU as TryFrom<PU>>::Error>,
```
wimax-grapl, line 316: It's a little hard for me to reason about the names of these type vars; would you be amenable to e.g.

ProtoRequest, NativeRequest, ProtoResponse, NativeResponse 
ProtoReq, NativeResp
or even PReq, NResp? 

If you'd like to keep them short, that's fine, but could you leave a comment about what each of the [P,N][T,U] represents? 

i'm assuming PT/PU is just from T, U in the alphabet?

src/rust/rust-proto/src/graplinc/grapl/api/event_source/v1beta1/client.rs
```diff
@@ -1,102 +1,83 @@
...
+    protobufs::graplinc::grapl::api::event_source::v1beta1::event_source_service_client::EventSourceServiceClient,
 };
 
-pub type EventSourceServiceClientError = GrpcClientError;
+#[async_trait::async_trait]
+impl Connectable for EventSourceServiceClient<tonic::transport::Channel> {
+    async fn connect(endpoint: Endpoint) -> Result<Self, ClientError> {
+        Ok(Self::connect(endpoint).await?)
+    }
+}
```
wimax-grapl, line 21: I guess I'm a tad confused here - I assume this isn't a recursive call - is there another `connect` and we're just converting the error type?

src/rust/rust-proto/src/graplinc/grapl/api/event_source/v1beta1/client.rs
```diff
@@ -1,102 +1,83 @@
...
-            create_event_source,
-            proto::CreateEventSourceRequest,
-            native::CreateEventSourceResponse,
-            RpcConfig::default(),
-        )
+    ) -> Result<native::CreateEventSourceResponse, ClientError> {
+        self.client
+            .execute(
+                request,
+                |status| status.code() == tonic::Code::Unavailable,
```
wimax-grapl, line 47: Potential followup thing to do, not necessary:

it'd be nice to have somewhere an easily-accessible-but-still-needs-to-be-specified retry predicate like
```
fn retry_on_unavailable(status: tonic_or_rust_proto_i_forget::Status) {
  status.code() == tonic::Code::Unavailable
}
```
and you could just pass that function pointer;
but I'm not sure if an `fn` can be passed as an `Fn` (or maybe it needs to be casted?)

wimax-grapl: PR checkpoint before lunch

src/rust/analyzer-dispatcher/src/main.rs
jgrillo-grapl: fixed

src/rust/plugin-sdk/generator-sdk/src/test_utils/test_ctx.rs
```diff
@@ -71,25 +71,28 @@ impl GeneratorTestContextInternals {
...
+            Duration::from_millis(500),
+            Duration::from_millis(500),
+            2,
+            Duration::from_millis(1),
+            Duration::from_millis(100),
+            Duration::from_millis(10),
+            20,
+            Duration::from_millis(10),
+            Duration::from_millis(250),
+        );
```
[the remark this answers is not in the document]
jgrillo-grapl: This is a constructor. It's here because none of the fields in this struct are optional, so a builder pattern isn't applicable. If, as you suggest, we made a constructor like this:

```rust
fn new(SomeStruct) -> Self { ... }
```

we'd still have to specify a constructor for `SomeStruct`.. unless you want to make all the fields on `SomeStruct` public and mutable? I think that's a pretty unfavorable tradeoff for what amounts to an aesthetic issue.

As you note, this constructor generally won't be used, except in tests. These objects are constructed by deserializing something given by a `figment::providers::*`. Currently we're using `figment::providers::Env`.

src/rust/rust-proto/src/graplinc/grapl/api/client.rs
jgrillo-grapl: yeah the idea here is we have a `P` for "proto" and `N` for "native" and the normal notation for a function of type `F` is `F<T, U>` which takes domain `T` to range `U`. I can make this more verbose.

src/rust/rust-proto/src/graplinc/grapl/api/event_source/v1beta1/client.rs
jgrillo-grapl: I'll try to document these a lot more clearly now that I've finished removing the healthcheck stuff. There is a little complexity here due to the desire to maintain encapsulation of the protocol buffer types.

The general idea is that we have a _blanket implementation_ of `Connect<C>` which is our public inteface--you give it a `ClientConfiguration` and it gives you a connected client ready to make requests. So rust-proto users need only import `Connect` and their client of choice--e.g. `PipelineIngressClient`. This will give them the ability to invoke `PipelineIngressClient::connect(client_configuration).await?`.

src/rust/plugin-execution-sidecar/src/work/plugin_work_processor.rs
```diff
@@ -52,20 +52,20 @@ pub trait PluginWorkProcessor {
...
-        config: &PluginExecutorConfig,
+        plugin_id: Uuid,
         work: ExecutionJob,
     ) -> Result<Self::ProducedMessage, PluginWorkProcessorError>;
 
     async fn ack_work(
         &self,
-        config: &PluginExecutorConfig,
-        pwq_client: &mut PluginWorkQueueServiceClient,
+        plugin_id: Uuid,
```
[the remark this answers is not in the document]
jgrillo-grapl: If at some future time we find a need for more information we'll have two options

(1) create some configuration struct
(2) add a field to the method signature

It seemed to me here that the `ack_work` verb in this API can't do anything without a `plugin_id`, since we can't know which plugin we're acknowledging work _for_. Same for all the other verbs in this API. Since the `PluginExecutorConfig` only had one field--the `plugin_id` I think it made sense to lift this to an actual first-class entity in the API--it doesn't _add_ anything to hide the fact that there's a `plugin_id` here.

src/rust/rust-proto/src/graplinc/grapl/api/graph_query/v1beta1/messages.rs
```diff
@@ -42,6 +45,15 @@ impl From<QueryId> for proto::QueryId {
     }
 }
 
+impl type_url::TypeUrl for QueryId {
+    const TYPE_URL: &'static str =
+        "graplsecurity.com/graplinc.grapl.api.graph_query_service.v1beta1.QueryId";
```
wimax-grapl, line 50: !!! This may have changed since you started this PR, but I renamed these to `graph_query` to remove this extraneous `_service`. 

https://github.com/grapl-security/grapl/pull/2032

src/rust/rust-proto/src/graplinc/grapl/api/event_source/v1beta1/client.rs
jgrillo-grapl: `Self` has type `EventSourceServiceClient<tonic::transport::Channel>` so this is calling the `connect` method on the protocol buffer compiler generated type `EventSourceServiceClient`. The `Connectable` interface is our bridge between the "generated" world and our application types. We implement `Connectable` on the generated types, and then `Connect<C: Connectable>` for our application types.

This sidesteps the issue that the generated types don't provide some interface. No matter, we'll just impose our own interface on them. Aren't traits cool?

src/rust/rust-proto/src/graplinc/grapl/api/graph_query_proxy/v1beta1/messages.rs
```diff
@@ -40,6 +41,15 @@ impl From<QueryGraphWithUidRequest> for proto::QueryGraphWithUidRequest {
     }
 }
 
+impl type_url::TypeUrl for QueryGraphWithUidRequest {
+    const TYPE_URL: &'static str =
+        "graplsecurity.com/graplinc.grapl.api.graph_query_proxy.v1beta1.QueryGraphWithUidRequest";
```
wimax-grapl, line 46: (graph_query_proxy, however, is correct, for the record)

src/rust/rust-proto/src/graplinc/grapl/api/event_source/v1beta1/client.rs
jgrillo-grapl: I hope these comments kind of make it more clear, I'll work on documenting these types now.

src/rust/rust-proto/src/graplinc/grapl/api/plugin_registry/v1beta1_client.rs
```diff
@@ -1,226 +1,196 @@
...
-        .chain(plugin_artifact.map(native::CreatePluginRequest::Chunk));
+        .chain(plugin_artifact.map(native::CreatePluginRequest::Chunk))
+        .map(proto::CreatePluginRequest::from);
 
-        self.create_plugin_raw(request).await
+        self.client
+            .execute_streaming(proto_stream, |mut client, request| async move {
+                client.create_plugin(request).await
+            })
+            .await
```
wimax-grapl, line 62: very slick and really impressed you got even the streaming case generalized :+1:

src/rust/e2e-tests/src/test_utils/context.rs
jgrillo-grapl: This is just saying "hydrate the configuration from environment variables containing this prefix". So in this service (the e2e-tests service) the `event_source_client_config` will be hydrated by environment variables prefixed with `"EVENT_SOURCE_CLIENT_"`. This basically establishes a contract between the service's environment and the service. I think, therefore, it would be inappropriate to define this prefix somewhere else (e.g. on `ClientConfig` or on `EventSourceClient`. It's not up to those types what the prefix should be, it's up to e2e-tests.

wimax-grapl (approved):
Overall, looks great - a couple nitpicks here and there but this is basically shippable once you've got integration tests working. Ping me for an approve once things are fixed up or review comments are responded to.

src/rust/rust-proto/src/graplinc/grapl/api/event_source/v1beta1/client.rs
jgrillo-grapl: Docs are done, I've elected to keep the terse type parameters but document what they mean.

src/rust/plugin-execution-sidecar/src/sidecar_client/generator_client.rs
jgrillo-grapl: fixed

src/rust/plugin-execution-sidecar/src/work/analyzer_work_processor.rs
jgrillo-grapl: fixed

nomad/grapl-core.nomad
```diff
@@ -236,8 +236,27 @@ job "grapl-core" {
 
       env {
         # Upstreams
-        PLUGIN_WORK_QUEUE_CLIENT_ADDRESS = "http://${NOMAD_UPSTREAM_ADDR_plugin-work-queue}"
-        PLUGIN_REGISTRY_CLIENT_ADDRESS   = "http://${NOMAD_UPSTREAM_ADDR_plugin-registry}"
+        PLUGIN_WORK_QUEUE_CLIENT_ADDRESS                       = "http://${NOMAD_UPSTREAM_ADDR_plugin-work-queue}"
+        PLUGIN_WORK_QUEUE_CLIENT_REQUEST_TIMEOUT               = "1s"
```
wimax-grapl, line 240: i sort of don't love all of these being in env{} - perhaps an eventual config file instead, since figment supports all sorts of difference sources - **but this is totally fine to land for now**

wimax-grapl (approved):

PR #2079, opened 2022-10-26
Add the "grapl-metrics" crate and an integration test

wimax-grapl:
https://github.com/grapl-security/issue-tracker/issues/1048

Adds a library that makes it easy (well, as easy as otel can be) to emit custom application metrics from Rust code to the otel-collector (and, from there, to Lightstep). 

Unfortunately, there's not really a great way to **automated-ly** test that the metrics actually went through. 
However, I've included instructions on how to manually test that they reached the otel-collector.

### DIFF
--- a/nomad/observability.nomad
+++ b/nomad/observability.nomad
@@ -25,7 +25,8 @@ job "otel-collector" {
 
       # Receivers
       port "otlp-grpc" {
-        to = 4317
+        to     = 4317
+        static = 4317
       }
 
       port "otlp-http" {
@@ -56,6 +57,10 @@ job "otel-collector" {
 
     service {
       port = "otlp-http"
+    }
+
+    service {
+      port = "otlp-grpc"
     }
 
     service {
--- a/nomad/rust-integration-tests.nomad
+++ b/nomad/rust-integration-tests.nomad
@@ -55,6 +55,14 @@ variable "kafka_credentials" {
   })
 }
 
+variable "observability_env_vars" {
+  type        = string
+  description = <<EOF
+With local-grapl, we have to inject env vars for Opentelemetry.
+In prod, this is currently disabled.
+EOF
+}
+
 variable "rust_log" {
   type        = string
   description = "Controls the logging behavior of Rust-based services."
@@ -92,6 +100,10 @@ variable "plugin_work_queue_db" {
 
 locals {
   dns_servers = [attr.unique.network.ip-address]
+
+  # Currently we use the same version for all containers. As such we pick one container to get the version from
+  app_version                      = split(":", var.container_images["rust-integration-tests"])[1]
+  default_otel_resource_attributes = "service.version=${local.app_version},host.hostname=${attr.unique.hostname}"
 }
 
 job "rust-integration-tests" {
@@ -192,7 +204,12 @@ job "rust-integration-tests" {
         image = var.container_images["rust-integration-tests"]
       }
 
-      # This writes an env file that gets read by the task automatically
+      template {
+        data        = var.observability_env_vars
+        destination = "observability.env"
+        env         = true
+      }
+
       template {
         data        = var.aws_env_vars_for_local
         destination = "aws-env-vars-for-local.env"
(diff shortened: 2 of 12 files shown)


nomad/rust-integration-tests.nomad
```diff
@@ -92,6 +100,10 @@ variable "plugin_work_queue_db" {
 
 locals {
   dns_servers = [attr.unique.network.ip-address]
+
+  # Currently we use the same version for all containers. As such we pick one container to get the version from
+  app_version                      = split(":", var.container_images["rust-integration-tests"])[1]
```
christophermaier, line 105: What, specifically, does `app_version` mean? What will the impact be when the version is not the same for all containers?

Note that we _don't_ actually use the same version for all containers. If a container image hasn't changed, it shouldn't be built and uploaded again, meaning that we can have a number of different versions in play at the same time.

src/rust/grapl-metrics/src/setup_metrics.rs
```diff
@@ -0,0 +1,89 @@
...
+    // who accidentally call `setup_metrics()` from a default `tokio::test`.
+
+    // This will be better if https://github.com/tokio-rs/tokio/issues/5088
+    // ever gets landed.
+    let err_if_singlethread = catch_unwind(|| task::block_in_place(|| {}));
+    match err_if_singlethread {
+        Ok(_) => runtime::Tokio,
+        Err(_) => panic!(
+            r#"
+            `setup_metrics()` called from a single-threaded Tokio runtime;
```
christophermaier, line 53: 🤔 It's a little odd for assumptions about where the function is called from to be embedded in this message.

Maybe it's better to just embed this into `setup_metrics` directly?

nomad/observability.nomad
```diff
@@ -25,7 +25,8 @@ job "otel-collector" {
 
       # Receivers
       port "otlp-grpc" {
-        to = 4317
+        to     = 4317
+        static = 4317
```
christophermaier, line 29: What's the significance of this change?

src/rust/grapl-metrics/src/setup_metrics.rs
wimax-grapl: Yes, I could inline it, but I'd much prefer to not include this kind-of-gross-mess in an otherwise straightforward function about setting up metrics. 

It's not a public function, and its entire purpose is to assert things about the runtime it's currently running in. 

Keep in mind: this function is here as a guard rail so other people don't get stuck down the 4-day "what the heck is going wrong" hole I got down in, so perhaps I'm being a bit too precious about this

nomad/rust-integration-tests.nomad
wimax-grapl: This is a pattern I carried over from the rest of our Nomad files; perhaps ping Thomas about it?

image

christophermaier: Ping @twunderlich-grapl ☝️

nomad/observability.nomad
wimax-grapl: Basically, the previous version meant

"There is a named port named `otlp-grpc`. In the container, that port points at :4317. From the outside, I make no promises about which port number `otlp-grpc` is served from - it's dynamically chosen." 

The new version means
"There is a named port named `otlp-grpc`. In the container, that port points at :4317. From the outside, I make a promise that you can access it at {attr.unique.network.ip-address}:4317"

nomad/rust-integration-tests.nomad
wimax-grapl: sorry, i actually copy pasted that comment but since there's only one container here then that comment doesnt apply

christophermaier (approved):
Approved, though I'm still curious to hear @twunderlich-grapl's thoughts re: https://github.com/grapl-security/grapl/pull/2079#discussion_r1006047475

nomad/rust-integration-tests.nomad
twunderlich-grapl: @christophermaier,
This currently assumes that the version tag is the same for the entire job, since I thought we were retagging all the containers with the same version. If that's incorrect, we just need to move the app_version split statement into each job.

twunderlich-grapl: I can go through and update that in a follow-up PR

PR #2084, opened 2022-10-27
Set service.version on the service level

ghost:
Fixes https://github.com/grapl-security/issue-tracker/issues/1051

This sets the service.version tag at the service level instead of the job level. As Grapl only rebuilds and tags containers that have changes, the versions for each service within a job are not the same.

Tests

### DIFF
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -189,7 +189,7 @@ locals {
   # Set up default tags for otel traces via the OTEL_RESOURCE_ATTRIBUTES env variable. Format is key=value,key=value
   # We're setting up defaults on a per-job basis, but these can be expanded on a per-service basis as necessary.
   # Examples of keys we may add in the future: language, instance_id/ip, team
-  default_otel_resource_attributes = "service.version=${local.app_version},host.hostname=${attr.unique.hostname}"
+  default_otel_resource_attributes = "host.hostname=${attr.unique.hostname}"
 }
 
 job "grapl-core" {
--- a/nomad/grapl-core.nomad
+++ b/nomad/grapl-core.nomad
@@ -16,6 +16,14 @@ variable "container_images" {
 EOF
 }
 
+variable "container_versions" {
+  type        = map(string)
+  description = <<EOF
+  A map of $NAME_OF_TASK to that task's docker image version.
+  (See DockerImageVersion in Pulumi for further documentation.)
+EOF
+}
+
 variable "aws_env_vars_for_local" {
   type        = string
   description = <<EOF
@@ -181,9 +189,6 @@ locals {
   # Set up default tags for otel traces via the OTEL_RESOURCE_ATTRIBUTES env variable. Format is key=value,key=value
   # We're setting up defaults on a per-job basis, but these can be expanded on a per-service basis as necessary.
   # Examples of keys we may add in the future: language, instance_id/ip, team
-
-  # Currently we use the same version for all containers. As such we pick one container to get the version from
-  app_version                      = split(":", var.container_images["analyzer-dispatcher"])[1]
   default_otel_resource_attributes = "service.version=${local.app_version},host.hostname=${attr.unique.hostname}"
 }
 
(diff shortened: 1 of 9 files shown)


twunderlich-grapl:
Right now I've chosen to create a second Mapping. An alternative approach (and more efficient) would be to create a single container_meta Mapping that looks something like
```
{
    container_name: {
        image: container_image_id
        version: container_version
    }
}
```

The tradeoff here is between verbosity (every affected line in Nomad would be longer) and efficiency

pulumi/infra/docker_images.py
```diff
@@ -16,6 +16,18 @@
 - an image pulled from Cloudsmith
     "docker.cloudsmith.io/grapl/raw/graph-merger:20211105192234-a86a8ad2"
 """
+DockerImageVersion = NewType("DockerImageId", str)
```
christophermaier, line 19: Should this be `NewType("DockerImageVersion", str)` instead?

nomad/grapl-core.nomad
```diff
@@ -258,7 +263,7 @@ job "grapl-core" {
         RUST_BACKTRACE = local.rust_backtrace
         RUST_LOG       = var.rust_log
 
-        OTEL_RESOURCE_ATTRIBUTES = local.default_otel_resource_attributes
+        OTEL_RESOURCE_ATTRIBUTES = "${local.default_otel_resource_attributes},service.version=${var.container_versions["analyzer-dispatcher"]}"
```
twunderlich-grapl, line 266: Surprisingly the `"${var.foo["bar"}"` is the correct syntax. Single quotes are illegal in hcl2 and escaping quotes within interpolated values is not allowed

pulumi/grapl/__main__.py
```diff
@@ -93,6 +93,15 @@ def _container_images(artifacts: ArtifactGetter) -> Mapping[str, DockerImageId]:
     }
 
 
+def _container_versions(
+    container_images: Mapping[str, DockerImageId]
+) -> Mapping[str, DockerImageVersion]:
+    container_versions = dict()
+    for key, value in container_images.items():
+        container_versions[key] = value.split(":")[1]
+    return container_versions
```
christophermaier, line 102: It looks like this logic is the same in all the Pulumi projects; can we create a common implementation to cut down the repetition?

wimax-grapl, line 102: slightly easier but not necessarily worth doing: a dict comprehension

```
return {
  key: value.split(":")[1]
  for key, value in container_images.items()
}
```

pulumi/infra/docker_images.py
wimax-grapl, line 19: change the string to match the type's name

pulumi/python_integration_tests/__main__.py
```diff
@@ -70,10 +79,15 @@ def main() -> None:
         # on-disk Grapl repo.
         # FIXME: make python integration tests work in AWS.
 
+        container_images = (_python_integration_container_images(artifacts),)
```
wimax-grapl, line 82: uhhhhh you sure about this? this makes container_images a tuple

pulumi/rust_integration_tests/__main__.py
```diff
@@ -35,6 +35,15 @@ def _rust_integration_container_images(
...
 
+def _rust_integration_container_versions(
+    container_images: Mapping[str, DockerImageId]
+) -> Mapping[str, DockerImageVersion]:
+    container_versions = dict()
+    for key, value in container_images.items():
+        container_versions[key] = value.split(":")[1]
+    return container_versions
+
+
```
wimax-grapl, line 46: i feel like we could just centralize this on say docker_images.py

pulumi/infra/docker_images.py
twunderlich-grapl: Yes 🤦

pulumi/python_integration_tests/__main__.py
twunderlich-grapl: Ugh, no. I must have typed a paren by accident or soemthing here and my IDE autoexpanded. Thanks for catching this

pulumi/rust_integration_tests/__main__.py
twunderlich-grapl: Done

christophermaier (approved):

wimax-grapl (approved):

PR #2093, opened 2022-11-07
Add a simpler analyzer, which is also now successfully emitting ExecutionHits!

wimax-grapl:
Our existing analyzer, which looks for a parent-child relationship, was not triggering and was very hard to debug why.

I've added lots of debug logging to help diagnose, which led me to realize analyzers are currently relatively broken. So I fixed those breaks!

Anyway, this new analyzer now successfully emits an ExecutionHit back to PWQ. I'll integrate that into e2e test enforcement next!

### DIFF
--- a/src/python/grapl-plugin-sdk/example_analyzers/process_named_svchost.py
+++ b/src/python/grapl-plugin-sdk/example_analyzers/process_named_svchost.py
@@ -1,3 +1,7 @@
+"""
+This is meant as an extremely simple Analyzer to get the pipeline to fire
+during integration tests.
+"""
 from datetime import datetime
 
 from grapl_plugin_sdk.analyzer.analyzer import (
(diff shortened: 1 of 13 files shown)


src/python/grapl-plugin-sdk/example_analyzers/process_named_svchost.py
```diff
@@ -0,0 +1,81 @@
...
+)
+from python_proto.api.plugin_sdk.analyzers.v1beta1.messages import (
+    AnalyzerName,
+    ExecutionHit,
+)
+from python_proto.common import Timestamp
+from python_proto.grapl.common.v1beta1.messages import NodeType, PropertyName
+
+
+class ProcessNamedSvchost(Analyzer):
```
wimax-grapl, line 27: # The majority of this analyzer, including its comments like "???" are just copied from the suspicious_svchost_analyzer. The main difference is the node query is way simpler.

ghost (approved):
So far this makes sense to me

jgrillo-grapl (approved):

PR #2094, opened 2022-11-07
Dump metrics and traces as files

ghost:
This exports two files from the otel-collector, one for metrics and one for traces.

make up 
make dump-artifacts-local


jgrillo-grapl (approved):
👏

PR #2096, opened 2022-11-09
Analyzers now emit their ExecutionHit back to PWQ and then to a Kafka topic. Test this under E2E. 

wimax-grapl:
Previously, when a plugin would `acknowledge_analyzer`, we'd simply send back a `success: bool`.

Now, we actually send the data generated by the Analyzer: an ExecutionHit.

Then, PWQ will send that generated data on to the next service over Kafka - just like it does with generators and the `generated-graphs` topic.

Finally, this means we can actually test the ExecutionHit in the e2e test! Sure enough, our analyzer for `process_named_svchost.py` results in an ExecutionHit for svchost.exe!!!


src/rust/rust-proto/tests/test_utils/strategies.rs
```diff
@@ -1236,12 +1217,87 @@ pub mod analyzer_sdk {
...
-            string_property_updates().prop_map(Update::StringProperty),
-            uint_64_property_updates().prop_map(Update::Uint64Property),
-            int_64_property_updates().prop_map(Update::Int64Property),
-            edge_updates().prop_map(Update::Edge),
+            Just(native::ExecutionResult::ExecutionMiss(
+                native::ExecutionMiss {}
+            )),
+            // TODO wimax Nov 10 2022: Flesh this out to also include ExecutionHit.
+            // Currently those are only tested in python, not rust.
+            // Just(native::ExecutionResult::ExecutionHit(native::ExecutionHit{})),
```
wimax-grapl, line 1300: unfortunately it'd be a lot of proptest boilerplate that I don't think is the right use of my time given Certain Timing Constraints

src/proto/graplinc/grapl/api/graph/v1beta1/types.proto
```diff
@@ -200,19 +200,3 @@ message Lens {
   // score (absent prior to engagement-creator)
   optional uint64 score = 4;
 }
-
-// Result computed by an analyzer when it makes a detection.
-message ExecutionHit {
```
wimax-grapl, line 205: This is replaced by the similar ExecutionHit in analyzer_sdk. This was ported over before we did GraphQuery. The structure is largely the same.

src/rust/e2e-tests/tests/sysmon_log_e2e_test.rs
```diff
@@ -140,6 +141,23 @@ async fn test_sysmon_log_e2e(ctx: &mut E2eTestContext) -> eyre::Result<()> {
     ))
     .await;
 
+    let execution_hits_scanner_handle = KafkaTopicScanner::new(
+        ConsumerConfig::with_topic("analyzer-executions"),
```
wimax-grapl, line 145: (fwiw, this topic is probably poorly named but matches the one defined in grapl-infrastructure
```
            # This list must match the set of all topics specified in
            # grapl-infrastructure/pulumi/confluent_cloud/index.ts
```
)

src/python/grapl-plugin-sdk/grapl_plugin_sdk/analyzer/service_impl.py
```diff
@@ -69,6 +69,7 @@ class AnalyzerServiceImpl:
 
     def __post_init__(self) -> None:
         self._graph_query = self._analyzer.query().into_graph_query()
+        LOGGER.info("Graph query", graph_query=str(self._graph_query))
```
wimax-grapl, line 72: This happens once at the top of the analyzer and helped me debug some stuff, so, keeping it here

src/rust/rust-proto/src/graplinc/grapl/api/graph_query/v1beta1/messages.rs
```diff
@@ -18,6 +18,21 @@ use crate::{
     SerDeError,
 };
 
+/// Basically cribbed from std::collection::HashMap's implementation.
```
jgrillo-grapl, line 21: why are we using this non-standard hashmap?

jgrillo-grapl (approved):

src/rust/rust-proto/src/graplinc/grapl/api/graph_query/v1beta1/messages.rs
wimax-grapl: Great question. It's what I received in a Colin code-dump and never really questioned it. FWIW it appears to be semi-standard (they use it in rustc)?

PR #2098, opened 2022-11-09
Back out an undesirable regression in integration tests being run

wimax-grapl:
See PR. Introduced with https://github.com/grapl-security/grapl/pull/2093 by accident.

### DIFF
--- a/src/rust/Dockerfile
+++ b/src/rust/Dockerfile
@@ -178,10 +178,6 @@ sysmon-generator/integration_tests
 uid-allocator/integration_tests
 )
 
-INTEGRATION_TEST_FEATURES=(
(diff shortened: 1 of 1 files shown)


src/rust/Dockerfile
```diff
@@ -178,10 +178,6 @@ sysmon-generator/integration_tests
 uid-allocator/integration_tests
 )
 
-INTEGRATION_TEST_FEATURES=(
-e2e-tests/integration_tests
-)
-
```
wimax-grapl, line 184: (this overrode the above INTEGRATION_TEST_FEATURES=)

jgrillo-grapl (approved):

PR #2099, opened 2022-11-09
Customize which integ tests build/run with an environment variable

wimax-grapl:
TLDR: This enables
```
RUST_INTEGRATION_TEST_FEATURES_OVERRIDE="e2e-tests/integration_tests,uid-allocator/integration_tests" \
  make test-integration-rust
```

It enables the behavior I initially wanted (and accidentally committed) that led to https://github.com/grapl-security/grapl/pull/2098

---

I tested it locally without specifying it and with specifying it.

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -92,6 +92,14 @@ variable "DIST_DIR" {
 variable "PROTOC_VERSION" {
 }
 
+# Override this variable to limit the integration tests run by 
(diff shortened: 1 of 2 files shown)


twunderlich-grapl (approved):
Nice improvement.
Long-term, I'd probably want to figure out a way to generate the INTEGRATION_TEST_FEATURE but that can wait

PR #2105, opened 2022-11-17
Universal server tracing function

wimax-grapl:
Previously we had the same trace logic repeated over and over again - and worse, it usually included the full `headers` which contains all sorts of useless fields that clog up logging. (also `uri` is mostly useless when `method` is included).

So now I just do request id and trace id.

Don't worry - services can, in the future, specify addl trace fns if they really want to!

### DIFF
--- a/src/rust/rust-proto/src/server_tracing.rs
+++ b/src/rust/rust-proto/src/server_tracing.rs
@@ -1,0 +1,10 @@
+pub fn server_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
+    let headers = request.headers();
+    tracing::info_span!(
+        "server_trace_fn",
+        request_id = ?headers.get("x-request-id"),
+        trace_id = ?headers.get("x-trace-id"),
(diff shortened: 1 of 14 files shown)


src/rust/rust-proto/src/server_tracing.rs
```diff
@@ -0,0 +1,10 @@
+pub fn server_trace_fn<T>(request: &http::Request<T>) -> tracing::Span {
+    let headers = request.headers();
+    tracing::info_span!(
+        "server_trace_fn",
```
christophermaier, line 4: This names all the spans `server_trace_fn`, right? If so, is that desirable?

christophermaier (approved):
Apart from a small question, this looks like a good cleanup.

ghost (approved):

wimax-grapl:
Abandoning because 
1. we don't want each span to have the same name or they'll all get bucketed together
2. info_span! expects a const, unfortunately, so you gotta specify it there it can't be passed in

PR #2106, opened 2022-11-17
A number of debug-ability improvements for analyzers, and a few bugfixes here and there

wimax-grapl:
I've been trying to figure out why suspicious_svchost doesn't work. I still don't know why it fails. 

But: I've found a few errors along the way and wanted to get this work in.
1. `    def into_graph_query(self) -> GraphQuery:` was failing because we didn't include the neighbor queries.
2. sometimes we'd get timeouts waiting for analyzer/generators to work, so I bumped their timeouts.
3. some work on propagating trace ids

### DIFF
--- a/src/python/grapl-plugin-sdk/grapl_plugin_sdk/analyzer/service_impl.py
+++ b/src/python/grapl-plugin-sdk/grapl_plugin_sdk/analyzer/service_impl.py
@@ -4,7 +4,7 @@ import dataclasses
 import os
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import TYPE_CHECKING, Final, Protocol, runtime_checkable, Mapping
+from typing import TYPE_CHECKING, Final, Mapping, Protocol, runtime_checkable
 from uuid import UUID, uuid4
 
 import grpc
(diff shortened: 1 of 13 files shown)


src/python/grapl-plugin-sdk/grapl_plugin_sdk/analyzer/service_impl.py
```diff
@@ -60,6 +61,21 @@ class PropertyUpdate(Protocol):
...
+
+    def __init__(self, grpc_metadata: grpc_aio.Metadata | None) -> None:
+        grpc_metadata = grpc_metadata or {}
+        self.trace_id = grpc_metadata.get("x-trace-id")
+
+    def to_grpc_metadata(self) -> GrpcOutboundMetadata:
+        metadata = [
+            ("x-trace-id", self.trace_id),
+        ]
+        metadata_no_nones = [(k, v) for (k, v) in metadata if v is not None]
```
christophermaier, line 75: Is more going to be added to `metadata` at any point? Otherwise it seems like all this could be simplified to
```python
return (("x-trace-id", self.trace_id)) if self.trace_id else ()
```

src/python/python-proto/python_proto/metadata.py
```diff
@@ -0,0 +1,2 @@
+# N tuples of (str, str)
+GrpcOutboundMetadata = tuple[tuple[str, str | bytes], ...]
```
christophermaier, line 2: Is there a reason this is an unbounded tuple of tuples? That seems like a more awkward data type to interact with than a dict.

wimax-grapl: this is unfortunately what grpc expects at this callsite:
```
    def query_with_uid(
        self,
        node_uid: Uid,
        graph_query: GraphQuery,
        metadata: GrpcOutboundMetadata | None = None,
    ) -> QueryGraphWithUidResponse:
        request = QueryGraphWithUidRequest(
            node_uid=node_uid,
            graph_query=graph_query,
        )
        proto_response = self.proto_client.QueryGraphWithUid(
            request.into_proto(),
            timeout=5,
            metadata=metadata,  # <------- here
```

i'm happy to keep it wrapped as a dict up to that point?

christophermaier: At first glance, that would seem to be the way to go; keep all the GRPC nastiness confined to as small a footprint as possible and make the rest of the code easier to interact with.

src/python/grapl-plugin-sdk/grapl_plugin_sdk/analyzer/service_impl.py
christophermaier: This question probably just disappears if metadata is modeled as a dict to begin with:
```python
metadata = {}
if self.trace_id:
    metadata["x-trace-id"] = self.trace_id
# add more as needed
return metadata
```

colin-grapl (approved):

PR #2107, opened 2022-11-17
Change a test-utils panic into a Result; add a wait-for-healthy-generator to analyzer_dispatcher_integration_tests to harden it

wimax-grapl:
1. We had this ugly panic in our check-PWQ utils that I've replaced with an eyre::Result.
2. analyzer_dispatcher_integration_tests fails sometimes doe to the generator not being ready in time. I fortunately already wrote `assert_eventual_health` for this :)

### DIFF
--- a/src/rust/plugin-work-queue/src/test_utils.rs
+++ b/src/rust/plugin-work-queue/src/test_utils.rs
@@ -99,7 +99,8 @@ impl PsqlQueueTestExtensions for PsqlQueue {
 pub async fn scan_for_generator_plugin_message_in_pwq(
     psql_queue: PsqlQueue,
     plugin_id: uuid::Uuid,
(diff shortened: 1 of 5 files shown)


src/rust/plugin-work-queue/src/test_utils.rs
```diff
@@ -99,62 +99,66 @@ impl PsqlQueueTestExtensions for PsqlQueue {
 pub async fn scan_for_generator_plugin_message_in_pwq(
     psql_queue: PsqlQueue,
     plugin_id: uuid::Uuid,
-) -> Option<NextExecutionRequest> {
```
wimax-grapl, line 102: as it turns out, this as an Option never made sense - it was either Some or panic!
so now it's Ok or Err

twunderlich-grapl (approved):

